PR #11, opened 2021-12-14
[Bugfix] Prevent Incorrect Failure Handling Calling SDK Error

JoshuaLai:
## Purpose
Updating logic to check against the error first before checking the call object
Calling SDK has an upcoming task where they will no longer be creating a call object when they return us an error. 
So when that change comes down we will actually fail incorrectly and not bubble the error up to the user. 

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```


palatter (dismissed):
Looks good.

kevinyulu (dismissed):
tested group call and teams call when connected and disconnected from internet. looks good

palatter (approved):

kevinyulu (approved):

PR #13, opened 2021-12-14
[Bugfix] Enforce Orientation Restrictions Upon Navigation Back to Setup Screen

kevinyulu:
## Purpose

* Fixed the incorrect screen orientation issue when app lands back in setup screen due to error. The app should set back to the supported orientation when lands back in the setup screen.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code
1. Install and run the app, tap either UIKit or Swift UI
2. Fill the token, and tokenURL, display name, group ID (or teams link)
3. Tap Start Experience button, entering the setup screen
4. when entering the setup screen, immediately rotate device from portrait to landscape 

## What to Check
* check the screen orientation should set back to portrait in setup view


AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Container/ContainerUIHostingController.swift
```diff
@@ -56,9 +56,21 @@ class ContainerUIHostingController: UIHostingController<ContainerUIHostingContro
...
+                    if orientation == .portrait &&
+                        (UIDevice.current.orientation.isLandscape
+                         || UIDevice.current.orientation.isFlat
+                         || UIDevice.current.orientation == .unknown
+                        ) {
+
+                        // Apply a delay here to allow the previous orientation change to finish,
+                        // then reset orientations
+                        DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
+                            UIDevice.current.rotateTo(oritation: UIInterfaceOrientation.portrait)
```
AmyL219, line 68: I am not quite understand why do we need to add this step. Let's have a call regarding this

kevinyulu: yes this is to make sure app will transition to setup screen and switch the orientation back to portrait. 
The reason of adding a delay here is to allow the previous orientation change to complete as the previous orientation change can be competing with the new orientation reset from `rotateTo`. If there is no delay then the new orientation reset will not be executed and setup screen will be in landscape orientation.

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Container/ContainerUIHostingController.swift
```diff
@@ -56,9 +56,21 @@ class ContainerUIHostingController: UIHostingController<ContainerUIHostingContro
             .dropFirst()
```
AmyL219, line 56: .removeDuplicated

kevinyulu: updated

AmyL219 (dismissed):

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Container/ContainerUIHostingController.swift
```diff
@@ -53,12 +53,20 @@ class ContainerUIHostingController: UIHostingController<ContainerUIHostingContro
...
             .receive(on: RunLoop.main)
-            .dropFirst()
+            .removeDuplicates()
             .sink(receiveValue: { orientation in
                 if orientation == .portrait || orientation == .landscape {
-                    UIDevice.current.endGeneratingDeviceOrientationNotifications()
+                    // Apply a delay here to allow the previous orientation change to finish,
+                    // then reset orientations
+                    DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
+
```
JoshuaLai, line 62: Why does this logic have to happen in the `ContainerUIHostingController`? Could we not implement this logic somewhere in the `SetupView`?

AmyL219 (approved):

JoshuaLai:
Tested on device and it works. It's a little odd that we see the set up page in landscape mode for just a second but it eventually fixes its orientation.

JoshuaLai (approved):

kevinyulu:
> Tested on device and it works. It's a little odd that we see the set up page in landscape mode for just a second but it eventually fixes its orientation.

Yes the 0.5 wait time is to let the previous orientation change to finish (if any). If there is no such wait time then there is an edge case such that the orientation set to portrait (orientation change 1) will coincide with the ongoing previous orientation change (orientation change 0) , resulting in a failure in app setting the orientation back to portrait. This edge case would require us to time it in order to get a repro. 
 
Repro step
* launch app, fill the call info, 
* go to setup screen, disconnect internet
* go to call screen, time it and wait till the error is about to be thrown from SDK (~1 second after seeing the calling screen)
* then immediately rotate device from portrait to landscape (aka the orientation change 0 will be triggered) 
* app will take user back to setup while the orientation is transition to landscape as well
* user will see setup screen in landscape

PR #18, opened 2021-12-15
[Bugfix] Dispatch camera preview no matter the current state

cheukchuen:
## Purpose

* Current state may be .on or .paused (background mode) prior to network disconnect

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## What to Check
* Setup screen video preview shows camera feed when navigating back to setup on error (if camera permission granted)
  * when video is off during call
  * when video is on during call
  * when app is in background mode during call

### DIFF
--- a/AzureCommunicationUI/AzureCommunicationUI/Redux/Middleware/CallingMiddlewareHandler.swift
+++ b/AzureCommunicationUI/AzureCommunicationUI/Redux/Middleware/CallingMiddlewareHandler.swift
@@ -46,8 +46,7 @@ class CallingMiddlewareHandler: CallingMiddlewareHandling {
                 case .failure(let error):
                     self.handle(error: error, errorCode: CallCompositeErrorCode.callJoin, dispatch: dispatch)
                 case .finished:
-                    if state.permissionState.cameraPermission == .granted,
-                       state.localUserState.cameraState.operation == .off {
+                    if state.permissionState.cameraPermission == .granted {
                         dispatch(LocalUserAction.CameraPreviewOnTriggered())
                     }
                 }


jimchou-dev (approved):

kevinyulu (approved):

AzureCommunicationUI/AzureCommunicationUI/Redux/Middleware/CallingMiddlewareHandler.swift
```diff
@@ -46,8 +46,7 @@ class CallingMiddlewareHandler: CallingMiddlewareHandling {
                 case .failure(let error):
                     self.handle(error: error, errorCode: CallCompositeErrorCode.callJoin, dispatch: dispatch)
                 case .finished:
-                    if state.permissionState.cameraPermission == .granted,
-                       state.localUserState.cameraState.operation == .off {
+                    if state.permissionState.cameraPermission == .granted {
                         dispatch(LocalUserAction.CameraPreviewOnTriggered())
```
AmyL219, line 50: Instead of removing state.localUserState.cameraState.operation == .off. 
Shouldn't we change operation to .off when we back from CallingView to SetupView?
state.localUserState.cameraState.operation indicate if it's a localVideoStream with/without an ongoing call. When it navigate back to setup view, we should update the operation to off

AmyL219: Sorry, I was confused `operation` with `transmission` in my last comment.
But for this bug, no matter if I close/open the local video on the calling view, the preview view is always on in my testing. I don't think we need to remove the .operation check here. If we remove, it would call an extra CameraPreviewOnTriggered() action when the view back SetupView when the local video view is already on. 

I can't reproduce the bug. Confirmed with @iaulakh, he saw the bug once in 20 times, and I tried 30 times but can't reproduce it. I think we can close this bug ticket for now.

JoshuaLai: I agree. I don't think this PR will actually fix bug we are trying to solve. It might be that there is actually an issue in the CallinSDK side.

PR #20, opened 2021-12-17
[BugFix] Listening for new clean up action 

JoshuaLai:
## Purpose
New `CleanupCallStates` action that will reset the the recording and transcript value. 

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
User A and User B join teams call
User A begins recording and transcription
User B loses internet and is kicked out into the set up view
User A stops the recording and transcription
User B turns on internet and rejoins call
Verify the banner is not showing in call


palatter (commented):
Looks good and tested.

palatter (approved):

AzureCommunicationUI/AzureCommunicationUI/Presentation/Manager/CompositeErrorManager.swift
```diff
@@ -33,8 +33,10 @@ class CompositeErrorManager: ErrorManager {
             case .fatal:
                 update(error: error)
                 respondToFatalError(code: error.code)
+                store.dispatch(action: CallingAction.CleanupCallStates())
             case .callState:
                 update(error: error)
+                store.dispatch(action: CallingAction.CleanupCallStates())
```
kevinyulu, line 39: I tested the iOS app by following the repro step in the PR. The UX looks good. One observation is that when the call has an error, this `case` is triggered many times. By adding dispatch here will make app dispatch many `CallingAction.CleanupCallStates()` actions as well. Maybe there is a loop going on?

JoshuaLai: This is an interesting find. I will take a look into it. If we are dispatching the same error multiple times we may need a check in here. Did this happen when you were disconnecting from the internet?

kevinyulu: > Did this happen when you were disconnecting from the internet?

@JoshuaLai I remember I started to see the dispatches as soon as the call was disconnected do to Internet error. (UX: as soon as I was landed back on the setup screen)

AzureCommunicationUI/AzureCommunicationUI/Redux/Action/CallingAction.swift
```diff
@@ -25,6 +25,8 @@ struct CallingAction {
     struct TranscriptionStateUpdated: Action {
         let isTranscriptionActive: Bool
     }
+
+    struct CleanupCallStates: Action {}
```
jimchou-dev, line 29: Should there be unit test for this action?

JoshuaLai: Yes thats a good point, I need to run this by Wayne and Inderpal as well. So I will introduce the unit tests after that conversation.

JoshuaLai:
Discussed a new plan for how to clean up the call states and error object.

PR #21, opened 2021-12-21
[Bugfix] Make app remember audio option (if available) from the previous call.

kevinyulu:
## Purpose
Make app remember audio option (if available) from the previous call. The OS currently default the audio option to **speaker** if there is no previous call. 

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code
* Launch the app, fill the call info
* go to setup screen, 
* check the audio option
* join the call, (try changing the audio option)
* and then leave it call,
* rejoin the call
* check audio option in setup screen
```
```

## What to Check
* check the audio option in setup screen is set to the option from the previously established call.

### DIFF
--- a/AzureCommunicationUI/AzureCommunicationUI/Presentation/Manager/AudioSessionManager.swift
+++ b/AzureCommunicationUI/AzureCommunicationUI/Presentation/Manager/AudioSessionManager.swift
@@ -14,28 +14,19 @@ class AppAudioSessionManager: AudioSessionManager {
     private let logger: Logger
     private let store: Store<AppState>
     var cancellables = Set<AnyCancellable>()
-    var isSpeakerphoneOn: Bool = false
 
     init(store: Store<AppState>,
          logger: Logger) {
         self.store = store
         self.logger = logger
-        initializeAudioDeviceState()
+        let currentAudioDevice = getCurrentAudioDevice()
         self.setupAudioSession()
+        store.dispatch(action: LocalUserAction.AudioDeviceChangeRequested(device: currentAudioDevice))
         store.$state
             .sink { [weak self] state in
                 self?.receive(state: state)
             }.store(in: &cancellables)
 
-    }
-
-    private func initializeAudioDeviceState() {
-        isSpeakerphoneOn = getCurrentAudioDevice() == .speaker
-        if isSpeakerphoneOn {
-             store.dispatch(action: LocalUserAction.AudioDeviceChangeSucceeded(device: .speaker))
-        } else {
-             store.dispatch(action: LocalUserAction.AudioDeviceChangeSucceeded(device: .receiver))
-        }
     }
 
     private func receive(state: AppState) {
@@ -62,7 +53,6 @@ class AppAudioSessionManager: AudioSessionManager {
                                                            .interruptSpokenAudioAndMixWithOthers,
                                                            .allowBluetoothA2DP]
             try audioSession.setCategory(.playAndRecord, mode: .voiceChat, options: options)
-            try audioSession.overrideOutputAudioPort(isSpeakerphoneOn ? .speaker : .none)
             try audioSession.setActive(true)
         } catch let error {
             logger.error("Failed to set audio session category:\(error.localizedDescription)")
@@ -97,12 +87,10 @@ class AppAudioSessionManager: AudioSessionManager {
         do {
             try audioSession.setActive(true)
             try audioSession.overrideOutputAudioPort(audioPort)
-            isSpeakerphoneOn = audioPort == .speaker
             store.dispatch(action: LocalUserAction.AudioDeviceChangeSucceeded(device: selectedAudioDevice))
         } catch let error {
             logger.error("Failed to select audio device, reason:\(error.localizedDescription)")
             store.dispatch(action: LocalUserAction.AudioDeviceChangeFailed(error: error))
-            isSpeakerphoneOn = false
         }
     }
 
--- a/AzureCommunicationUI/AzureCommunicationUI/Presentation/Manager/AudioSessionManager.swift
+++ b/AzureCommunicationUI/AzureCommunicationUI/Presentation/Manager/AudioSessionManager.swift
@@ -14,16 +14,28 @@ class AppAudioSessionManager: AudioSessionManager {
     private let logger: Logger
     private let store: Store<AppState>
     var cancellables = Set<AnyCancellable>()
+    var isSpeakerphoneOn: Bool = false
 
     init(store: Store<AppState>,
          logger: Logger) {
         self.store = store
         self.logger = logger
+        initializeAudioDeviceState()
         self.setupAudioSession()
         store.$state
             .sink { [weak self] state in
                 self?.receive(state: state)
             }.store(in: &cancellables)
+
+    }
+
+    private func initializeAudioDeviceState() {
+        isSpeakerphoneOn = getCurrentAudioDevice() == .speaker
+        if isSpeakerphoneOn {
+             store.dispatch(action: LocalUserAction.AudioDeviceChangeSucceeded(device: .speaker))
+        } else {
+             store.dispatch(action: LocalUserAction.AudioDeviceChangeSucceeded(device: .receiver))
+        }
     }
 
     private func receive(state: AppState) {
@@ -50,6 +62,7 @@ class AppAudioSessionManager: AudioSessionManager {
                                                            .interruptSpokenAudioAndMixWithOthers,
                                                            .allowBluetoothA2DP]
             try audioSession.setCategory(.playAndRecord, mode: .voiceChat, options: options)
+            try audioSession.overrideOutputAudioPort(isSpeakerphoneOn ? .speaker : .none)
             try audioSession.setActive(true)
         } catch let error {
             logger.error("Failed to set audio session category:\(error.localizedDescription)")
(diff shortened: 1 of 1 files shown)


AzureCommunicationUI/AzureCommunicationUI/Presentation/Manager/AudioSessionManager.swift
```diff
@@ -14,16 +14,28 @@ class AppAudioSessionManager: AudioSessionManager {
     private let logger: Logger
     private let store: Store<AppState>
     var cancellables = Set<AnyCancellable>()
+    var isSpeakerphoneOn: Bool = false
```
AmyL219, line 17: Do we need this local variable here? We need it because we want to use this varibale inside `setupAudioSession`?

AzureCommunicationUI/AzureCommunicationUI/Presentation/Manager/AudioSessionManager.swift
```diff
@@ -50,6 +62,7 @@ class AppAudioSessionManager: AudioSessionManager {
                                                            .interruptSpokenAudioAndMixWithOthers,
                                                            .allowBluetoothA2DP]
             try audioSession.setCategory(.playAndRecord, mode: .voiceChat, options: options)
+            try audioSession.overrideOutputAudioPort(isSpeakerphoneOn ? .speaker : .none)
```
AmyL219, line 65: Do we need to override the  output audioport?

AzureCommunicationUI/AzureCommunicationUI/Presentation/Manager/AudioSessionManager.swift
AmyL219: Just want to confirm why the audio is override with `.receiver` when it's re-launched is because: 
1. We setup AudioSession when the audio session manager init and this would reset `.speaker` to `.receiver`. 
2. The default value for `LocalUserState.AudioDeviceSelectionStatus` is `receiverSelected` when the composite is initialized. 

To fix this, we have to get the user's previous selected device before setup AudioSession. And set the value after audioSession is active.

AzureCommunicationUI/AzureCommunicationUI/Presentation/Manager/AudioSessionManager.swift
```diff
@@ -14,16 +14,28 @@ class AppAudioSessionManager: AudioSessionManager {
...
     private let store: Store<AppState>
     var cancellables = Set<AnyCancellable>()
+    var isSpeakerphoneOn: Bool = false
 
     init(store: Store<AppState>,
          logger: Logger) {
         self.store = store
         self.logger = logger
+        initializeAudioDeviceState()
         self.setupAudioSession()
```
AmyL219, line 24: Wonder if we should send audio device `request` action instead of `changeSucceeded` action here. 
```
 init(store: Store<AppState>,
         logger: Logger) {
        self.store = store
        self.logger = logger
        let currentAudioDevice = getCurrentAudioDevice() 
        self.setupAudioSession()
        store.dispatch(action: LocalUserAction.AudioDeviceChangeSucceeded(device: currentAudioDevice))
.....
```

And remove `overrideOutputAudioPort` inside `setupAudioSession()` and `isSpeakerphoneOn`

palatter:
Missing the [bug] prefix in the title and the 'bug' tag.

AzureCommunicationUI/AzureCommunicationUI/Presentation/Manager/AudioSessionManager.swift
kevinyulu: > Do we need this local variable here? We need it because we want to use this varibale inside `setupAudioSession`?

Yes I think this `isSpeakerphoneOn` is to tell app to switch to the last audio option in `setupAudioSession`. Since this `getCurrentAudioDevice()` would be reset before and after the call `setupAudioSession()`.

AzureCommunicationUI/AzureCommunicationUI/Presentation/Manager/AudioSessionManager.swift
kevinyulu: @AmyL219 to support the comment you mentioned earlier:

> We setup AudioSession when the audio session manager init and this would reset .speaker to .receiver.
Basically, app will lose the previous the audio option from previous call, once `setupAudioSession()` is called, because after calling setupAudioSession(), the value `.receiver` would be the audio option. Hence, dispatching the `LocalUserAction.AudioDeviceChangeSucceeded` after `setupAudioSession()` would always set audio option to `.receiver`, instead of the previous audio option. 

With this regard, I think we'd need to use `audioSession.overrideOutputAudioPort()` in `setupAudioSession()` so that the app would honour the previous audio option from the previous call.

AzureCommunicationUI/AzureCommunicationUI/Presentation/Manager/AudioSessionManager.swift
```diff
@@ -14,16 +14,28 @@ class AppAudioSessionManager: AudioSessionManager {
...
+
+    }
+
+    private func initializeAudioDeviceState() {
+        isSpeakerphoneOn = getCurrentAudioDevice() == .speaker
+        if isSpeakerphoneOn {
+             store.dispatch(action: LocalUserAction.AudioDeviceChangeSucceeded(device: .speaker))
+        } else {
+             store.dispatch(action: LocalUserAction.AudioDeviceChangeSucceeded(device: .receiver))
+        }
```
kevinyulu, line 38: this code is referred from Android. 
Android initially has the same default LocalUserAction state, as iOS. What Android also does is to check the current audio status from Android OS `AudioManager` and conditionally dispatch the LocalUserAction above based on the value from `AudioManager`.

AzureCommunicationUI/AzureCommunicationUI/Presentation/Manager/AudioSessionManager.swift
AmyL219: Oh, my bad in my code example, the action should be dispatch is `requested`: `AudioDeviceChangeRequested`, something like:
```
init(store: Store<AppState>,
         logger: Logger) {
        self.store = store
        self.logger = logger
        let currentAudioDevice = getCurrentAudioDevice() 
        self.setupAudioSession()
        store.dispatch(action: LocalUserAction.AudioDeviceChangeRequested(device: currentAudioDevice))
.....

```
This would call `switchAudioDevice` function and dispatch  `AudioDeviceChangeSucceed` function after Redux store has received `AudioDeviceChangeRequested` and state updated. 
The main benefit is it decouples the `setupAudioSession` with `overrideOutputAudioPort`, and without adding extra logic to send  `AudioDeviceChangeSucceed` action on the initialization.

AmyL219: But If we want to directly call `audioSession.overrideOutputAudioPort()` inside `setupAudioSession` during AudioSessionManger initialization, we still can change the `isSpeakerphoneOn` to an parameter since it's only used inside `setupAudioSession`, correct? something like
```
init(store: Store<AppState>,
         logger: Logger) {
        self.store = store
        self.logger = logger
        let currentAudioDevice = getCurrentAudioDevice() 
        self.setupAudioSession(previousDevice: currentAudioDevice)
        store.dispatch(action: LocalUserAction.AudioDeviceChangeSucceed(device: currentAudioDevice))
.....

```

kevinyulu: @AmyL219 noted, thanks for clarifying. yes I'd vote for having a local `currentAudioDevice` to keep track of the last audio option, then use it in the dispatch
` LocalUserAction.AudioDeviceChangeRequested(device: currentAudioDevice))`
I will try and update PR.

AzureCommunicationUI/AzureCommunicationUI/Presentation/Manager/AudioSessionManager.swift
```diff
@@ -19,11 +19,14 @@ class AppAudioSessionManager: AudioSessionManager {
...
         self.store = store
         self.logger = logger
+        let currentAudioDevice = getCurrentAudioDevice()
         self.setupAudioSession()
+        store.dispatch(action: LocalUserAction.AudioDeviceChangeRequested(device: currentAudioDevice))
         store.$state
             .sink { [weak self] state in
                 self?.receive(state: state)
             }.store(in: &cancellables)
+
```
JoshuaLai, line 29: Nitpick: no need to have the extra line.

JoshuaLai (dismissed):
Tested and looks good to me

AmyL219 (dismissed):

AzureCommunicationUI/AzureCommunicationUI/Presentation/Manager/AudioSessionManager.swift
kevinyulu: changed done

AzureCommunicationUI/AzureCommunicationUI/Presentation/Manager/AudioSessionManager.swift
kevinyulu: no longer needed this logic

AzureCommunicationUI/AzureCommunicationUI/Presentation/Manager/AudioSessionManager.swift
kevinyulu: done

AzureCommunicationUI/AzureCommunicationUI/Presentation/Manager/AudioSessionManager.swift
kevinyulu: the override is removed

AmyL219 (approved):

AmyL219 (commented):
The new implementation looks quite clearer

JoshuaLai (approved):

jimchou-dev (approved):
Tested, LGTM

PR #23, opened 2021-12-22
[Refactor] Update cameraStatus after triggering camera preview 

AmyL219:
## Purpose

* Bug: the local camera preview in setup view failed to open if the `localUserState.cameraState.operation` is updated to `.pending`
* The causes of the bug are:
1.  whenever the camera permission status is updated to `.granted` (when the composite is initialized in this bug scenario), we dispatch `CameraPreviewOnTriggered` action. 
2. However, this action is dispatched before `deviceManager` created inside `CallingSDK`. So this action is always failed and useless. 
3. The actual action open the setup view's camera preview is another `CameraPreviewOnTriggered` action triggered after `setupCall`, but we only send this action when the `cameraState.operation` is `off` which indicates there is no pending action in opening the local camera. 
  
* The fix is:

1. When camera permission status update to `.granted`, we check the current `cameraPermission`, only when the status is `requesting` which indicates that the `.granted` permission is newly requested from users ( like users first time tap the camera button, the app would ask camera permission and change status to `requesting`), we would dispatch `CameraPreviewOnTriggered` action. Otherwise, it just updates permission status without any following actions. 
2. Update `cameraState.operation` to `.pending` when the action `CameraPreviewOnTriggered` is received inside Redux
3. Move the logic of sending `CameraPreviewOnTriggered` inside `setupCall` completion closure to receiveValue closure. The reason is the completion closure is used to indicated the subscription is finished, and it doesn't comes with any values. Because the `setupCall` function return `Void` value, the composite would behave same no matter putting codes in completion or receiveValue closure. But by taking the [reference from Apple](https://developer.apple.com/documentation/combine/using-combine-for-your-app-s-asynchronous-code), receiveValue closure is the right place to put the follow up action. 

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

1. Test local camera preview on setup view
2. Test ask camera permission on setup/calling view


kevinyulu (dismissed):
tested the two scenarios listed in the PR. looks good

jimchou-dev (dismissed):
Tested, LGTM

palatter (approved):
Looks good and tested.

jimchou-dev (approved):

PR #27, opened 2022-01-11
[Refactor] Bundle Short Version String Update

AmyL219:
## Purpose

* Remove alphabet in CFBundleShortVersionString
* Use a static version string including alphabet when sending the tags to Telemetry

Based on the Apple [requirement](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleshortversionstring), only `numeric characters (0-9) and periods` are allowed in CFBundleShortVersionString. `Cocoapods` would trim the CFBundleShortVersionString to make it only contains number when customers install our Library. However, the `DiagnosticConfig.swift` which is reading CFBundleShortVersionString would not be accurate when the library is installed through Cocoapods and the version got trimmed. So we have to use a static version string (like [CallingSDK](https://skype.visualstudio.com/SCC/_git/client_calling_spool-native?path=%2FlibShared%2Fsrc%2FPAL.cpp&version=GBmaster&_a=contents)) when sending the versioning to Telemetry, and not dynamically rely on reading CFBundleShortVersionString.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

### DIFF
--- a/AzureCommunicationUI/AzureCommunicationUI.xcodeproj/project.pbxproj
+++ b/AzureCommunicationUI/AzureCommunicationUI.xcodeproj/project.pbxproj
@@ -1654,7 +1654,7 @@
 					"@executable_path/Frameworks",
 					"@loader_path/Frameworks",
 				);
-				MARKETING_VERSION = "1.0.0-alpha.2";
+				MARKETING_VERSION = 1.0.0;
 				PRODUCT_BUNDLE_IDENTIFIER = com.azure.ios.communication.ui.meetings.AzureCommunicationUI;
 				PRODUCT_NAME = "$(TARGET_NAME:c99extidentifier)";
 				SKIP_INSTALL = YES;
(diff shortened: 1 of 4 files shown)


AmyL219:
Any thought if we should set up another environment value (like UICompositeVersionString) inside `info.plist` to store the value of `1.0.0-beta2`. 
The benefit of it is we can easily automate the version increase in CI if needed

palatter:
> Any thought if we should set up another environment value (like UICompositeVersionString) inside `info.plist` to store the value of `1.0.0-beta2`. The benefit of it is we can easily automate the version increase in CI if needed

That's a good idea!

JoshuaLai (approved):

cheukchuen (approved):

kevinyulu (approved):

AmyL219:
Tested with Telemetry board. https://dataexplorer.azure.com/clusters/kusto.aria.microsoft.com/databases/4cf9ed87f7dc4e148f77855cd1f2fdca?query=H4sIAAAAAAAAA0XMsQ7CIBAA0N2vuA0wAQG/wMHB3Z2ceLYkeDTlKosfbzfXN7yNS2PA3FPGWgtP6QiHL4yZVoLrh1hu/GrpXt4ED5JBxKClPVFIdtMq+hhsiDYGZZzjNrQx/2HrtF6mfYHcWLBwB4W5hOBPwXnnLdZlRndWP1fn9JGIAAAA

PR #29, opened 2022-01-13
[Feature] Supporting pinch, zoom, pan gestures and "double tap to zoom" in video screen share view in a call

kevinyulu:
## Purpose
In this PR I drafted the feature supporting user pinch, zoom and pan gestures in video screen share in a call. The functionalities resembles the MS Teams application UX.
Additional features introduced:
* double tap to zoom in / zoom out
* only zooming the actual video content, not the entire rendered view from SDK
* Improved screen share video quality in iPhone.

https://user-images.githubusercontent.com/90668345/149402996-ff02c754-88ae-4d9a-b1dd-4bacf8c96a94.mov

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code
* Join a call either using group or Teams call, in iOS app and Teams application (desktop)
* Turn on screenshare in Teams application (desktop)
* Start to perform pinch, zoom and pan gesture on the screen share video 
* Follow **What to Check section**
```

## What to Check
* Check if the pinch, zoom and pan gestures works in screen share video, in all supported orientations during a call
* Check if double tap to zoom in / out is working as well.

## Other Information
<!-- Add any other helpful information that may be needed here. -->

### CODE REVIEW

File: AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/Cell/ParticipantGridCellView.swift [hunk 1]
```diff
@@ -10,6 +10,7 @@ import Combine
 struct ParticipantGridCellView: View {
     @ObservedObject var viewModel: ParticipantGridCellViewModel
     let getRemoteParticipantRendererView: (RemoteParticipantVideoViewId) -> UIView?
+    let getRemoteParticipantRenderViewStreamSize: (RemoteParticipantVideoViewId) -> CGSize?
```
AmyL219, line 13: I wonder if it's necessary to have a separated function to get the videoStreamSize?
Could we just get size from the function `getRemoteParticipantRendererView: (RemoteParticipantVideoViewId) -> (UIView, CGSize)?`
When we return the rendererView, we can just also pass the CGSize of the renderer from the videoStreamCache which is an easy calculation. And this can make code cleaner and easier for read

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/Cell/ParticipantGridCellViewModel.swift
```diff
@@ -5,11 +5,13 @@
 
 import Foundation
 import Combine
+import AzureCommunicationCalling
```
AmyL219, line 8: We are not going to couple the viewModel with `AzureCommunicationCalling` SDK. Could we remove the import here?

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/Cell/ZoomableVideoRenderView.swift
```diff
@@ -0,0 +1,112 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import Foundation
+import SwiftUI
+
+struct ZoomableVideoRenderView: UIViewRepresentable {
```
AmyL219, line 9: Curious about if CallingSDK will support zooming in the video stream view? Or that's need to be done in the UI level by ourself?

kevinyulu: As far as I know, the `RendererView` from CallingSDK is not capable of supporting user zoom nor panning. We'll need to either check with the SDK team to provide support for that or supporting by ourself.

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/Cell/ParticipantGridCellViewModel.swift
kevinyulu: yes I think we can use an enum to represent the scaling mode.

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/Cell/ParticipantGridCellView.swift
kevinyulu: yeah this is a very good point. Actually this was what I had tried with Inderpal as well initially. So far we've found that the size info won't be ready till sometime **after** the `RenderView` has been displayed on the device. This has made me think to create a callback like this, to get a actual size later time, when user starts to zoom the RenderView.

kevinyulu: Alternatively I am checking if we can ask the SDK team to provide the size info for the video so we'd skip creating extra functions to get the size info.

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/Cell/ParticipantGridCellViewModel.swift
kevinyulu: removed `AzureCommunicationCalling` dependency

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/Cell/ParticipantGridCellVideoView.swift
```diff
@@ -6,20 +6,40 @@
 import SwiftUI
 import FluentUI
 import Combine
+import AzureCommunicationCalling
```
waynemo, line 9: Are we able to encapsulate all API calls and types related to the Calling SDK so that they are all within the VideoViewManager? VideoViewManager would only return UIViews so that only it needs to reference AzureCommunicationCalling.

waynemo, line 9: Are we able to encapsulate all API calls and types related to the Calling SDK so that they are all within the VideoViewManager? VideoViewManager would only return UIViews so that only it needs to reference AzureCommunicationCalling.

waynemo (changes requested):

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/Cell/ParticipantGridCellVideoView.swift
kevinyulu: > Are we able to encapsulate all API calls and types related to the Calling SDK so that they are all within the VideoViewManager? VideoViewManager would only return UIViews so that only it needs to reference AzureCommunicationCalling.

The reason of importing AzureCommunicationCalling into this view is to pass the reference of the VideoStreamRenderer to `ZoomableVideoRenderView` (screenshare) so that the `ZoomableVideoRenderView` can become a delegate of the VideoStreamRenderer (screenshare) and respond to video frame updates. I can try to make the VideoViewManager become a delegate of the VideoStreamRenderer so the views will just have to import 
AzureCommunicationCalling

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/Cell/ZoomableVideoRenderView.swift
```diff
@@ -0,0 +1,223 @@
...
+        return scrollView
+    }
+
+    func makeCoordinator() -> Coordinator {
+        Coordinator(self, shouldShowScaleForiPad: isiPadScreen())
+    }
+
+    func updateUIView(_ uiView: UIScrollView, context: Context) {
+        // Set zoom scale per each orientation to fix the zoom issue (blury screen upon orientation change)
+        let orientationChangeZoomScale = isiPadScreen() ? Constants.minScaleiPad : Constants.minScaleiPhone
```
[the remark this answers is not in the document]
kevinyulu: yes, the iPad scale needs to be different than iPhone. 

Context: the renderer view SDK calculate the size of the screenshare video based on the iOS hosting view. The size of the screenshare video will determine the quality of the video. The view/device screen, the larger the renderer view size SDK would return, (the better the video quality of the screen share it can have). For iPhone, given the limited physical dimension of the screen, the video quality returned is blurry and that's why the scale in iPhone needs to be different from iPad.

Given the large physical dimension of the device screen, we can keep the scale of iPad at (min 1, max 4).

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/Cell/ParticipantGridCellVideoView.swift
AmyL219: Instead of having `getRemoteParticipantScreenShareVideoStreamRenderer` and `getRemoteParticipantRendererView` could we combine these two closure by returning `VideoStreamCache`?

AzureCommunicationUI/AzureCommunicationUI/Presentation/VideoViewManager.swift
```diff
@@ -113,6 +113,26 @@ class VideoViewManager {
...
     }
 
+    func getScreenShareVideoStreamRenderer(_ videoViewId: RemoteParticipantVideoViewId) -> VideoStreamRenderer? {
+        let videoStreamId = videoViewId.videoStreamIdentifier
+        let cacheKey = generateCacheKey(userIdentifier: videoViewId.userIdentifier,
+                                        videoStreamId: videoStreamId)
+        return displayedRemoteParticipantsRendererView.value(forKey: cacheKey)?.renderer
+    }
+
+    func getRemoteParticipantVideoRendererViewSize(_ videoViewId: RemoteParticipantVideoViewId) -> CGSize? {
```
AmyL219, line 123: Does this function being called anywhere?

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/Cell/ZoomableVideoRenderView.swift
```diff
@@ -0,0 +1,223 @@
...
+        }
+
+        // MARK: RendererDelegate
+
+        func videoStreamRenderer(didRenderFirstFrame renderer: VideoStreamRenderer) {
+            streamSize = CGSize(width: Int(renderer.size.width), height: Int(renderer.size.height))
+            rendererView.updateRendererViewSize()
+        }
+
+        func videoStreamRenderer(didFailToStart renderer: VideoStreamRenderer) {
```
AmyL219, line 219: Should we put a log here?

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/Cell/ZoomableVideoRenderView.swift
```diff
@@ -0,0 +1,223 @@
...
+            let finalScale = (currentScale == minScale) ? toScale : minScale
+
+            rendererView.zoomScrollView(basedOn: point, scale: finalScale)
+        }
+
+        // MARK: RendererDelegate
+
+        func videoStreamRenderer(didRenderFirstFrame renderer: VideoStreamRenderer) {
+            streamSize = CGSize(width: Int(renderer.size.width), height: Int(renderer.size.height))
+            rendererView.updateRendererViewSize()
```
AmyL219, line 216: Curious about when this function would be called?

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/Cell/ParticipantGridCellViewModel.swift
```diff
@@ -50,4 +57,11 @@ class ParticipantGridCellViewModel: ObservableObject, Identifiable {
         let cameraVideoStreamIdentifier = participantModel.cameraVideoStreamModel?.videoStreamIdentifier
         return screenShareVideoStreamIdentifier ?? cameraVideoStreamIdentifier
     }
+
+    private func getDisplayingVideoStreamType(
```
AmyL219, line 61: Could we combine this function with `getDisplayingVideoStreamId` to some function like 
```
getDisplayingVideoStreamIdAndType(...) -> (Stirng?, VideoStreamInfoModel.MediaStreamType?)
```

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/Cell/ParticipantGridCellVideoView.swift
```diff
@@ -35,8 +55,21 @@ struct ParticipantGridCellVideoView: View {
...
     }
 
+    var videoRenderView: some View {
+        VideoRendererView(rendererView: self.rendererView)
+    }
+
+    var zoomableVideoRenderView: some View {
+        ZoomableVideoRenderView(getRemoteParticipantScreenShareVideoStreamRenderer:
+                                    getRemoteParticipantScreenShareVideoStreamRenderer,
+                                rendererView: self.rendererView).gesture(TapGesture(count: 2).onEnded({}))
```
AmyL219, line 70: `onEnded` is waiting for update in next PR, correct?

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/Cell/ParticipantGridCellVideoView.swift
```diff
@@ -6,20 +6,40 @@
...
 struct ParticipantGridCellVideoView: View {
+
+    private struct Constants {
+        static let homebarHeight: CGFloat = 22
+        static let borderColor = Color(StyleProvider.color.primaryColor)
+    }
+
     let rendererView: UIView
+    let getRemoteParticipantScreenShareVideoStreamRenderer: () -> VideoStreamRenderer?
+    let zoomable: Bool
```
AmyL219, line 20: Should zoomable also be the type @Binding through? in case the remote participant update to screenshare from local video stream? Probably change the property to media stream type

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/Cell/ParticipantGridCellVideoView.swift
```diff
@@ -6,20 +6,40 @@
...
+                                  && UIDevice.current.hasHomeBar)
         ZStack(alignment: .bottomLeading) {
             VStack(alignment: .center, spacing: 0) {
-                VideoRendererView(rendererView: rendererView)
+                GeometryReader { geometry in
+                    if zoomable {
+                        // reduce height as work-around to resolve the double tap issue, when lanscapeHasHomeBar is true
+                        // To be improved in the next PR
+                        zoomableVideoRenderView
+                            .frame(width: geometry.size.width - 0,
```
AmyL219, line 36: ` - 0` here? Could we remove it then

AzureCommunicationUI/AzureCommunicationUI/Presentation/VideoViewManager.swift
kevinyulu: I think this was called before the refactor, it is removed now.

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/Cell/ParticipantGridCellVideoView.swift
kevinyulu: yes this is a work around to resolve the double tap / single tap interference. I am going to look into ways to improve in the next PR (another user story)

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/Cell/ParticipantGridCellVideoView.swift
kevinyulu: yes, I will remove

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/Cell/ZoomableVideoRenderView.swift
kevinyulu: sure

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/Cell/ParticipantGridCellVideoView.swift
kevinyulu: Yes, returning VideoStreamCache would be one way for the `ZoomableVideoRenderView` to be set as the **delegate** of the VideoStreamRenderer. One thing I'd concern is that to set this **delegate** (which is of type `RendererDelegate`), in `ZoomableVideoRenderView`, `ZoomableVideoRenderView` has to import reference to AzureCommunicationCalling. Are we trying to avoid introducing AzureCommunicationCalling references into our view?

To avoid having to refer to `AzureCommunicationCalling` in the Swift UI views,I added a delegate `VideoScreenShareDelegate` which has pretty much same function from the RendererDelegate from SDK, and made the child view become the delegate to the `VideoViewManager`.

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/Cell/ParticipantGridCellViewModel.swift
kevinyulu: yes, updated the code with the suggested changes

AzureCommunicationUI/AzureCommunicationUI/Presentation/VideoViewManager.swift
```diff
@@ -113,6 +123,13 @@ class VideoViewManager {
 
     }
 
+    func getScreenShareVideoStreamRenderer(_ videoViewId: RemoteParticipantVideoViewId) -> VideoStreamRenderer? {
```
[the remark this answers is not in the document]
kevinyulu: done

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/Cell/ParticipantGridCellView.swift
```diff
@@ -13,6 +13,7 @@ struct ParticipantGridCellView: View {
     @State var displayedVideoStreamId: String?
     @State var isVideoChanging: Bool = false
     let avatarSize: CGFloat = 56
+    let videoViewManager: VideoViewManager
```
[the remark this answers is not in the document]
kevinyulu: I had to use a class instead of a struct here. This `VideoScreenShareDelegate` will be set to `context.coordinator` in the `func makeUIView(context: Context)` in `ZoomableVideoRenderView`, and declaring this `ParticipantRendererViewWithDelegate` as set would require us to change this function to a `mutating` type, which will make the `ZoomableVideoRenderView` not conforming to `UIViewRepresentable` anymore.  
also would it be good to rename this `ParticipantRendererViewWithDelegate` to `ParticipantRendererViewInfo`?

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/CallingViewModel.swift
```diff
@@ -101,6 +101,10 @@ class CallingViewModel: ObservableObject {
 
     func receive(_ state: AppState) {
         guard state.lifeCycleState.currentStatus == .foreground else {
+            if state.lifeCycleState.currentStatus == .background {
```
AmyL219, line 104: Do we need an extra check here?
the currentStatus can be either `foreground` or `background` though?
if state.lifeCycleState.currentStatus != .foreground, then state.lifeCycleState.currentStatus should be background

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/Cell/ZoomableVideoRenderView.swift
```diff
@@ -0,0 +1,283 @@
...
+        rendererView!.frame = scrollView.bounds
+        scrollView.addSubview(rendererView!)
+        scrollView.contentSize = rendererView.bounds.size
+
+        // Double tap action
+        let doubleTapGesture = UITapGestureRecognizer(target: context.coordinator,
+                                                      action: #selector(Coordinator.doubleTapped))
+        doubleTapGesture.numberOfTapsRequired = Constants.maxTapRequired
+        doubleTapGesture.delegate = context.coordinator
+        videoRendererViewInfo.delegate = context.coordinator
```
AmyL219, line 54: Need to remove the line 54 here. Otherwiese I can't run the branch code.

AzureCommunicationUI/AzureCommunicationUI/Presentation/VideoViewManager.swift
```diff
@@ -11,17 +11,38 @@ struct RemoteParticipantVideoViewId {
     let videoStreamIdentifier: String
 }
 
-class VideoViewManager {
+class ParticipantRendererViewInfo {
+    let rendererView: UIView
+    var streamSize: CGSize
+    weak var videoManager: VideoViewManager?
```
AmyL219, line 17: Thanks for taking my suggestion to use a struct to wrap the needed properties. 
But if we have to wrap the VideoViewManager inside the struct, I think it makes more sense to directly pass the a protocol all the way down to the `ParticipantGridView`, `ParticipantGridLayoutView`, `ParticipantGridCellVideoView`. Replace the `getRemoteParticipantRendererView` with the protocol, and the `videoViewManager` is the protocol implementation. Reasons are:  

- This can simplify the videoViewManager function and makes the code clean and easy for maintenance. 
- From performance side, reading the properties from closure or from manager shouldn't have difference. I think the original idea is to isolate view from knowing videoViewManager. To achieve the same propose, we can have a protocol, , and the `videoViewManager` is the protocol implementation. like 
```
protocol ParticipantRendererViewManager: AnyObject {
getRemoteParticipantVideoRendererView() -> RendererView
getRemoteParticipantVideoRenderer() -> Render
getRemoteParticipantVideSize() -> CGSize
getRemoteParticipantVideoDelegate() -> Delegate (or a property)
}
```
and pass the protocol object to view layer with `getRemoteParticipantRendererView`

AzureCommunicationUI/AzureCommunicationUI/Presentation/VideoViewManager.swift
```diff
@@ -101,16 +127,32 @@ class VideoViewManager {
...
             let cache = VideoStreamCache(renderer: newRenderer,
-                                         rendererView: newRendererView)
+                                         rendererView: newRendererView,
+                                         mediaStreamType: videoStream.mediaStreamType)
             displayedRemoteParticipantsRendererView.append(forKey: cacheKey,
                                                            value: cache)
 
-            return newRendererView
+            if videoStream.mediaStreamType == .screenSharing {
+                newRenderer.delegate = self
```
AmyL219, line 136: I know it works for now, but what's in future we need to handle multiple delegate from different views? Ideally it's more recommenned to bind the delegate with the view containing the renderer view, because the view is the final consumer from the renderer updates. 
But I think for now, this should be good.

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/Cell/ZoomableVideoRenderView.swift
kevinyulu: yes, I just removed line 54

AzureCommunicationUI/AzureCommunicationUI/Presentation/VideoViewManager.swift
kevinyulu: Agreed. Binding the SDK delegate (RendererDelegate) with the child view would introduce the `AzureCommunicationCalling` reference into the child view as well though. I think the previous approach from this PR had this binding and we removed that to avoid in introducing the `AzureCommunicationCalling` reference into the child view (`ZoomableVideoRenderView`)

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/CallingViewModel.swift
kevinyulu: I think the `lifeCycleState.currenStatus` has two values `.foreground` or `.background`. if the `lifeCycleState.currenStatus` is not .foreground then it will go to the else case in the guard where participantGridsViewModel will be updated. What additional check are you referring to @AmyL219?

AzureCommunicationUI/AzureCommunicationUI/Presentation/VideoViewManager.swift
kevinyulu: I can try that option. `RendererView` and `Render` are Types from SDK. We'll need references to `AzureCommunicationCalling` in our view that calls the function from the ParticipantRendererViewManager object.  Is that intended?

kevinyulu: we might to modify the methods from the protocol ParticipantRendererViewManager to only return non SDK types: 
```
protocol ParticipantRendererViewManager: AnyObject {
getRemoteParticipantVideoRendererView() -> UIView?
getRemoteParticipantVideSize() -> CGSize
getRemoteParticipantVideoDelegate() -> a custom delegate, not `RendererDelegate` from SDK  (or a property)
}
```

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/Cell/ParticipantGridCellViewModel.swift
```diff
@@ -45,9 +52,14 @@ class ParticipantGridCellViewModel: ObservableObject, Identifiable {
         }
     }
 
-    private func getDisplayingVideoStreamId(_ participantModel: ParticipantInfoModel) -> String? {
+    private func getDisplayingVideoStreamIdAndType(_ participantModel: ParticipantInfoModel) ->
```
waynemo, line 55: This would be more readable if it returned some kind of type instead of a tuple. 

On the Android side, this view model publishes a `VideoViewModel`. Could we do the same, where `videoStreamId` and `videoStreamType` are properties of `VideoViewModel`?

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/ParticipantGridViewModel.swift
```diff
@@ -14,13 +14,17 @@ class ParticipantGridViewModel: ObservableObject {
...
     @Published var gridsCount: Int = 0
     @Published var displayedParticipantInfoModelArr: [ParticipantInfoModel] = []
+    @Published var isAppInForeground: Bool = true
 
     init(compositeViewModelFactory: CompositeViewModelFactory) {
         self.compositeViewModelFactory = compositeViewModelFactory
     }
 
-    func update(remoteParticipantsState: RemoteParticipantsState) {
+    func update(remoteParticipantsState: RemoteParticipantsState, lifeCycleState: LifeCycleState) {
```
waynemo, line 23: I don't think it is necessary to pass the `LifeCycleState` as part of the update just so we can pass `isAppInForeground` all the way to the `ZoomableVideoRenderView`.

Instead, we could expose a method on `AppLifeCycleManager` to get the current foreground/background state. `AppLifeCycleManager` can then be injected into `ZoomableVideoRenderView` and called to check the foreground/background state for rendering.

waynemo (dismissed):

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/Cell/ParticipantGridCellViewModel.swift
```diff
@@ -10,6 +10,7 @@ class ParticipantGridCellViewModel: ObservableObject, Identifiable {
     let id = UUID()
 
     @Published var videoStreamId: String?
+    @Published var videoStreamType: VideoStreamInfoModel.MediaStreamType?
```
kevinyulu, line 13: instead of publishing videoStreamId, and videoStreamType, consider publishing a new type that wrap these two attributes. Referred to [ https://github.com/Azure/communication-ui-library-android/blob/develop/beta/azure-communication-ui/azure-communication-ui/src/main/java/com/azure/android/communication/ui/presentation/fragment/calling/participant/grid/ParticipantGridCellViewModel.kt#L48](url)

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/ParticipantGridViewModel.swift
kevinyulu: to use environment variable that carry the foreground/background info from AppLifeCycleManager

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/Cell/ParticipantGridCellViewModel.swift
kevinyulu: added

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/CallingViewModel.swift
AmyL219: I mean does this work the same?
```
 guard state.lifeCycleState.currentStatus == .foreground else {

participantGridsViewModel.update(remoteParticipantsState: state.remoteParticipantsState,
                                                 lifeCycleState: state.lifeCycleState)
return
}
```

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/Cell/ParticipantGridCellViewModel.swift
```diff
@@ -6,10 +6,20 @@
 import Foundation
 import Combine
 
+class VideoViewModel {
```
AmyL219, line 9: Should this be a `Struct`?

AmyL219: https://forums.raywenderlich.com/t/question-about-struct-vs-class/61980

AmyL219:
There is a bug that the screensharing view wouldn't be display, please try the reproduce steps:
1. Join a team's call **with the remote participant open his camera**
2. Put the device in **landscape** mode after being admitted
3. Start screen sharing inside the Teams client side
4. You would see the grey view as the screen sharing view
MicrosoftTeams-image (25)

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/CallingViewModel.swift
JoshuaLai: Agree with @AmyL219 I think the `if` is a redundant check and can be removed.

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/Cell/ParticipantGridCellViewModel.swift
JoshuaLai: If we are moving this to a struct(which I think you can do, it doesn't look like you are holding onto state and nothing is mutating) lets make sure to change the properties from `var` to `let`

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/Cell/ZoomableVideoRenderView.swift
```diff
@@ -0,0 +1,275 @@
...
+        guard scrollView.frame != .zero else {
+            return
+        }
+        let adjustedScale = isiPadScreen() ? Constants.minScaleiPad : Constants.minScaleiPhone
+        let currentScale = scrollView.zoomScale
+        guard adjustedScale != currentScale, zoomToRect != .zero else {
+            return
+        }
+        updateRendererViewSize()
+        DispatchQueue.main.asyncAfter(deadline: DispatchTime.now() + 0) {
```
AmyL219, line 144: Could we update to 
```
 DispatchQueue.main.async {
}
```
Since there no usage of delay?

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/CallingViewModel.swift
```diff
@@ -38,6 +39,7 @@ class CallingViewModel: ObservableObject {
         let isCallConnected = store.state.callingState.status == .connected
         let hasRemoteParticipants = store.state.remoteParticipantsState.participantInfoList.count > 0
         isParticipantGridDisplayed = isCallConnected && hasRemoteParticipants
+        appState = store.state.lifeCycleState.currentStatus == .foreground ? .foreground : .background
```
AmyL219, line 42: Would this work?
`appState = state.lifeCycleState.currentStatus`

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/Cell/ParticipantGridCellViewModel.swift
```diff
@@ -6,10 +6,15 @@
 import Foundation
 import Combine
 
+struct VideoViewModel {
```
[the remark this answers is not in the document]
kevinyulu: sure

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/Cell/ZoomableVideoRenderView.swift
kevinyulu: for some reason DispatchQueue.main.async would not work. I ended up using the `DispatchQueue.main.asyncAfter` as a workaround

AmyL219 (dismissed):
Let's create bug tickets for the remaining issues. Be free the bug to me if helps needed

kevinyulu:
For the bugs @AmyL219 @palatter relayed, I am going to create new bug ticket.

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/Cell/ZoomableVideoRenderView.swift
```diff
@@ -0,0 +1,275 @@
...
+    func makeUIView(context: Context) -> UIScrollView {
+        // Setup scrollview and renderview
+        scrollView.delegate = context.coordinator
+        scrollView.maximumZoomScale = isiPadScreen() ? Constants.maxScaleiPad : Constants.maxScaleiPhone
+        scrollView.minimumZoomScale = isiPadScreen() ? Constants.minScaleiPad : Constants.minScaleiPhone
+        scrollView.bouncesZoom = false
+        scrollView.showsVerticalScrollIndicator = false
+        scrollView.showsHorizontalScrollIndicator = false
+        scrollView.decelerationRate = .fast
+        scrollView.zoomScale = isiPadScreen() ? Constants.minScaleiPad : Constants.minScaleiPhone
```
vhuseinova-msft, line 43: minimumZoomScale value can be created and reused in scrollView.minimumZoomScale and scrollView.zoomScale

vhuseinova-msft: Maybe even create it as a private property as it's reused later in updateUIView(_ uiView: UIScrollView, context: Context) and zoomScrollView(basedOn point: CGPoint, scale: CGFloat) and a couple of other places

vhuseinova-msft (dismissed):
Added an minor code improvement

vhuseinova-msft (approved):

AmyL219 (approved):

PR #30, opened 2022-01-15
[Feature][DemoApp] Implement deep link to join call

jimchou-dev:
## Purpose

* Implemented feature to supporting iOS deep link to join group call or teams call. When clicking on deep link matching `acsui://` will open AzureCommunicationUIDemoApp and navigate to swiftUI scene to join call and apply parameters if present in url
  * Group call deep link = `acsui://calling?acstoken=<token>&name={displayName}&groupid={groupCallId}
  * Teams call deep link = `acsui://calling?acstoken=<token>&name={displayName}&teamsurl={teamsMeetingLink}

```
[ ] Yes
[ x ] No
```

## Pull Request Type

```
[ ] Bugfix
[ x ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

* Build and run the app on device/simulator
* Go to a browser and go to the link; or click on a send link matching the following format to start a call:
  * Group call deep link = `acsui://calling?acstoken=<token>&name={displayName}&groupid={groupCallId}
  * Teams call deep link = `acsui://calling?acstoken=<token>&name={displayName}&teamsurl={teamsMeetingLink} 
* Follow **What to Check section**

## What to Check
* Make sure the url query parameter are properly passed into the fields
* Able to start and end a call

## Other Information

* When already in another composite call and start another call from deeplink will cause the app to crash, created separate backlog task 2695632 to handle error

### DIFF
--- a/AzureCommunicationUI/AzureCommunicationUIDemoApp/Info.plist
+++ b/AzureCommunicationUI/AzureCommunicationUIDemoApp/Info.plist
@@ -2,6 +2,19 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>CFBundleURLTypes</key>
+	<array>
+		<dict>
+			<key>CFBundleTypeRole</key>
+			<string>Editor</string>
+			<key>CFBundleURLName</key>
+			<string>com.azure.ios.communication.ui.meetings.AzureCommunicationUIDemoApp</string>
+			<key>CFBundleURLSchemes</key>
+			<array>
+				<string>acsui</string>
+			</array>
+		</dict>
+	</array>
 	<key>UIApplicationSceneManifest</key>
 	<dict>
 		<key>UIApplicationSupportsMultipleScenes</key>
@@ -25,19 +38,15 @@
 	<array>
 		<string>audio</string>
 	</array>
-	<key>NSMicrophoneUsageDescription</key>
-	<string>This is required for calling.</string>
-	<key>NSCameraUsageDescription</key>
-	<string>This is required for calling.</string>
-    <key>acsToken</key>
-    <string>${acsToken}</string>
-    <key>acsTokenUrl</key>
-    <string>${acsTokenUrl}</string>
-    <key>displayName</key>
-    <string>${displayName}</string>
-    <key>groupCallId</key>
-    <string>${groupCallId}</string>
-    <key>teamsMeetingLink</key>
-    <string>${teamsMeetingLink}</string>
+	<key>acsToken</key>
+	<string>${acsToken}</string>
+	<key>acsTokenUrl</key>
+	<string>${acsTokenUrl}</string>
+	<key>displayName</key>
+	<string>${displayName}</string>
+	<key>groupCallId</key>
+	<string>${groupCallId}</string>
+	<key>teamsMeetingLink</key>
+	<string>${teamsMeetingLink}</string>
 </dict>
 </plist>
--- a/AzureCommunicationUI/AzureCommunicationUIDemoApp/Info.plist
+++ b/AzureCommunicationUI/AzureCommunicationUIDemoApp/Info.plist
@@ -15,6 +15,10 @@
 			</array>
 		</dict>
 	</array>
+	<key>NSMicrophoneUsageDescription</key>
+	<string>This is required for calling.</string>
+	<key>NSCameraUsageDescription</key>
+	<string>This is required for calling.</string>
 	<key>UIApplicationSceneManifest</key>
 	<dict>
 		<key>UIApplicationSupportsMultipleScenes</key>
--- a/AzureCommunicationUI/AzureCommunicationUIDemoApp/SceneDelegate.swift
+++ b/AzureCommunicationUI/AzureCommunicationUIDemoApp/SceneDelegate.swift
@@ -50,13 +50,29 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
         // Group call = acsui://calling?acstoken={}&name={}&groupid={}
         // Teams call = acsui://calling?acstoken={}&name={}&teamsurl={}
         let queryDict = URLContexts.first?.url.queryDictionary ?? [:]
-        var swiftUiDemoView = SwiftUIDemoView()
-        swiftUiDemoView.deepLinkValues = queryDict
-
+        lazy var deepLinkCoordinator = DeepLinkData()
+        let swiftUiDemoView = SwiftUIDemoView()
+            .environment(\.deepLinkCoordinator, deepLinkCoordinator)
+        deepLinkCoordinator.deepLinkValues = queryDict
         guard let windowScene = scene as? UIWindowScene else { return }
         let window = UIWindow(windowScene: windowScene)
         window.rootViewController = UIHostingController(rootView: swiftUiDemoView)
         self.window = window
         window.makeKeyAndVisible()
+    }
+}
+
+class DeepLinkData {
+    @Published var deepLinkValues: [String:String] = [:]
+}
+
+struct DeepLinkKey: EnvironmentKey {
+    static let defaultValue: DeepLinkData = DeepLinkData()
+}
+
+extension EnvironmentValues {
+    var deepLinkCoordinator: DeepLinkData {
+        get { self[DeepLinkKey.self] }
+        set { self[DeepLinkKey.self] = newValue }
     }
 }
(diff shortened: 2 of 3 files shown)


AzureCommunicationUI/AzureCommunicationUIDemoApp/Info.plist
```diff
@@ -25,19 +38,15 @@
 	<array>
 		<string>audio</string>
 	</array>
-	<key>NSMicrophoneUsageDescription</key>
```
JoshuaLai, line 28: Do we not need these permission descriptions anymore?

jimchou-dev: Good catch, this was accidentally removed when trying to add the CFBundleURLTypes programatically. Now they're added back =)

JoshuaLai (dismissed):

AzureCommunicationUI/AzureCommunicationUIDemoApp/SceneDelegate.swift
```diff
@@ -44,4 +45,18 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
         // to restore the scene back to its current state.
     }
 
+    func scene(_ scene: UIScene, openURLContexts URLContexts: Set<UIOpenURLContext>) {
+
+        // Group call = acsui://calling?acstoken={}&name={}&groupid={}
+        // Teams call = acsui://calling?acstoken={}&name={}&teamsurl={}
+        let queryDict = URLContexts.first?.url.queryDictionary ?? [:]
+        var swiftUiDemoView = SwiftUIDemoView()
+        swiftUiDemoView.universalLinkValues = queryDict
```
kevinyulu, line 54: A side thing in addition to the SwiftUI layout, the initial app launch offers rendering UIKit views. Should we also considering getting the query and populate the query items for UIKit version?

AzureCommunicationUI/AzureCommunicationUIDemoApp/Views/SwiftUIDemoView.swift
```diff
@@ -38,6 +39,9 @@ struct SwiftUIDemoView: View {
                 message: Text(errorMessage),
                 dismissButton: .default(Text("Dismiss")))
         }
+        .onAppear(perform: {
+            applyDeepLinkValues()
```
kevinyulu, line 43: minor: I notice that when a user leaves a call, this onAppear is also called. Should app be calling the applyDeepLinkValues() after a user leaves a call?

kevinyulu:
Tested with the following links in group (see below). The parameters are passed correctly into the SwiftUI views. One thing to notice is that, the parameters from the deep link are only passed into SwiftUI views when a user had launched the app before. If the user never run the app or has killed the app, tapping on the link would only launch the app (initial UI option screen is shown) and the parameters will not be passed into the views.

> acsui://calling?acstoken=eyJhbGciOiJSUzI1NiIsImtpZCI6IjEwMyIsIng1dCI6Ikc5WVVVTFMwdlpLQTJUNjFGM1dzYWdCdmFMbyIsInR5cCI6IkpXVCJ9.eyJza3lwZWlkIjoiYWNzOjcxZWM1OTBiLWNiYWQtNDkwYy05OWM1LWI1NzhiZGFjZGU1NF8wMDAwMDAwZi0wOTYyLTlhYTQtMDJjMy01OTNhMGQwMGJlNTIiLCJzY3AiOjE3OTIsImNzaSI6IjE2NDI0NDc1NjYiLCJleHAiOjE2NDI1MzM5NjYsImFjc1Njb3BlIjoidm9pcCIsInJlc291cmNlSWQiOiI3MWVjNTkwYi1jYmFkLTQ5MGMtOTljNS1iNTc4YmRhY2RlNTQiLCJpYXQiOjE2NDI0NDc1NjZ9.nbY7PSJklSUAejG4JVZp9SVtpjquEBwcL8m0g3dhCdRVL1ypfF3B3zVfgwKVxM2Yf2bwYEXPrEN_0s68SxtDv8N7TRFf5xEiZKjPtcdjmP8C-b-CFdObyUlP3oAzyMLHqy7GlZgf6xwijk0aS0p8BiwkYLE9KYHiJkcsuduF7Lr9nRTkR_sZo5uo0hgj_U0aGwK0cUYWfRoYrk46g1VjPh3azqLgYS9tGk9yiU7NT8yf2jQkURycMg5_bgiVwRxPASO1BwgDbXsZlehVq_ycwgqG-hrA4y14VdzUzuHpVaSfmuUyT6xFOOHWaukN3cVsiUDxtzMR_yVUB8bGVizFQg&name=someDisplayName&groupid=d047ee00-d86a-11eb-8e92-d7baf38a4ecd

AzureCommunicationUI/AzureCommunicationUIDemoApp/Views/SwiftUIDemoView.swift
jimchou-dev: I couldn't come up with a way to set the @State variable after initializing the SwiftUI view. When the view is initialized due to deep link, it will set the values in `deepLinkValues`, and applyDeepLinkValues() will check if those values exists so should not affect when user enters the view and .onAppear is called.

@kevinyulu Do you know if there is a better way to directly set the state variable after the View is created?

AzureCommunicationUI/AzureCommunicationUIDemoApp/SceneDelegate.swift
jimchou-dev: A deep link can definitely be directly launching the UIKit instead, but I just arbitrarily chose to launch the SwiftUI layout. Since the purpose is to be able to share the same deeplink `acsui://calling?acstoken={}&name={}&...` as Android in https://acs-ui-dev-tokengen.azurewebsites.net/api/AndroidDeepLink (new deep link Azure function will be created to remove the Android name since it will work on both)

And having the ability to launch SwiftUI and UIKit version are just demo app to showcase that the composite works for both SwiftUI and UIKit, so I think having deep link only launching the experience in SwiftUI is sufficient.

AzureCommunicationUI/AzureCommunicationUIDemoApp/Views/SwiftUIDemoView.swift
AmyL219: We could wrap those `acsToken`, `acsTokenUrl`, `displayName`...`teamsMeetingLink` into an environment state variable. And change the environment state when there is deep link jump received. 
See this example for SwiftUI navigation on deep link: https://nalexn.github.io/swiftui-deep-linking/

AmyL219 (commented):
Quick question, how do we handle if user already open the UIKit View, and switch to deep link page and tap jumping back to the app?

jimchou-dev:
> Quick question, how do we handle if user already open the UIKit View, and switch to deep link page and tap jumping back to the app?

Currently the app will crash, but for sample code purpose will ignore the case for now. Will create backlog item to manage error handling.

AzureCommunicationUI/AzureCommunicationUIDemoApp/SceneDelegate.swift
```diff
@@ -44,4 +45,34 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
...
+        let queryDict = URLContexts.first?.url.queryDictionary ?? [:]
+        lazy var deepLinkCoordinator = DeepLinkData()
+        let swiftUiDemoView = SwiftUIDemoView()
+            .environment(\.deepLinkCoordinator, deepLinkCoordinator)
+        deepLinkCoordinator.deepLinkValues = queryDict
+        guard let windowScene = scene as? UIWindowScene else { return }
+        let window = UIWindow(windowScene: windowScene)
+        window.rootViewController = UIHostingController(rootView: swiftUiDemoView)
+        self.window = window
+        window.makeKeyAndVisible()
```
AmyL219, line 61: Wonder if we have started the thread to discuss if it's okey to replace the whole window with a new swiftUI view when there is a deep link jump, even the user is already in a calling experience.

jimchou-dev: I spoke with @waynemo he said currently just launching a new view serves the purpose of having the deep link to populate the value to be available for UI testing. If it doesn't replace the whole window then the demo app would be more complex as it would need to store the state of the composite and navigation.
I have created 2695632 task to follow up on the issue when launching the app with deep-link while a call is on-going will crash. I probably will not be working on it the next little while as I'm occupied with some other task/projects, but the person who picks up the task can start a discussion thread.

PR #31, opened 2022-01-18
[Bugfix] Corrected the screen orientation update when the screen should update orientation

kevinyulu:
## Purpose
This PR fixes the issue in the calling screen, such that tapping on a participant drawer would trigger an orientation change due to incorrect orientation mapping from the orientation update logic.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code
* Launch the app, join a call (group or teams)
* In calling screen, switch device orientation to landscape (try landscape left and landscape right)
* Launch participant drawer or audio selection drawer 

## What to Check
* Make sure there is no further orientation switch when user launch participant drawer or audio selection drawer, during a call.

### DIFF
--- a/AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Container/ContainerUIHostingController.swift
+++ b/AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Container/ContainerUIHostingController.swift
@@ -55,12 +55,14 @@ class ContainerUIHostingController: UIHostingController<ContainerUIHostingContro
             .receive(on: RunLoop.main)
             .removeDuplicates()
             .sink(receiveValue: { orientation in
-                if orientation == .portrait || orientation == .landscape {
+                let portrait = orientation == .portrait
+                let landscape = orientation == .landscapeLeft || orientation == .landscapeRight
+                if  portrait || landscape {
                     // Apply a delay here to allow the previous orientation change to finish,
                     // then reset orientations
                     DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
                         let rotateOrientation: UIInterfaceOrientation = orientation == .portrait ?
-                            .portrait : (orientation == .landscapeLeft ? .landscapeLeft : .landscapeRight)
+                            .portrait : (orientation == .landscapeLeft ? .landscapeRight : .landscapeLeft)
                         UIDevice.current.rotateTo(oritation: rotateOrientation)
                         UIDevice.current.endGeneratingDeviceOrientationNotifications()
                     }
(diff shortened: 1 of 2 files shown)


AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Container/ContainerUIHostingController.swift
```diff
@@ -55,12 +55,14 @@ class ContainerUIHostingController: UIHostingController<ContainerUIHostingContro
             .receive(on: RunLoop.main)
             .removeDuplicates()
             .sink(receiveValue: { orientation in
-                if orientation == .portrait || orientation == .landscape {
+                let portrait = orientation == .portrait
+                let landscape = orientation == .landscapeLeft || orientation == .landscapeRight
```
AmyL219, line 59: Wouldn't orientation `.landscape` include `.landscapleLeft` and `.landscapeRight`?

AmyL219 (approved):

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Container/ContainerUIHostingController.swift
kevinyulu: The orientation `.landscape`  would represent the landscape orientation "in general". However in this scenario the app would need to know the current version of the landscape mode (left vs right) in order not to fall back to the incorrect one.

AmyL219: Since then, `if orientation == .portrait || orientation == .landscape { }` should also work, correct? Because `.landscape` includes `landscapeLeft` and `landscapeRight`

kevinyulu: @AmyL219 ,  I think if we use `orientation == .landscape` here, the rotation would not work for the views conform to `LockPhoneOrientation`, mainly because the supportedOrientation for `LockPhoneOrientation` for was set to either .landscapeLeft and landscapeRight.

The if clause for `orientation == .landscape` would return false for either `.landscapeRight` or `.landscapeLeft`. We'd need the exact left/right orientation in `LockPhoneOrientation` to make sure the screen orientation won't be accidentally changed.

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/ViewComponents/LockPhoneOrientation.swift
```diff
@@ -18,7 +18,7 @@ struct LockPhoneOrientation: ViewModifier {
         case .iphonePortraitScreenSize:
             return .portrait
         case .iphoneLandscapeScreenSize:
-            return .landscape
+            return UIDevice.current.orientation == .landscapeLeft ? .landscapeRight : .landscapeLeft
```
kevinyulu, line 21: Previously only `.landscape` was returned. App does not know which landscape orientation the phone was on, resulting in an accidental orientation change in one scenario. Hence I am adding to return the exact landscape oration info for this view.

jimchou-dev (approved):
Tested, LGTM

PR #36, opened 2022-01-27
[Bugfix] Updated the participant list UIs to match the latest UX requirement and Android for consistency

kevinyulu:
## Purpose
Updated the participant list UIs to match the latest UX requirement and Android for consistency
* Muted icon: use the unfilled style as indicated in the spec and on Android
* Updated display name local participant postfix from "(You)" to "(me)"
* Updated participant avatar style conditionally based on unnamed participant
* Updated text colour of unnamed participant name

Simulator Screen Shot - iPhone 12 Pro - 2022-01-27 at 00 19 19

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[x] Other... Please describe: UX update
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

* Join a call (using group or Teams) (with name filled, or without name filled)
* Launch participant drawer

## What to Check
In participant drawer: 
* Check if the muted icon style is the unfilled (regular) style as indicated in the spec and on Android 
* Check the display name local participant postfix is "(me)"
* Check if the participant avatar style (abbreviation) is conditionally based on the unnamed participant
* Check if the text colour of unnamed participant name is (#6E6E6E, light) and (#919191, dark)
* Check if the colour of the unfilled (regular) style **mute** icon is (#ACACAC, light) and (#919191, dark)


AzureCommunicationUI/AzureCommunicationUI/Presentation/FluentUI/Wrapper/CompositeParticipantsListCell.swift
```diff
@@ -10,12 +10,13 @@ class CompositeParticipantsListCell: TableViewCell {
     func setup(displayName: String, isMuted: Bool, accessoryType: TableViewCellAccessoryType = .none) {
         let isNameEmpty = displayName.trimmingCharacters(in: .whitespaces).isEmpty
         let avatar = MSFAvatar(style: .accent, size: .medium)
-        avatar.state.primaryText = displayName
+        let isDisplayNameUnnamed = displayName.contains(StringConstants.defaultEmptyName)
```
JoshuaLai, line 13: I am a little confused why we are doing this extra `let` here? Doesn't the `isNameEmpty` tell us the exact same thing?  
My only concern here is in the very odd edge case where a user actually types in `Unnamed participant` as a display name. 
The check below I think can just be 
```
avatar.state.primaryText = isNameEmpty? nil : displayName
```
And we can remove the extra `isDisplayNameUnnamed` in the or condition below.

kevinyulu: I think when a user leaves the displayName empty, the local user viewModel updates the name to "Unnamed Participant" and this "Unnamed Participant" is passed here in the `setup`. Thus the `isNameEmpty` will return `false`. Then this `isDisplayNameUnnamed` was used to check whether the name contains "Unnamed Participant". 

> My only concern here is in the very odd edge case where a user actually types in Unnamed participant as a display name.

yeah there is chance of this. I'll check

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/CallingViewComponent/ParticipantsListCellViewModel.swift
```diff
@@ -7,21 +7,31 @@ import Foundation
...
         self.isMuted = localUserState.audioState.operation != .on
     }
 
     init(participantInfoModel: ParticipantInfoModel) {
-        self.displayName = participantInfoModel.displayName
+        if participantInfoModel.displayName.trimmingCharacters(in: .whitespaces).isEmpty {
+            self.displayName = "\(StringConstants.defaultEmptyName)"
+            self.isParticipantNameEmpty = true
+        } else {
+            self.displayName = participantInfoModel.displayName
```
AmyL219, line 31: Can't let  `displayName =  participantInfoModel.displayName` in the viewModel. 
For specific logic how to handle displaying empty name, put it inside the participantsListCell?
Which could remove `isParticipantNameEmpty` property here

AmyL219: Feel like using `displayName.isEmpty` as the single source of truth is easier for understanding and maintainrance

AzureCommunicationUI/AzureCommunicationUI/Presentation/FluentUI/Wrapper/CompositeParticipantsListCell.swift
```diff
@@ -7,15 +7,26 @@ import Foundation
...
+
+    /// Set up the participant list item  in the participant list
+    /// - Parameters:
+    ///   - displayName: The display name shown on the participant list cell
+    ///   - isMuted: tells whether this participant is muted
+    ///   - partipantHasEmptyName: tells whether this participant had an empty name when joined the call
+    ///   - accessoryType: accessory type
+    func setup(displayName: String,
+               isMuted: Bool,
+               partipantHasEmptyName: Bool,
```
JoshuaLai, line 19: 2 things. 
Spelling here:
```suggestion
               participantHasEmptyName: Bool,
```
And if we are now passing in a boolean to represent this I don't think we need the 
`let isNameEmpty = displayName.trimmingCharacters(in: .whitespaces).isEmpty` so we should remove that.

JoshuaLai (approved):

AmyL219 (approved):

AmyL219:
Sorry, just remember of the ChangeLog update. 
Could you update the changelog to include this bug fix in your next PR?

PR #44, opened 2022-02-11
[Feature] Joining Experience

AmyL219:
## Purpose

* Update the UI/UX for users' joining experience
* Update Redux state to reflect the joining experience
* Code refactoring
* Remaining bugs when cancel the joining experience

Simulator Screen Shot - iPhone 11 - 2022-02-11 at 12 06 44

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```


AzureCommunicationUI/AzureCommunicationUI/Redux/Middleware/CallingMiddlewareHandler.swift
```diff
@@ -252,33 +253,24 @@ extension CallingMiddlewareHandler {
...
+                    return
                 }
+                let errorCode = callInfoModel.errorCode
+                let callStatus = callInfoModel.status
+                dispatch(CallingAction.StateUpdated(status: callStatus))
 
-                let action = CallingAction.StateUpdated(status: status)
-                dispatch(action)
+                self.logger.debug("Dispatch State Update: \(callStatus)")
+                self.handleInfo(callingStatus: callStatus, dispatch: dispatch)
```
kevinyulu, line 264: Should we be calling the self.handleInfo after checking the error code and call status?

AzureCommunicationUI/AzureCommunicationUI/Service/Calling/CallingSDKEventsHandler.swift
```diff
@@ -176,8 +176,11 @@ extension CallingSDKEventsHandler: CallDelegate,
 
     private func determineErrorType(previousStatus: CallingStatus, callEndReason: Int32) -> String {
         if callEndReason > 0 {
+            print("-------------the call end reason:\(callEndReason)")
             if callEndReason == 401 {
                 return CallCompositeErrorCode.tokenExpired
+            } else if callEndReason == 487 {
```
kevinyulu, line 182: Does 487 represent user leaves the meeting / is removed from the meeting by host/other participant?

kevinyulu (dismissed):
Tested the app in iPhone XR, the joining experience looks good. 
One minor side effect found: when trying to join the call without internet, the joining experience would transition to error state in setup view, which is good. However, if user dismisses the error message in the setup screen, the error won't appear for subsequent join meeting attempt.

AzureCommunicationUI/AzureCommunicationUI/Service/Calling/CallingSDKEventsHandler.swift
AmyL219: > Tested the app in iPhone XR, the joining experience looks good. One minor side effect found: when trying to join the call without internet, the joining experience would transition to error state in setup view, which is good. However, if user dismisses the error message in the setup screen, the error won't appear for subsequent join meeting attempt.

Good catch! Need to refactor the presentatin logic in errorBannerViewModel then

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Setup/SetupViewModel.swift
```diff
@@ -9,12 +9,15 @@ import Combine
 class SetupViewModel: ObservableObject {
     private let logger: Logger
     private let store: Store<AppState>
+    private var callingStatus: CallingStatus = .none
     var cancellables = Set<AnyCancellable>()
 
+    @Published var isJoinRequested: Bool = false
```
JoshuaLai, line 15: Is this extra complexity? Should we just use the state of the `CallingStatus`? When the status changes to `connecting` would be equivalent to `isJoinRequested` being true.

AzureCommunicationUI/AzureCommunicationUI/Service/Calling/CallingSDKEventsHandler.swift
AmyL219: 487 is the code indicate the user cancel the meeting which is an expected behaviro

kevinyulu (approved):
Tested the app and verified that the issue seen before about not able to show the dismissed error banner is no longer reproducible. Other app function (i.e, calls, mute/unmute, video screen share and camera share) works as expected. lgtm

AzureCommunicationUI/AzureCommunicationUI/Redux/Middleware/CallingMiddlewareHandlerExtension.swift
```diff
@@ -0,0 +1,47 @@
...
+            action = ErrorAction.FatalErrorUpdated(error: error)
+        } else {
+            action = ErrorAction.StatusErrorAndCallReset(error: error)
+        }
+
+        dispatch(action)
+        completion?()
+    }
+
+    func handle(callingStatus: CallingStatus, dispatch: @escaping ActionDispatch) {
```
JoshuaLai, line 28: This is a great idea. I like this addition

AzureCommunicationUI/AzureCommunicationUI/Redux/Action/CallingAction.swift
```diff
@@ -36,7 +36,7 @@ struct ErrorAction: Action {
         let error: ErrorEvent
     }
 
-    struct CallStateErrorUpdated: Action {
+    struct StatusErrorAndCallReset: Action {
```
JoshuaLai, line 39: nitpick: I like including the `CallStatus` somewhere in the name so we know exactly what status we are referring to.

JoshuaLai (approved):
Tested a lot of a error scenarios and everything is working as expected.

AzureCommunicationUI/AzureCommunicationUI/Redux/Action/CallingAction.swift
AmyL219: Main reason for the change is because we have `Call` before `Reset`. I think we can reuse this error action in other status update though

PR #48, opened 2022-02-16
[Refactor] Add unit tests for ParticipantGridViewModel

vhuseinova-msft:
## Purpose

* Added unit tests for state updates in ParticipantGridViewModel
* Disable public set for `participantsCellViewModelArr` variable in ParticipantGridViewModel

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```


AzureCommunicationUI/AzureCommunicationUITests/Presentation/Calling/ParticipantGridsViewModelTests.swift
```diff
@@ -126,6 +126,114 @@ class ParticipantGridsViewModelTests: XCTestCase {
...
+        sut.update(remoteParticipantsState: updatedState)
+        guard let updatedParticipant = sut.participantsCellViewModelArr.first else {
+            XCTFail("Failed with empty participantsCellViewModelArr")
+            return
+        }
+        XCTAssertIdentical(updatedParticipant, participant)
+        XCTAssertEqual(updatedParticipant.isMuted, isMuted)
+        XCTAssertEqual(updatedParticipant.isSpeaking, isSpeaking)
+        XCTAssertEqual(updatedParticipant.displayName, displayName)
+        XCTAssertEqual(updatedParticipant.videoStreamId, videoStreamId)
```
AmyL219, line 161: Could we only verify the `update` function inside `participantsCellViewModel` rather than verify the properties inside the participants. 
The function inside the `MockingParticipantsCellViewModel` would be 
```
func update(participantModel: ParticipantInfoModel) { 
      XCTAssertEqual(participantModel, expectedParticipantModel)
}
```

vhuseinova-msft: @AmyL219 creating a mock class for `ParticipantGridCellViewModel`and overriding `update(participantModel: ParticipantInfoModel)` func won't change anything and will break the logic for ` func update(remoteParticipantsState: RemoteParticipantsState)` as participantModel and expectedParticipantModel will never be the same (participantModel's properties won't be updated with new values). The main idea of the test, is that the `ParticipantGridCellViewModel` list will be updated and if there is an object with the same id in the updated list , the existed object in the list will be updated (not replaced).

However, I can create a separate expectedParticipantModel object and check its equality to updatedParticipant, not its properties. @AmyL219 Please let me know if there are any concerns with this solution.

vhuseinova-msft: Agreed with @AmyL219:

As was disscussed with the team previously, to use mock models for subviews' models in order to follow the same pattern for all subviews' models

AzureCommunicationUI/AzureCommunicationUITests/Presentation/Calling/ParticipantGridsViewModelTests.swift
```diff
@@ -288,10 +382,18 @@ class ParticipantGridsViewModelTests: XCTestCase {
 }
 
 extension ParticipantGridsViewModelTests {
```
AmyL219, line 384: Opsss, forgot to rename. Could you rename the class to `ParticipantGridViewModelTests`

AzureCommunicationUI/AzureCommunicationUITests/Mocking/ViewModels/ParticipantGridCellViewModelMocking.swift
```diff
@@ -0,0 +1,23 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import Foundation
+@testable import AzureCommunicationUI
+
+class ParticipantGridCellViewModelMocking: ParticipantGridCellViewModel {
+    private let updateParticipantModelCompletion: (() -> Void)?
```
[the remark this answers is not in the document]
vhuseinova-msft: Updated as requested

AzureCommunicationUI/AzureCommunicationUITests/Presentation/Calling/ParticipantGridsViewModelTests.swift
vhuseinova-msft: Updated!

AzureCommunicationUI/AzureCommunicationUITests/Presentation/Calling/ParticipantGridsViewModelTests.swift
```diff
@@ -98,16 +103,7 @@ class ParticipantGridsViewModelTests: XCTestCase {
         let state2 = RemoteParticipantsState(participantInfoList: [updatedStampInfoModel1, updatedStampInfoModel2],
```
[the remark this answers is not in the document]
vhuseinova-msft: Updated!

AzureCommunicationUI/AzureCommunicationUITests/Presentation/Calling/ParticipantGridsViewModelTests.swift
```diff
@@ -86,7 +86,12 @@ class ParticipantGridsViewModelTests: XCTestCase {
                                                          isSpeaking: false, recentSpeakingStamp: previousDate2)
         let state = RemoteParticipantsState(participantInfoList: [infoModel1, infoModel2],
                                             lastUpdateTimeStamp: previousDate2)
-        let sut = makeSUT()
+        let expectation = XCTestExpectation(description: "Participants list updated expectation")
+        expectation.expectedFulfillmentCount = 2
+        expectation.assertForOverFulfill = true
+        let sut = makeSUT { _ in
+            expectation.fulfill()
```
AmyL219, line 93: Is that possible to validate the update ParticipantInfoModel here? We probably only need one `updatedStampInfoModel1`
Something like:
```
let sut = makeSUT { infoModel in
            XCTAssertEqual(infoModel, updatedStampInfoModel1 )
            expectation.fulfill()
```

vhuseinova-msft: No, I don't think that it would be a good idea to add this check here. The same completion is used for both models infoModel1 and infoModel2. So the check will look like:
infoModel = infoModel1 || infoModel == infoModel2
That's why this check was added to test_participantGridsViewModel_updateParticipantsState_when_participantViewModelStateChanges_then_participantViewModelsUpdated test to check that the model is the same (there is the check of 1 model update)

AmyL219: Do we need two `updatedStampInfoModel`? since the function is just testing `update`? For unit test, should we verify the function output/input is correct?

vhuseinova-msft: @AmyL219, the two models are required here as per naming, the tests checks situation when existedTwoInfoModelsTimeStampUpdate.
 Also, the tests were updated as per[ this comment ](https://github.com/Azure/communication-ui-library-ios/pull/48#discussion_r808338287) to check not properties but the fact that inside models update function will be called.
 
Please let me know if there are still concerns, I'll update the code with propesed solution above.

AmyL219 (commented):
Looks good on the test cases naming. Look forward to have more discussion on unit tests

palatter (approved):
Looks good.

AmyL219 (approved):

PR #53, opened 2022-02-24
[Bug] Fix the local pip view not showing the video stream

AmyL219:
## Purpose

* The local pip view doesn't show the video renderView when user joining a meeting with more than 1 participant. After some debugging, find the root cause is the quick shift from `false` to `true` in `isParticipantGridDisplayed` after the view navigate to the calling view
* When the `isParticipantGridDisplayed` update from `false` to `true`, the calling view would show `localVideoFullscreenView` first and then immediately convert into `localVideoPipView`
* This would cause the `localVideoRenderView` quickly being pass to `localVideoPipView` from `localVideoFullscreenView`, and failed to be renderer for unknown reason(may related to the SwiftUI internal bridging). The `localVideoRenderView` is always correctly attached inside the `localVideoPipView` UIViewRepresentable, but just failed for renderering (don't know why)
* Prevent the shift from `isParticipantGridDisplayed` could fix this issue

MicrosoftTeams-image (24)

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

### DIFF
--- a/AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/CallingViewModel.swift
+++ b/AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/CallingViewModel.swift
@@ -9,7 +9,7 @@ import Combine
 class CallingViewModel: ObservableObject {
     @Published var isLobbyOverlayDisplayed: Bool = false
     @Published var isConfirmLeaveOverlayDisplayed: Bool = false
-    @Published var isParticipantGridDisplayed: Bool = false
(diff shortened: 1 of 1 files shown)


kevinyulu (approved):
Tested in iPhone XR, in group call, local pip view was showing. Issue was resolved.
tested Teams call as well, issue is not reproducible with or without the fix.

JoshuaLai (approved):
lgtm

PR #54, opened 2022-02-25
[Feature] Localization public api interface and configure locale

jimchou-dev:
## Purpose

* Public facing api for Localization
  * Initialize LocalizationConfiguration with `locale` and `localizationFilename` (custom localizable.strings file name), `isRightToLeft`
  * Initialize LocalizationConfiguration with `locale`, `customTranslations`, `isRightToLeft`
  * Get default supported languages
* Remove locale in compliance banner URL, will redirect to appropriate language based on user's browser language
* Update CallComposite initializer `withOptions` to be optional with default value nil

```
[ ] Yes
[ x ] No
```

## Pull Request Type

```
[ ] Bugfix
[ x ] Feature
[ ] Other... Please describe:
```

## What to Check
* Test when providing a unsupported language in LocalizationConfiguration will log warning message
* Can use static method to get supported languages
* Switch to a different language on device and test Teams call with recording/transcription, and click on `Learn more` or `Privacy Policy` will open website to system/browser language

## Other Information

* Will add UI for localization settings for Demo app in separate PR
* Documentation will be in separate PR

### DIFF
--- a/AzureCommunicationUI/AzureCommunicationUI/CallCompositeOptions/CallCompositeOptions.swift
+++ b/AzureCommunicationUI/AzureCommunicationUI/CallCompositeOptions/CallCompositeOptions.swift
@@ -9,11 +9,16 @@ import UIKit
 /// User-configurable options for creating CallComposite.
 public struct CallCompositeOptions {
     var themeConfiguration: ThemeConfiguration?
+    var localizationConfiguration: LocalizationConfiguration?
 
     /// Creates an instance of CallCompositeOptions with related options.
     /// - Parameter themeConfiguration: ThemeConfiguration for changing color pattern
-    public init(themeConfiguration: ThemeConfiguration) {
+    /// - Parameter localizationConfiguration: LocalizationConfiguration for changing
+    ///  locale, right-to-left mirroring layout, and customize locale strings
+    public init(themeConfiguration: ThemeConfiguration? = nil,
+                localizationConfiguration: LocalizationConfiguration? = nil) {
         self.themeConfiguration = themeConfiguration
+        self.localizationConfiguration = localizationConfiguration
     }
 
     public init() { }
--- a/AzureCommunicationUI/AzureCommunicationUI/CallCompositeOptions/CallCompositeOptions.swift
+++ b/AzureCommunicationUI/AzureCommunicationUI/CallCompositeOptions/CallCompositeOptions.swift
@@ -20,6 +20,4 @@ public struct CallCompositeOptions {
         self.themeConfiguration = themeConfiguration
         self.localizationConfiguration = localizationConfiguration
     }
-
-    public init() { }
 }
--- a/AzureCommunicationUI/AzureCommunicationUI/CallCompositeOptions/CallCompositeOptions.swift
+++ b/AzureCommunicationUI/AzureCommunicationUI/CallCompositeOptions/CallCompositeOptions.swift
@@ -8,8 +8,8 @@ import UIKit
 
 /// User-configurable options for creating CallComposite.
 public struct CallCompositeOptions {
-    var themeConfiguration: ThemeConfiguration?
-    var localizationConfiguration: LocalizationConfiguration?
+    private(set) var themeConfiguration: ThemeConfiguration?
+    private(set) var localizationConfiguration: LocalizationConfiguration?
 
     /// Creates an instance of CallCompositeOptions with related options.
     /// - Parameter themeConfiguration: ThemeConfiguration for changing color pattern
--- a/AzureCommunicationUI/AzureCommunicationUI/CallCompositeOptions/LocalizationConfiguration.swift
+++ b/AzureCommunicationUI/AzureCommunicationUI/CallCompositeOptions/LocalizationConfiguration.swift
@@ -8,37 +8,41 @@ import SwiftUI
 public struct LocalizationConfiguration {
     let customStrings: [String: String]
     let locale: String
-    let customLocalizableName: String
+    let localizableFilename: String
     let isRightToLeft: Bool
+
+    /// Create an instance of LocalizationConfiguration to customize localization.
+    /// - Parameter locale: Name of the language locale.
+    ///  Default value is "en" (English).
+    /// - Parameter localizableFilename:Name of the localizable strings file to override
+    ///  predefined Call Composite's localization string or to provide localization for an unsupported
+    ///  language.  The key of the string should be matched with the one in AzureCommunicationUI.
+    ///  Default value is "" (empty strings).
+    /// - Parameter isRightToLeft: Boolean for mirroring layout for right-to-left.
+    ///  Default value is `false`.
+    public init(locale: String = "en",
+                localizableFilename: String = "",
+                isRightToLeft: Bool = false) {
+        self.customStrings = [:]
+        self.locale = locale
+        self.localizableFilename = localizableFilename
+        self.isRightToLeft = isRightToLeft
+    }
 
     /// Create an instance of LocalizationConfiguration to customize localization.
     /// - Parameter customStrings: A dictionary of key-value pairs to override override
     ///  predefined Call Composite's localization string. The key of the string should be matched
     ///  with the one in AzureCommunicationUI.
+    /// - Parameter locale: Name of the language locale.
+    ///  Default value is "en" (English).
     /// - Parameter isRightToLeft: Boolean for mirroring layout for right-to-left.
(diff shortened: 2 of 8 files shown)


AzureCommunicationUI/AzureCommunicationUI/CallCompositeOptions/CallCompositeOptions.swift
```diff
@@ -9,11 +9,16 @@ import UIKit
 /// User-configurable options for creating CallComposite.
 public struct CallCompositeOptions {
     var themeConfiguration: ThemeConfiguration?
+    var localizationConfiguration: LocalizationConfiguration?
```
vhuseinova-msft, line 12: Could we make this property as private(set) (as it's initialized in init function) or a user should be able to update it later too?

vhuseinova-msft:
Tested, LGTM

vhuseinova-msft (approved):

AzureCommunicationUI/AzureCommunicationUI/CallCompositeOptions/CallCompositeOptions.swift
jimchou-dev: Make sense, will be consistent and also change themeConfiguration to private(set)

AzureCommunicationUI/AzureCommunicationUI/CallCompositeOptions/CallCompositeOptions.swift
```diff
@@ -8,12 +8,17 @@ import UIKit
...
-    var themeConfiguration: ThemeConfiguration?
+    private(set) var themeConfiguration: ThemeConfiguration?
+    private(set) var localizationConfiguration: LocalizationConfiguration?
 
     /// Creates an instance of CallCompositeOptions with related options.
     /// - Parameter themeConfiguration: ThemeConfiguration for changing color pattern
-    public init(themeConfiguration: ThemeConfiguration) {
+    /// - Parameter localizationConfiguration: LocalizationConfiguration for changing
+    ///  locale, right-to-left mirroring layout, and customize locale strings
+    public init(themeConfiguration: ThemeConfiguration? = nil,
```
waynemo, line 18: Should we have separate initializers that handle ThemeConfiguration and LocalizationConfiguration individually?

AzureCommunicationUI/AzureCommunicationUI/CallCompositeOptions/LocalizationConfiguration.swift
```diff
@@ -0,0 +1,60 @@
...
+
+    /// Create an instance of LocalizationConfiguration to customize localization.
+    /// - Parameter customStrings: A dictionary of key-value pairs to override override
+    ///  predefined Call Composite's localization string. The key of the string should be matched
+    ///  with the one in AzureCommunicationUI.
+    /// - Parameter locale: Name of the language locale.
+    /// - Parameter isRightToLeft: Boolean for mirroring layout for right-to-left.
+    ///  Default value is `false`.
+    public init(customStrings: [String: String],
+                locale: String,
```
waynemo, line 39: Why would we need `locale` passed in here? With customStrings, wouldn't it just use those regardless of the locale?

waynemo (changes requested):

AzureCommunicationUI/AzureCommunicationUI/CallCompositeOptions/LocalizationConfiguration.swift
jimchou-dev: Android team discussion suggests we should allow Contoso to provide a fallback locale, if they did not call `getLocaleStringDict()` and set it as the customStrings, if they just created a dictionary to override a few keys to in the customString, it should allow them to have a fallback language instead of always falling back to english (if unsupported locale is provided and key is not provided in customStrings dict, English strings will be used as default).

AzureCommunicationUI/AzureCommunicationUI/CallCompositeOptions/CallCompositeOptions.swift
jimchou-dev: Having default parameter allow Contoso to treat it as separate initializer if they only choose to do one or the other or both. Azure review board for iOS said this is okay, otherwise in the future if we have more configurable parameters, there will be too many initializers to manage.

AzureCommunicationUI/AzureCommunicationUI/CallCompositeOptions/LocalizationConfiguration.swift
```diff
@@ -0,0 +1,60 @@
...
+import SwiftUI
+
+public struct LocalizationConfiguration {
+    let locale: String
+    let localizableFilename: String
+    let customStrings: [String: String]
+    let isRightToLeft: Bool
+
+    /// Create an instance of LocalizationConfiguration to customize localization.
+    /// - Parameter locale: Name of the language locale.
```
JoshuaLai, line 15: I worry "Name" might be a little confusing. Developers might accidentally put in the "french" as a string. And since it is just a string there is a lot of room of interpretation and confusion
```suggestion
    /// - Parameter locale: String representing the locale code (ie. en, en_CA, fr, fr_FR)
```

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/CallingViewComponent/Banner/BannerInfoType.swift
```diff
@@ -75,11 +75,11 @@ enum BannerInfoType: Equatable {
              .transcriptionStoppedStillRecording,
              .transcriptionStarted,
              .recordingStoppedStillTranscribing:
-            return "https://privacy.microsoft.com/en-US/privacystatement#mainnoticetoendusersmodule"
+            return "https://privacy.microsoft.com/privacystatement#mainnoticetoendusersmodule"
```
JoshuaLai, line 78: Interesting. How do we know to go to the proper language if we don't pass in the locale code?

JoshuaLai (approved):
lgtm just a quick suggestion about your comment block and a small question about the privacy url

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/CallingViewComponent/Banner/BannerInfoType.swift
jimchou-dev: Its the solution suggested by localization team, without locale, it will redirect based on the browser/system language

AzureCommunicationUI/AzureCommunicationUI/CallCompositeOptions/LocalizationConfiguration.swift
jimchou-dev: The comment is not finalized, still need to sync with Android, will have a separate PR to correct them. Good suggestion locale code is probably better =)

AzureCommunicationUI/AzureCommunicationUI/CallCompositeOptions/CallCompositeOptions.swift
jimchou-dev: Travis said having initializer with optional value that is default `nil` is okay. Also suggested to change CallComposite initializer to have `withOptions` optional with default nil value

AzureCommunicationUI/AzureCommunicationUI/CallCompositeOptions/LocalizationConfiguration.swift
```diff
@@ -49,12 +49,4 @@ public struct LocalizationConfiguration {
     public static func getSupportedLanguages() -> [String] {
         return LocalizationProvider.supportedLocales
     }
-
-    /// Get the locale string dictionary for key-value pairs of a locale that is supported by Call Composite.
-    /// - Parameter locale: String representing locale code that is supported by Call Composite.
-    /// - Returns: A dictionary of key-value pair for localization keys with the string value of a given locale.
-    ///  Returns empty dictionary if the locale is not supported.
-    public static func getLocaleStringDict(locale: String) -> [String: String] {
```
JoshuaLai: Was there a reason for why we wanted this removed?

jimchou-dev: Android is having some issue converting strings.xml to hashMap, so to keep API consistent across platform, we won't expose this function as part of the API

PR #55, opened 2022-02-25
[Refactor] Add unit tests for SetupViewModel, PreviewAreaViewModel and SetupControlBarViewModel

vhuseinova-msft:
## Purpose

Added unit tests for:
- the receive function in SetupViewModel
- the update function in SetupControlBarViewModel
- the update function in PreviewAreaViewModel

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

### DIFF
--- a/AzureCommunicationUI/AzureCommunicationUITests/CompositeViewModelFactoryMocking.swift
+++ b/AzureCommunicationUI/AzureCommunicationUITests/CompositeViewModelFactoryMocking.swift
@@ -133,8 +133,7 @@ class CompositeViewModelFactoryMocking: CompositeViewModelFactory {
     }
 
     func makeBannerTextViewModel() -> BannerTextViewModel {
-        return bannerTextViewModel ??
-            BannerTextViewModel()
+        return bannerTextViewModel ?? BannerTextViewModel()
     }
 
     // MARK: SetupViewModels
--- a/AzureCommunicationUI/AzureCommunicationUITests/CompositeViewModelFactoryMocking.swift
+++ b/AzureCommunicationUI/AzureCommunicationUITests/CompositeViewModelFactoryMocking.swift
@@ -12,7 +12,23 @@ class CompositeViewModelFactoryMocking: CompositeViewModelFactory {
     var store: Store<AppState>
 
     var bannerTextViewModel: BannerTextViewModel?
+    var controlBarViewModel: ControlBarViewModel?
+    var infoHeaderViewModel: InfoHeaderViewModel?
+    var localVideoViewModel: LocalVideoViewModel?
+    var participantGridViewModel: ParticipantGridViewModel?
+    var participantsListViewModel: ParticipantsListViewModel?
+    var bannerViewModel: BannerViewModel?
+    var previewAreaViewModel: PreviewAreaViewModel?
+    var setupControlBarViewModel: SetupControlBarViewModel?
+    var errorInfoViewModel: ErrorInfoViewModel?
+    var audioDeviceListViewModel: AudioDeviceListViewModel?
+    var primaryButtonViewModel: PrimaryButtonViewModel?
+    var iconButtonViewModel: IconButtonViewModel?
+    var setupViewModel: SetupViewModel?
+    var callingViewModel: CallingViewModel?
+
     var createMockParticipantGridCellViewModel: ((ParticipantInfoModel) -> ParticipantGridCellViewModel?)?
+    var createIconWithLabelButtonViewModel: ((CompositeIcon) -> IconWithLabelButtonViewModel?)?
 
     init(logger: Logger,
          store: Store<AppState>) {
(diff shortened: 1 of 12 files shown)


AzureCommunicationUI/AzureCommunicationUITests/Presentation/Setup/SetupViewModelTests.swift
```diff
@@ -8,24 +8,23 @@ import XCTest
...
 
 class SetupViewModelTests: XCTestCase {
-
-    var storeFactory: StoreFactoryMocking!
-    var cancellable: CancelBag!
-    var setupViewModel: SetupViewModel!
+    fileprivate var storeFactory: StoreFactoryMocking!
+    fileprivate var factoryMocking: CompositeViewModelFactoryMocking!
+    fileprivate var cancellable: CancelBag!
+    fileprivate var logger: LoggerMocking!
```
JoshuaLai, line 14: Why are we making these `fileprivate`?

AzureCommunicationUI/AzureCommunicationUITests/Presentation/Setup/PreviewAreaViewModelTests.swift
```diff
@@ -8,19 +8,16 @@ import XCTest
 @testable import AzureCommunicationUI
 
 class PreviewAreaViewModelTests: XCTestCase {
-    var storeFactory: StoreFactoryMocking!
-    var previewAreaViewModel: PreviewAreaViewModel!
+    fileprivate var storeFactory: StoreFactoryMocking!
+    fileprivate var factoryMocking: CompositeViewModelFactoryMocking!
```
JoshuaLai, line 12: Why are we making this `fileprivate`?

AzureCommunicationUI/AzureCommunicationUITests/Presentation/Setup/SetupViewModelTests.swift
vhuseinova-msft: Good catch, all of them should be private

AzureCommunicationUI/AzureCommunicationUITests/Presentation/Setup/PreviewAreaViewModelTests.swift
vhuseinova-msft: same as for the previous comment

AzureCommunicationUI/AzureCommunicationUITests/CompositeViewModelFactoryMocking.swift
```diff
@@ -27,33 +37,37 @@ class CompositeViewModelFactoryMocking: CompositeViewModelFactory {
     }
 
     func getSetupViewModel() -> SetupViewModel {
-        return SetupViewModel(compositeViewModelFactory: self, logger: logger, store: store)
+        return setupViewModel ?? SetupViewModel(compositeViewModelFactory: self,
```
JoshuaLai, line 40: I am not convinced about the value of saving / returning a reference gives us in tests. Don't we want to always have a fresh view model in each test?

vhuseinova-msft: These models are set for every object of CompositeViewModelFactoryMocking in the test where they are needed. So, for example in `test_setupControlBarViewModel_updateStates_when_stateUpdated_then_audioDeviceListViewModelUpdated` only audioDeviceListViewModel is set as a mocking model and all others are created from scratch.
This is needed for possibility to set a mocking model instead of a real one.
For testing situations when a couple of models can be created for the same test, the completions are used like `createMockParticipantGridCellViewModel` and `createIconWithLabelButtonViewModel`.

Also, please note, [that as setUp function](https://developer.apple.com/documentation/xctest/xctestcase/set_up_and_tear_down_state_in_your_tests) is used for initializing properties and/or makeSut for initializing sut, we always have the clean version of CompositeViewModelFactoryMocking and all its properties

AzureCommunicationUI/AzureCommunicationUITests/Presentation/Setup/SetupViewModelTests.swift
vhuseinova-msft: @JoshuaLai updated!

AzureCommunicationUI/AzureCommunicationUITests/Presentation/Setup/PreviewAreaViewModelTests.swift
vhuseinova-msft: Updated!

AmyL219 (approved):
Those test cases look pretty clear to me

JoshuaLai (approved):

PR #61, opened 2022-03-08
[Feature] Localization RTL support

jimchou-dev:
## Purpose

* Implemented right-to-left support for both SwiftUI and UIKit component

Sample screenshots uses customTranslation dictionary for Arabic only on `Join Call` button
| Normal (isRightToLeft=false) | Mirrored (isRightToLeft=true) |
| ---------------------------- | -----------------------------| 
| image | image
| image | image |
| image |  image |

```
[ ] Yes
[ x ] No
```

## Pull Request Type

```
[ ] Bugfix
[ x ] Feature
[ ] Other... Please describe:
```

## What to Check
* Adjust Is Right-to-Left in settings to see that layout will be mirrored
  * Text will not be mirrored
  * Only image that will be mirrored is the back arrow button in setupView


JoshuaLai:
Just a quick question about landscape mode. Not sure if the design is expected with the `rtl` being set to true now our end call button is at the top

Screen Shot 2022-03-09 at 2 51 30 PM

jimchou-dev:
> Just a quick question about landscape mode. Not sure if the design is expected with the `rtl` being set to true now our end call button is at the top
> 
> Screen Shot 2022-03-09 at 2 51 30 PM

I've just confirmed with Alex, its intended behaviour, with RTL=false it behaves like the following, so is just mirror left to right and not vertically mirrored.
image

vhuseinova-msft (approved):
Tested, LGTM

AzureCommunicationUI/AzureCommunicationUI/Presentation/Style/LocalizationProvider.swift
```diff
@@ -18,7 +19,7 @@ class AppLocalizationProvider: LocalizationProvider {
     private var locale: String = "en"
     private var localizableFilename: String = ""
     private var customTranslations: [String: String] = [:]
-    private var isRightToLeft: Bool = false
+    var isRightToLeft: Bool = false
```
vhuseinova-msft, line 22: Just a minor suggestion: it may be useful to set private(set) for isRightToLeft

AzureCommunicationUI/AzureCommunicationUI/Presentation/Style/LocalizationProvider.swift
```diff
@@ -38,6 +39,7 @@ class AppLocalizationProvider: LocalizationProvider {
         locale = localeConfig.locale
         localizableFilename = localeConfig.localizableFilename
         customTranslations = localeConfig.customTranslations
+        isRightToLeft = localeConfig.isRightToLeft
```
JoshuaLai, line 42: This is a little contradicting. Our protocol defines `var isRightToLeft: Bool { get }` so it is a getter. But here we are saving it as a `var` so we can set it again. I think this just needs a bit of clarity. Should the protocol just make the `isRightToLeft` a `{ get set }`?

AzureCommunicationUI/AzureCommunicationUI/CallComposite.swift
```diff
@@ -42,15 +42,17 @@ public class CallComposite {
         logger?.debug("launch composite experience")
 
         dependencyContainer.registerDependencies(callConfiguration)
-
+        let localizationProvider = dependencyContainer.resolve() as LocalizationProvider
```
AmyL219, line 45: I see there are a lot of `static` functions inside LocalizationProvider. So we want the provider to be singleton similar to `StyleProvider`?
But I saw the injection of localizationProvider both inside `SetupViewModel`, `DependencyContainer`. Could we make the structure more consistent?
Can we either 1. make the provider to be a object and get to it by injection to viewModels
2. make the provider with `static` method, and get access to the value from viewLayer, like the StyleProvider?

AzureCommunicationUI/AzureCommunicationUI/Presentation/Style/LocalizationProvider.swift
jimchou-dev: As @vhuseinova-msft suggested, `isRightToLeft` is only set when initialized and changed when `.apply(localeConfig)` so it make sense to be `private(set) var` and the protocol will only be for getter

AzureCommunicationUI/AzureCommunicationUI/CallComposite.swift
jimchou-dev: There has been some refactoring changing LocalizationProvider from previous design (static method) to singleton with dependency injection (https://github.com/Azure/communication-ui-library-ios/pull/62). So all the `static` are removed from LocalizationProvider and it is now a protocol. The dependency injection allow us to better unit testing.

AmyL219: Okey, I see. 
But static method also can be unit tested. The UI string verification work is usually validated by UI test. Even we can unit test view model, but how can we guarantee the viewModel's string has been correctly mapped to the view's component though?

waynemo: @AmyL219 @jimchou-dev, by using DI with a protocol, we can decouple the actual implementation of the `LocalizationProvider`. 

In the unit tests for the viewmodels, we can inject a mock for `LocalizationProvider` and verify that the correct strings are being fetched based on the mock. Here, we do not actually need to test the logic of how the strings are being fetched (from Localization.strings, overrides, etc.).

We can have separate tests for the implementation of `LocalizationProvider` to verify all the internal logic.

JoshuaLai (approved):

AzureCommunicationUI/AzureCommunicationUI/CallComposite.swift
AmyL219: Okey, I see the point. 
My concern is this would increase the workload when we are going to use the string. We need to add another string properties definition on the viewModel to mapping the string value from LocalizationProvider to string value, correct? 
Also, increase the workload in writing unit tests which we still need to write UI tests to finally verify the string.

jimchou-dev: Yes that is correct, but we should be having string property in viewModel regardless. Because some strings currently are directly hard coded and does not follow MVVM, and this would be an opportunity to make them conform to MVVM.
So when introducing a new string, instead of hardcoding the value, to use LocalizationProvider will need the following steps:
1. Add key-val to english Localizable.strings file
2. Add the key to the StringKey enum
3. Use localizationProvider.getLocalizedString(StringKey.enum)

waynemo: "we still need to write UI tests to finally verify the string"

This would be less critical to have if we are always using the properties on the viewmodels for strings.

AmyL219: Sounds good to me. You are going to do the string immigration to VMs and unit tests in coming PR, correct?

jimchou-dev: Correct, the string migration to VM will be done in separate task for SetupView, LobbyView, and CallingView, which would include associated unit tests.

PR #64, opened 2022-03-10
[Feature] Introduce the LobbyOverlayViewModel with localization

JoshuaLai:
## Purpose
Introduce new view model to include localization provider. 
Using this to have dynamic localized string in lobby overlay

```
[ ] Yes
[x] No
```

## Pull Request Type
```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
Using the new settings page to change the language to something other than `en`
Join teams meeting 
Check new content in the lobby overlay.


AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/CallingView.swift
```diff
@@ -64,7 +64,8 @@ struct CallingView: View {
                 viewModel.infoHeaderViewModel.toggleDisplayInfoHeader()
             })
             .modifier(PopupModalView(isPresented: viewModel.isLobbyOverlayDisplayed) {
-                LobbyOverlayView()
+                let lobbyViewModel = viewModel.getLobbyOverlayViewModel()
+                LobbyOverlayView(viewModel: lobbyViewModel)
```
jimchou-dev, line 68: nitpick: maybe do it in one line without having a variable to be more consistent with the
`PrimaryButton(viewModel: viewModel.get___ViewModel())`

jimchou-dev (approved):
Tested, LGTM

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/CallingView.swift
JoshuaLai: Applied

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/CallingViewModel.swift
```diff
@@ -58,6 +58,10 @@ class CallingViewModel: ObservableObject {
             }.store(in: &cancellables)
     }
 
+    func getLobbyOverlayViewModel() -> LobbyOverlayViewModel {
+        return LobbyOverlayViewModel(localizationProvider: localizationProvider)
```
AmyL219, line 62: Should we use the factory to generate the new subview models?

AzureCommunicationUI/AzureCommunicationUITests/Presentation/Calling/LobbyOverlayViewModelTests.swift
```diff
@@ -0,0 +1,33 @@
...
+    private var localizationProvider: LocalizationProvider!
+
+    override func setUp() {
+        super.setUp()
+        localizationProvider = LocalizationProviderMocking()
+    }
+
+    func test_lobbyOverlayViewModel_displaysTitleAccordingToLocalizationProvided() {
+        let sut = makeSUT()
+        XCTAssertEqual(sut.title, "ABC")
```
AmyL219, line 20: "ABC"?

JoshuaLai: Yes in our mocking we hardcode it to "ABC" so this ensures that if the client passes in something that will override the localization dictionary we are honouring it

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/CallingViewModel.swift
JoshuaLai: I was debating this as well. IMO because we only have lobby in the call view I think its fine to have in created in the `CallingViewModel` if in the future we find we need to have the lobby in other places in the flow than maybe we can move it to the factory.

waynemo: Even if it's only used here, I think it will be better to follow the pattern of using the factory. Otherwise, we might cause confusion in the future when new view models are needed.

AzureCommunicationUI/AzureCommunicationUITests/Presentation/Calling/LobbyOverlayViewModelTests.swift
waynemo: It might cause some confusion at first glance why we're validating against "ABC". Could we use something the resembles what the actual string might be? 
e.g. "Overlay Title"

waynemo (changes requested):

AzureCommunicationUI/AzureCommunicationUITests/LocalizationProviderMocking.swift
```diff
@@ -6,26 +6,32 @@
...
     func getSupportedLanguages() -> [String] {
+        isGetSupportedLanguages = true
         return ["en"]
     }
 
     func getLocalizedString(_ key: StringKey) -> String {
-        // return hard coded string
-        return "ABC"
+        isGetLocalizedString = true
+        return key.rawValue
```
AmyL219, line 30: This makes much sense than returning `ABC`. If all the key return same value `ABC`, how can we test if we have fetched the value with correct StringKey input?

AmyL219 (approved):

PR #68, opened 2022-03-11
[FEATURE] Accessibility for Participant Header on the Calling Screen

vhuseinova-msft:
## Purpose

* Keep Participant Header on Screen when using VoiceOver

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

## What to Check
* the Participant Header is shown when VoiceOver is on or turned on on the Calling Screen
* the default Participant Header behaviour when VoiceOver is off or turned off on the Calling Screen


palatter (commented):
Tested and it works for the situation where you have accessibility already on and you join call.
 
I noticed you've added a listener to adjust behavior for when you turn on/of voiceOver in the call. When testing this though, it wasn't consistently bringing up and keeping infoHeader on screen.

If this is difficult to resolve, we can leave this functionality out, as this isn't even support in Teams and it's unlikely someone would turn on VoiceOver once already in a call.

Otherwise it looks good!

vhuseinova-msft:
> Tested and it works for the situation where you have accessibility already on and you join call.
> 
> I noticed you've added a listener to adjust behavior for when you turn on/of voiceOver in the call. When testing this though, it wasn't consistently bringing up and keeping infoHeader on screen.
> 
> If this is difficult to resolve, we can leave this functionality out, as this isn't even support in Teams and it's unlikely someone would turn on VoiceOver once already in a call.
> 
> Otherwise it looks good!

Great catch! Updated!

palatter:
> > Tested and it works for the situation where you have accessibility already on and you join call.
> > I noticed you've added a listener to adjust behavior for when you turn on/of voiceOver in the call. When testing this though, it wasn't consistently bringing up and keeping infoHeader on screen.
> > If this is difficult to resolve, we can leave this functionality out, as this isn't even support in Teams and it's unlikely someone would turn on VoiceOver once already in a call.
> > Otherwise it looks good!
> 
> Great catch! Updated!

Works now!

palatter (approved):

AzureCommunicationUI/AzureCommunicationUI/Presentation/Manager/AccessibilityProvider.swift
```diff
@@ -6,19 +6,44 @@
...
+
+    /// Move focus to the element with the highest sort priority for the current view
+    func moveFocusToFirstElement()
+
+    /// Add observer for voiceOverStatusDidChangeNotification
+    /// - Parameter completion: completion that will be executed when receiving a notification
+    func subscribeToVoiceOverStatusDidChangeNotification(_ observer: AccessibilityProviderNotificationsObserver)
+}
+
+struct AppAccessibilityProvider: AccessibilityProvider {
```
jimchou-dev, line 25: I think there are two conventions for protocol, we probably should have separate PR to make them consistent. Currently there are AppXYZ for AppAudioSessionManager and AppPermissionManager

AzureCommunicationUI/AzureCommunicationUITests/CompositeViewModelFactoryMocking.swift
```diff
@@ -31,9 +32,11 @@ class CompositeViewModelFactoryMocking: CompositeViewModelFactory {
     var createIconWithLabelButtonViewModel: ((CompositeIcon) -> IconWithLabelButtonViewModel?)?
 
     init(logger: Logger,
-         store: Store<AppState>) {
+         store: Store<AppState>,
+         accessibilityProvider: AccessibilityProvider = AccessibilityProviderMocking()) {
```
jimchou-dev, line 36: minor: For mocking, probably don't need the default value, can pass in mocking object when the class object is being created

AzureCommunicationUI/AzureCommunicationUITests/AccessibilityProviderMocking.swift
```diff
@@ -0,0 +1,27 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import Foundation
+@testable import AzureCommunicationUI
+
+class AccessibilityProviderMocking: AccessibilityProvider {
```
jimchou-dev, line 9: probably want to add some variable like `var isPostQueuedAnnouncementCalled: Bool = false` to test if the function is called to test XCAssertTrue

jimchou-dev (commented):
Just  FYI some of the makeSUT might have conflict with the localization test. For localization regular `makeSUT` will use the AppLocalizationProvider, and `makeSUTLocalization` using the mocking object

AzureCommunicationUI/AzureCommunicationUITests/CompositeViewModelFactoryMocking.swift
vhuseinova-msft: I thought about this, but mostly we don't need to do anything for AccessibilityProvider and as CompositeViewModelFactoryMocking is created for every ViewModel (or almost every), I added the default value to decrease the duplications.

AzureCommunicationUI/AzureCommunicationUITests/AccessibilityProviderMocking.swift
vhuseinova-msft: @jimchou-dev  it's already tested by using completion block (`postQueuedAnnouncementBlock`) with a check for an announcement. Please let me know if you meant something else.

AzureCommunicationUI/AzureCommunicationUI/Presentation/Manager/AccessibilityProvider.swift
vhuseinova-msft: I've created a refactor task for this!

palatter:
Two more things I just noticed. 

1. When you navigate to call screen, we don't focus on waiting for participants element.
2. If you press the people button and then dismiss it, the infoHeader also dismisses, with no way to bring it back.

jimchou-dev (approved):
Tested, LGTM

vhuseinova-msft:
> Two more things I just noticed.
> 
> 1. When you navigate to call screen, we don't focus on waiting for participants element.
> 2. If you press the people button and then dismiss it, the infoHeader also dismisses, with no way to bring it back.

@palatter  the issues are fixed, the app should work propely now.
Also, updated behavour for this view for case when a user is in lobby

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Utilities/IconAndLabelConversion.swift
```diff
@@ -20,9 +20,9 @@ extension LocalUserState.AudioDeviceSelectionStatus {
     var label: String {
        switch self {
        case .receiverSelected:
-          return AudioDeviceType.receiver.name
+           return AudioDeviceType.receiver.name.rawValue
```
vhuseinova-msft, line 23: This will be updated in accessibility localization PR as agreed with @palatter

PR #70, opened 2022-03-14
[Bugfix] Force portrait mode in setup view

cheukchuen:
## Purpose

* Only stop receiving orientation notification in setup view
  * As a temporary workaround for local video stream to not rotate the video when rotating screen
* Calling view still receives orientation notification while orientation locked when displaying drawer
  * So that screen rotates to proper orientation when dismissing drawer for participants list/audio device selection

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## What to Check
* Setup view remains in portrait mode when screen was in landscape mode before entering setup screen
* Setup view remains in portrait mode when screen was in landscape mode during the call and after call has ended
* Calling view able to rotate to proper orientation after dismissing the drawer
  * Able to rotate to portrait mode after dismissing the drawer in landscape mode
  * Able to rotate to landscape mode after dismissing the drawer in portrait mode


palatter (dismissed):
Looks good.

Tested rotation with background/foreground, video and screen share

JoshuaLai (dismissed):
Tested and its working as expected. But I was never able to reproduce this issue so I would be curious @AmyL219 if this fixes it for you.

JoshuaLai (dismissed):
Looks good still the same changes

jimchou-dev (dismissed):
LGTM

AmyL219 (dismissed):
Test. Couldn't reproduce the bug in this morning! Looks good

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Container/ContainerUIHostingController.swift
```diff
@@ -49,37 +51,36 @@ class ContainerUIHostingController: UIHostingController<ContainerUIHostingContro
...
         environmentProperties
             .$isProximitySensorOn
             .receive(on: RunLoop.main)
+            .removeDuplicates()
             .sink(receiveValue: { isEnable in
                 UIDevice.current.toggleProximityMonitoringStatus(isEnabled: isEnable)
             }).store(in: cancelBag)
     }
 
+    private func haltSetupViewOrientation(containerView: ContainerView) {
```
AmyL219, line 77: Why do we need this in the initilizer? Would this be duplicated to line 60 -62 though?

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Container/ContainerUIHostingController.swift
```diff
@@ -49,37 +51,36 @@ class ContainerUIHostingController: UIHostingController<ContainerUIHostingContro
...
-                        UIDevice.current.rotateTo(oritation: rotateOrientation)
-                        UIDevice.current.endGeneratingDeviceOrientationNotifications()
-                    }
-                } else if (orientation == .all || orientation == .allButUpsideDown)
-                            && !UIDevice.current.isGeneratingDeviceOrientationNotifications {
+            .sink(receiveValue: { _ in
+                if containerView.router.currentView == .setupView,
+                   UIDevice.current.isGeneratingDeviceOrientationNotifications {
+                    UIDevice.current.endGeneratingDeviceOrientationNotifications()
+                } else if !UIDevice.current.isGeneratingDeviceOrientationNotifications {
```
AmyL219, line 63: I feel like the workaround Kevin implemented before doesn't work anymore. The local videostream keeps rotating inside the setup view. Could @kevinyulu you have test on PR? 
Either we remove the workaround completely or fix the workaround.

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Container/ContainerUIHostingController.swift
cheukchuen: Need this in `init` for when the composite first loads. `supportedOrientations` is set to `portrait` initially, so couldn't reach lines 60-62 due to `.removeDuplicates()`.

Of course we could also remove `.removeDuplicates()` and `haltSetupViewOrientation(ContainerView:)`, but this hides the issue and in the future someone else might unknowingly add `.removeDuplicates()` back

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Container/ContainerUIHostingController.swift
kevinyulu: @AmyL219  yes I tested and confirmed that the local video camera scene orientation "flipped" when device is rotated.

cheukchuen: Latest commit fixes the issue, please try again thanks!

kevinyulu: @cheukchuen  how about we add one check in the else if above? 

```
} else if containerView.router.currentView != .setupView,
!UIDevice.current.isGeneratingDeviceOrientationNotifications {
UIDevice.current.beginGeneratingDeviceOrientationNotifications()
 ```
 
 I tested this and it looked like it does the following:
 * stop camera orientation accidentally flipped when device is rotated in setup view, and
 * allow orientation update after participant drawer is dismissed. 
 Let me know if there is other side effect this approach introduces.

cheukchuen: That's correct @kevinyulu, in the setup view we want to force portrait mode and prevent camera orientation flipping, whereas for calling view we want the screen to be rotated after drawer dismissal at the expense of camera flipping.

Good idea on the suggestion as well. The initial flow condition we had for this PR resulted in the `else` condition being evaluated for setup view, hence the issue reported by @AmyL219. Made a change to use `switch` to be specific about what should occur for each view.

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Container/ContainerUIHostingController.swift
AmyL219: Okey, let keep them both there

kevinyulu (approved):

AmyL219 (approved):

PR #74, opened 2022-03-14
[Documentation] Added documentation for Localization public facing API 

jimchou-dev:
## Purpose

* Added documentation strings for public facing api
* Added readme file for localization public facing api
* Variable rename from `locale` to `languageCode`

```
[ ] Yes
[ x ] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ x ] Documentation content changes
[ ] Other... Please describe:
```

## What to Check
* Ensure app is still running and tests are passing


AzureCommunicationUI/AzureCommunicationUI/CallCompositeOptions/LocalizationConfiguration.swift
```diff
@@ -6,46 +6,49 @@ import Foundation
 import SwiftUI
 
 public struct LocalizationConfiguration {
-    let locale: String
+    let language: String
```
JoshuaLai, line 9: We changed this because it was going to align more with Android naming right? 
I am curious to hear Travis or review board feedback on this name change.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 - Update joining experience in setup view. [#44](https://github.com/Azure/communication-ui-library-ios/pull/44)
 - Support pinch & zoom, pan and double tap to zoom in screen share presentation. [#29](https://github.com/Azure/communication-ui-library-ios/pull/29)
 - Update audio device selection list when connected to bluetooth/headphones. [#56](https://github.com/Azure/communication-ui-library-ios/pull/56)
+- Implemented Localization support for 13 languages and right-to-left mirror layout. [#73](https://github.com/Azure/communication-ui-library-ios/pull/73)
```
JoshuaLai, line 9: I think we want to include in the change log the change to the CallCompositeOptions init function. Previously we had themeConfig as required but now it has been updated to being optional.

jimchou-dev: Before there was 2 initializer for CallCompositeOptions
-init(themeConfiguration: ThemeConfiguration)
-init()
So having the only one (new) initializer but optional with default parameter, i think it serves the same purpose. Do we still need to outline it in the changelog since to Contoso the behaviour won't change

JoshuaLai: Oh right we had an old `init()` I guess in that case we don't need to highlight it

JoshuaLai (approved):

kevinyulu (approved):

PR #78, opened 2022-03-16
[Feature] Adding new settings page to the demo app instead of the entry view

JoshuaLai:
## Purpose
Adding setting to demo app to be more convenient 

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
Launch the demo app and check the new settings 

## What to Check
* ...

### DIFF
--- a/AzureCommunicationUI/AzureCommunicationUIDemoApp/Views/UIKitDemoViewController.swift
+++ b/AzureCommunicationUI/AzureCommunicationUIDemoApp/Views/UIKitDemoViewController.swift
@@ -26,6 +26,7 @@ class UIKitDemoViewController: UIViewController {
     private var displayNameTextField: UITextField!
     private var groupCallTextField: UITextField!
     private var teamsMeetingTextField: UITextField!
+    private var settingsButton: UIButton!
     private var startExperienceButton: UIButton!
     private var acsTokenTypeSegmentedControl: UISegmentedControl!
     private var meetingTypeSegmentedControl: UISegmentedControl!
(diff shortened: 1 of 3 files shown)


jimchou-dev (approved):
Should we have the Settings button at top right to match Android?
image

Or maybe to be simple have the Settings button below `Start Experience` is a bit more logical to me.
Tested, LGTM

AmyL219 (approved):

AzureCommunicationUI/AzureCommunicationUIDemoApp/Views/UIKitDemoViewController.swift
```diff
@@ -262,6 +263,15 @@ class UIKitDemoViewController: UIViewController {
         return true
     }
 
+    @objc func onSettingsPressed() {
+        let settingsView = SettingsView(envConfigSubject: envConfigSubject, dismissAction: {
+            self.dismiss(animated: true, completion: nil)
+        })
+        let settingsViewHostingController = UIHostingController(rootView: settingsView)
+        settingsViewHostingController.modalPresentationStyle = .fullScreen
```
kevinyulu, line 271: nice to have: on iPad this is covering the whole screen, in the future we may want to use other presentation style.

vhuseinova-msft (approved):
I like  the Jim's idea about mooving the Settings button. Otherwise, LGTM, tested.

kevinyulu (approved):
tested on iPhone XR iOS 15.3, 
Settings button displays above Start Experience button, language settings works as expected. 
Tested the UX in portrait and landscape. UX looks good. 
Also tested light and dark mode. looks good.

JoshuaLai:
> I like the Jim's idea about mooving the Settings button. Otherwise, LGTM, tested.

Agreed. Jim and I chatted about it. To do that I think best practice would be for our entry view to implement a navigation controller. And use an accessory in that. It would be a larger refactor so we will track it as a upcoming refactor task

PR #83, opened 2022-03-17
[Feature] Resolve issue where VoiceOver would sometimes read a word for an icon. Also added label for new join call button.

palatter:
## Purpose
Resolve issue where VoiceOver would sometimes read a word for an icon. This was a super strange bug, I'm not entirely sure how it was even happening, but VoiceOver was somehow interpreting a word for some of the images used for the icons. In the case of the camera on button, it would read the word 'video' after the word 'button'. Using children ignore did not work either, so i had to hide the entire element.

Also added label for new join call button and an accessibility hint option for future use.
* ...

```
[ ] Yes
[ X] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...

### DIFF
--- a/AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Setup/SetupViewModel.swift
+++ b/AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Setup/SetupViewModel.swift
@@ -14,7 +14,7 @@ class SetupViewModel: ObservableObject {
 
     let previewAreaViewModel: PreviewAreaViewModel
     let title: String = "Setup"
-    
+
     var errorInfoViewModel: ErrorInfoViewModel
     var dismissButtonViewModel: IconButtonViewModel!
     var joinCallButtonViewModel: PrimaryButtonViewModel!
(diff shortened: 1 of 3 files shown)


AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Setup/SetupViewModel.swift
```diff
@@ -53,6 +53,7 @@ class SetupViewModel: ObservableObject {
                 }
                 self.joinCallButtonTapped()
         }
+        self.joinCallButtonViewModel.update(accessibilityLabel: "Join Call")
```
JoshuaLai, line 56: This is most likely something we will look into in the future. But I think we should be setting the localized string into the accessibility label.

palatter: I am planning on doing so for all accessibility labels as soon as we merge the feature into develop/beta, which now has localization.

JoshuaLai:
```suggestion
        self.joinCallButtonViewModel.update(accessibilityLabel:  localizationProvider.getLocalizedString(.joinCall))
```
Something like this

JoshuaLai (approved):

jimchou-dev (approved):
I notice that before, when testing, thought was intentional.

Tested, there was some trailing whitespace in SetupViewModel so pipeline is failing, other than that LGTM!

PR #86, opened 2022-03-18
[FEATURE] Localize Accessibility Strings

palatter:
## Purpose

* ...

```
[ ] Yes
[ ] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...


AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/CallingViewComponent/ControlBarViewModel.swift
```diff
@@ -78,7 +91,8 @@ class ControlBarViewModel: ObservableObject {
                 self.logger.debug("Hangup button tapped")
                 self.endCallButtonTapped()
         }
-        self.hangUpButtonViewModel.accessibilityLabel = "End call"
+        hangUpButtonViewModel.accessibilityLabel = self.localizationProvider.getLocalizedString(
```
JoshuaLai, line 94: Just a small call out here: The previous accessibility label was saying "End call" and now you are saying "Leave call" I assume this is intended

JoshuaLai (approved):

AzureCommunicationUI/AzureCommunicationUITests/Presentation/Calling/ParticipantGridsViewModelTests.swift
```diff
@@ -298,7 +298,7 @@ class ParticipantGridViewModelTests: XCTestCase {
         let state = makeRemoteParticipantState(count: 2)
         let callingState = CallingState(status: .connected)
         let accessibilityProvider = AccessibilityProviderMocking()
-        let expectedAnnouncement = "2 participants joined the meeting"
+        let expectedAnnouncement = "AzureCommunicationUI.CallingView.MutiplePeopleJoined"
```
vhuseinova-msft, line 301: Could you switch to LocalizationProvider usage so we can check that the used parameter is also correct?

vhuseinova-msft (approved):
Minor comment, otherwise lgtm
Tested.

AzureCommunicationUI/AzureCommunicationUITests/Presentation/Calling/ParticipantGridsViewModelTests.swift
jimchou-dev: You can use LocalizableKey enum with rawValue

vhuseinova-msft: But in this case ParticipantGridViewModel will still return a key as it uses LocalizationProviderMocking

PR #87, opened 2022-03-22
[Feature] Added logics to hide iOS home indicator when screenshare is presented during a call

kevinyulu:
## Purpose
Added logics to hide iOS home indicator when screenshare is presented during a call

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[x] Other... Please describe: UX improvement
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code
* Launch the app, join a call (with remote participant), either group call or Teams call is ok
* Turn on screenshare (done by remote participant)
* Follow **What to check** section
* Turn off screenshare (done by remote participant)
* Follow **What to check** section

## What to Check
* Check if the iOS indicator are hidden when screen share is presented on the screen during a call, in portrait and landscape mode.
* Check if the iOS indicate are shown when switched to other screen such as Setup view, Participant / Audio Drawer, or exit overlay.

### DIFF
--- a/AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Container/ContainerUIHostingController.swift
+++ b/AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Container/ContainerUIHostingController.swift
@@ -10,10 +10,12 @@ class ContainerUIHostingController: UIHostingController<ContainerUIHostingContro
     class EnvironmentProperty {
         @Published var supportedOrientations: UIInterfaceOrientationMask
         @Published var isProximitySensorOn: Bool
+        @Published var prefersHomeIndicatorAutoHidden: Bool
 
         init() {
             self.supportedOrientations = .portrait
             self.isProximitySensorOn = false
+            self.prefersHomeIndicatorAutoHidden = false
         }
     }
 
@@ -79,6 +81,16 @@ class ContainerUIHostingController: UIHostingController<ContainerUIHostingContro
             .sink(receiveValue: { isEnable in
                 UIDevice.current.toggleProximityMonitoringStatus(isEnabled: isEnable)
             }).store(in: cancelBag)
+
+        environmentProperties
+            .$prefersHomeIndicatorAutoHidden
+            .receive(on: RunLoop.main)
+            .sink(receiveValue: { [weak self] shouldHide in
+                guard let strongSelf = self else {
+                    return
+                }
+                strongSelf._prefersHomeIndicatorAutoHidden = shouldHide
+            }).store(in: cancelBag)
     }
 
     private func haltSetupViewOrientation(containerView: ContainerView) {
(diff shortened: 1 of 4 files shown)


jimchou-dev (commented):
Tested, work as described, just one concern since we are fading out the home indicator bar, there is now an odd grey area when in landscape view (see screenshot). Should we extend the participant view to the bottom?
image

kevinyulu:
> Tested, work as described, just one concern since we are fading out the home indicator bar, there is now an odd grey area when in landscape view (see screenshot). Should we extend the participant view to the bottom?
> image

@jimchou-dev yes this is a good call. There is currently an UI issue in zooming when in the double tap to zoom is performed in landscape, if the size of the screenshare view overlaps with iOS indicator. This bug is logged in the ui bug ticket [BUG 2767250] (will work on this asap). As a temporary solution the height of the screenshare has been reduced to so the view does not overlap with iOS home indicator.

jimchou-dev (approved):
Tested, LGTM

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Container/ContainerUIHostingController.swift
```diff
@@ -110,6 +122,19 @@ class ContainerUIHostingController: UIHostingController<ContainerUIHostingContro
...
                 }
+                .onPreferenceChange(PrefersHomeIndicatorAutoHiddenPreferenceKey.self) {
+                    self.environmentProperties.prefersHomeIndicatorAutoHidden = $0
+                }
         }
     }
+
+    // MARK: Prefers Home Indicator Auto Hidden
+
+    private var _prefersHomeIndicatorAutoHidden: Bool = false {
```
AmyL219, line 133: Curious why do we have seperate setter and getter, could we use one property though?

kevinyulu: @AmyL219 
Currently there isn't a way in SwiftUI to configure the `prefersHomeIndicatorAutoHidden` per screen so we'd need to use a work-around to hide the iOS home indicator on a particular screen in SwiftUI.

PreferenceKey (SwiftUI) --> `_prefersHomeIndicatorAutoHidden` --> prefersHomeIndicatorAutoHidden (UIKit property, get only)

Also since this `prefersHomeIndicatorAutoHidden` is a get only property, we'd need a variable that would take the update from preference and pass this update to UIViewController.

To build the connection, this variable `_prefersHomeIndicatorAutoHidden` does a few things:
1. connects the PreferenceKey from SwiftUI to the UIKit property `prefersHomeIndicatorAutoHidden`
2. notifies the UIViewController that the property `prefersHomeIndicatorAutoHidden` has been changed, when there is a PreferenceKey updates from SwiftUI.

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Container/ContainerUIHostingController.swift
```diff
@@ -79,6 +81,16 @@ class ContainerUIHostingController: UIHostingController<ContainerUIHostingContro
             .sink(receiveValue: { isEnable in
                 UIDevice.current.toggleProximityMonitoringStatus(isEnabled: isEnable)
             }).store(in: cancelBag)
+
+        environmentProperties
+            .$prefersHomeIndicatorAutoHidden
+            .receive(on: RunLoop.main)
+            .sink(receiveValue: { [weak self] shouldHide in
+                guard let strongSelf = self else {
```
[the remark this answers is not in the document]
kevinyulu: sure will update

AmyL219 (approved):

PR #88, opened 2022-03-24
[Refactor] Improve bridging RendererView into SwiftUI

vhuseinova-msft:
## Purpose

* Added VideoRendererUIView for correct bridging RendererView into SwiftUI
* Removed the workaround in ParticipantGridCellView

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

## What to Check
* Participants grid and videos are rendered correctly on the Call screen
* Local video is displayed on the Setup and Calling screens

### DIFF
--- a/AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/Cell/ParticipantGridCellView.swift
+++ b/AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/Grid/Cell/ParticipantGridCellView.swift
@@ -12,15 +12,12 @@ struct ParticipantGridCellView: View {
     let getRemoteParticipantRendererView: (RemoteParticipantVideoViewId) -> ParticipantRendererViewInfo?
     let rendererViewManager: RendererViewManager?
     @State var displayedVideoStreamId: String?
-    @State var isVideoChanging: Bool = false
     let avatarSize: CGFloat = 56
 
     var body: some View {
         Group {
             GeometryReader { geometry in
-                if isVideoChanging {
-                    EmptyView()
-                } else if let rendererViewInfo = getRendererViewInfo() {
+                if let rendererViewInfo = getRendererViewInfo() {
                     let zoomable = viewModel.videoViewModel?.videoStreamType == .screenSharing
                     ParticipantGridCellVideoView(videoRendererViewInfo: rendererViewInfo,
                                                  rendererViewManager: rendererViewManager,
@@ -32,21 +29,6 @@ struct ParticipantGridCellView: View {
                     avatarView
                         .frame(width: geometry.size.width,
                                height: geometry.size.height)
-                }
-            }
-        }
-        .onReceive(viewModel.$videoViewModel) { model in
-            let cachedVideoStreamId = displayedVideoStreamId
-            if model?.videoStreamId != displayedVideoStreamId {
-                displayedVideoStreamId = model?.videoStreamId
-            }
-
-            if model?.videoStreamId != cachedVideoStreamId,
-               model?.videoStreamId != nil {
-                // workaround to force rendererView being recreated
-                isVideoChanging = true
-                DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
-                    isVideoChanging = false
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


kevinyulu (commented):
Tested in iPhone XR iOS 15.3.1. The video shows in portrait. For some reason when I switch to landscape, the app is not responding on my device.

AmyL219 (commented):
I tried the constraints to resizing the rendererView, but it would caused the renderView failed in rendering in the old branch. 
So had to call `update(rendererView)` to resize the view manually by assigning frame. And we have to update the frame even if the `self.rendererView == rendererView`. Because the size of the VideoRenderView would be update by calling `UpdateUIView` and we need to update the frame manually every-time when the size change. 
But this would cause another issue which I have described inside the [document]([SwiftUIParticipantVideoRenderView.docx](https://microsoft-my.sharepoint-df.com/:w:/p/longamy/ETZauZ84HjxFrxoJDQvRp2UBLMAx1YQWk7FxRgH1dDrIxg?e=R3LcgY&isSPOFile=1&OR=Teams-HL&CT=1647905713375&params=eyJBcHBOYW1lIjoiVGVhbXMtRGVza3RvcCIsIkFwcFZlcnNpb24iOiIyOC8yMjAzMDcwMTYxMCJ9)).

vhuseinova-msft:
> I tried the constraints to resizing the rendererView, but it would caused the renderView failed in rendering in the old branch.
> So had to call `update(rendererView)` to resize the view manually by assigning frame. And we have to update the frame even if the `self.rendererView == rendererView`. Because the size of the VideoRenderView would be update by calling `UpdateUIView` and we need to update the frame manually every-time when the size change.
> But this would cause another issue which I have described inside the document.

@AmyL219 yes, I've removed the constraints usage after the @kevinyulu'c comment and changed it to autoresizingMask that should guarantee that the view will be resized as expected. In this case, I think we shouldn't set frame once more when self.rendererView == rendererView.
Also, this solution is not the same as was implemented in DebugRenderView branch so I haven't got the same issue with the gray view that isn't updated to the video.  Please let me know if you will notice any issues with the layout after the last commit.

AmyL219:
For autoresizingMask, it updates the subviews when it's bounds updated, see the [AutoresizingMask](https://developer.apple.com/documentation/uikit/uiview/1622559-autoresizingmask). So in this case, we should set the `VideoRendererUIView` autoresizing mask, because we want it resize the renderView which is the subview of `VideoRendererUIView`

in my test observation, the gridView would call `makeUIview` to make a new `VideoRendererUIView` whenever the resizing is happening. Not sure if autoresizing mask is working. I think we still need to test the scenario when `makeUIview` is not called but `updateUIView` is called for renderView resizing though.

vhuseinova-msft:
> For autoresizingMask, it updates the subviews when it's bounds updated, see the [AutoresizingMask](https://developer.apple.com/documentation/uikit/uiview/1622559-autoresizingmask). So in this case, we should set the `VideoRendererUIView` autoresizing mask, because we want it resize the renderView which is the subview of `VideoRendererUIView`
> 
> in my test observation, the gridView would call `makeUIview` to make a new `VideoRendererUIView` whenever the resizing is happening. Not sure if autoresizing mask is working. I think we still need to test the scenario when `makeUIview` is not called but `updateUIView` is called for renderView resizing though.

As per documentation, we don't have to update the superview as we need to set this only for our subview and the subview will be updated when bounds for the superview is changed. 

 `updateUIView` is called a couple of times (1 after makeUIview with frame = .zero and other times without makeUIview but with updates to frame) when there are a lot of participants and:
1) a participant, that is not present on the grid, starts to talk
2) screen orientation is changed

I've tested both cases on iPhone 13 mini and a couple of simulators, and haven't noticed any differences in UI when` self.rendererView?.frame = frame` is added when `self.rendererView == rendererView` or not.

AmyL219 (approved):
Okey, I haven't seen issue in my local testing. Let's keep an eye on this in the bug bash.

kevinyulu (approved):
tested, LGTM

PR #89, opened 2022-03-25
[Bug] Demo App AuthenticationHelper never called back in case of error

JoshuaLai:
## Purpose
Bug in our demo app, if the token fetcher fails for any reason we were never responding to the error, so the composite would freeze
Token struct decodable looks like it was also out of date. 
image

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```


AzureCommunicationUI/AzureCommunicationUIDemoApp/AuthenticationHelper.swift
```diff
@@ -9,21 +9,29 @@ import AzureCommunicationCalling
 class AuthenticationHelper {
     static func getCommunicationToken(tokenUrl: URL) -> TokenRefresher {
         struct TokenResponse: Decodable {
+            let userAndToken: UserAndTokenResponse
+            let endpointUrl: String?
+        }
+        struct UserAndTokenResponse: Decodable {
             let token: String
+            let expiresOn: String?
+            let user: String?
```
kevinyulu, line 18: Should this user be an object type? I looked into the response json from the url. looks like this user is an object type. this is causing the decoding to fail from my side.

AmyL219:
My token generation Url response is different from yours.So I couldn't login with token fetch url with the change. I think the response is generally apply to different Azure Function setting. For example, I only request token for `'voip'`

kevinyulu (dismissed):
Checked with @JoshuaLai about the case of url response getting `401` due to authentication and page redirect before seeing the json response. I think this is outside the scope of this ticket and can be discussed separately. PR looks good.

JoshuaLai:
> 



> My token generation Url response is different from yours.So I couldn't login with token fetch url with the change. I think the response is generally apply to different Azure Function setting. For example, I only request token for `'voip'` Screen Shot 2022-03-30 at 9 33 37 AM

Oh interesting I guess it makes sense that different urls return different bodies. Maybe I undo the change to the decodable object?

kevinyulu (approved):

palatter (approved):
Looks good and tested.

PR #90, opened 2022-03-26
[Feature] Added XCUITest for join call and leave call related UX

kevinyulu:
## Purpose
In this PR, I added a few XCUITests for join call related UX

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

Run the XCUITest in the Xcode

## What to Check
Run the XCUITest in the Xcode and check whether if there is any fail tests


JoshuaLai (dismissed):
lgtm

vhuseinova-msft (dismissed):

AzureCommunicationUI/AzureCommunicationUIDemoAppUITests/AzureCommunicationUIDemoAppUITests.swift
```diff
@@ -93,12 +147,76 @@ class AzureCommunicationUIDemoAppUITests: XCUITestBase {
         }
     }
 
+    // MARK: Private / helper functions
+
+    private func tapInterfaceFor(_ interface: CompositeSampleInterface) {
+        app?.buttons[interface.name].tap()
```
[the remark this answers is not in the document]
kevinyulu: yeah that's a good call. My thought of use helper functions was actually to make them represent the general UITest actions and to reduce the duplications of code with identical button name strings. I am also going to do some refactors on the other buttons in the next UITest tickets to reduce # of duplicated button names usages in the file.

AzureCommunicationUI/AzureCommunicationUIDemoAppUITests/AzureCommunicationUIDemoAppUITests.swift
```diff
@@ -15,73 +44,98 @@ class AzureCommunicationUIDemoAppUITests: XCUITestBase {
     }
 
     func testCallCompositeLaunch() {
```
AmyL219, line 46: Not sure if there is any naming convenient for UI test cases? Should we reach a consensus for the UI test naming convenient?

AzureCommunicationUI/AzureCommunicationUIDemoAppUITests/AzureCommunicationUIDemoAppUITests.swift
```diff
@@ -93,12 +147,76 @@ class AzureCommunicationUIDemoAppUITests: XCUITestBase {
...
+
+    private func tapInterfaceFor(_ interface: CompositeSampleInterface) {
+        app?.buttons[interface.name].tap()
+    }
+
+    private func tapMeetingType(_ meetingType: CompositeMeetingType) {
+        app?.buttons[meetingType.name].tap()
+    }
+
+    private func tapEnabledButton(buttonName: String, shouldWait: Bool) {
```
AmyL219, line 160: Should these helper method be wrapper into a separated helper class? Therefore we can reuse these functions in different class?

AzureCommunicationUI/AzureCommunicationUIDemoAppUITests/AzureCommunicationUIDemoAppUITests.swift
kevinyulu: yeah we can discuss the test function names with the group.

AzureCommunicationUI/AzureCommunicationUIDemoAppUITests/AzureCommunicationUIDemoAppUITests.swift
kevinyulu: yes I can refactor them into a new helper class or any existing helpers classes.

AmyL219 (dismissed):

AzureCommunicationUI/AzureCommunicationUIDemoAppUITests/AzureCommunicationUIDemoAppLaunchTests.swift
```diff
@@ -0,0 +1,153 @@
...
+            XCTFail("No App launch")
+            return
+        }
+
+        tapInterfaceFor(.swiftUI)
+        tapMeetingType(.teamsCall)
+        tapEnabledButton(accesiibilityLabel: .startExperienceAccessibilityLabel, shouldWait: true)
+        toggleSetupScreenControlButtons()
+        tapButton(accesiibilityLabel: .joinCallAccessibilityLabel, shouldWait: true)
+        tapButton(accesiibilityLabel: .hangupAccessibilityLabel, shouldWait: true)
```
AmyL219, line 78: typo: accesiibilityLabel -> accessibilityLabel
And should we use AccessibilityLabel? Rather than accessibilityIdentifier

kevinyulu: @AmyL219 I actually was trying to use the accessibility labels created in `LocalizationKeys` as the `accessibilityIdentifier` for the app UI elements. I can update the parameter names to  `tapButton(accesiibilityIdentifier: shouldWait)`

AmyL219: Yeah, `tapButton(accessibilityIdentifier:` makes more sense.

AzureCommunicationUI/AzureCommunicationUIDemoAppUITests/Utilities/XCUITestBase.swift
```diff
@@ -4,7 +4,104 @@
...
+    var app: XCUIApplication?
+}
+
+extension XCUITestBase {
+
+    /// Taps the button that matches with the given name
+    /// - Parameters:
+    ///   - accesiibilityLabel: accessibility label of the button
+    ///   - shouldWait: determienes whether app should wait for the tap test to complete
+    func tapEnabledButton(buttonName: String, shouldWait: Bool) {
```
vhuseinova-msft, line 48: `func tapEnabledButton(buttonName: String, shouldWait: Bool)`
and
`tapButton(buttonName: String, shouldWait: Bool)`
can be made private as they are not used anywhere but in  `func tapButton(accessibilityIdentifier: LocalizationKey, shouldWait: Bool)` and `func tapEnabledButton(accessibilityIdentifier: LocalizationKey, shouldWait: Bool)`

AzureCommunicationUI/AzureCommunicationUIDemoAppUITests/AzureCommunicationUIDemoAppCallTests.swift
```diff
@@ -0,0 +1,88 @@
...
+    /// Toggles the leave call overlay  in the calling screen
+    private func toggleLeaveCallOverlay(leaveCall: Bool) {
+        guard let app = app else {
+            XCTFail("No App launch")
+            return
+        }
+
+        tapButton(accessibilityIdentifier: .hangupAccessibilityLabel, shouldWait: true)
+        tapButton(accessibilityIdentifier: .cancelAccssibilityLabel, shouldWait: true)
+        tapButton(accessibilityIdentifier: .hangupAccessibilityLabel, shouldWait: true)
```
AmyL219, line 80: why do we need tap  hangup button -> cancel button -> hangup button here?

AzureCommunicationUI/AzureCommunicationUIDemoAppUITests/Utilities/XCUITestBase.swift
kevinyulu: added

AzureCommunicationUI/AzureCommunicationUIDemoAppUITests/Utilities/XCUITestBase.swift
```diff
@@ -4,7 +4,104 @@
...
+
+    /// Taps the enabled button that matches with the given accessibility label
+    /// - Parameters:
+    ///   - accesiibilityLabel: accessibility label of the button
+    ///   - shouldWait: determienes whether app should wait for the tap test to complete
+    func tapEnabledButton(accessibilityIdentifier: LocalizationKey, shouldWait: Bool) {
+        tapEnabledButton(buttonName: accessibilityIdentifier.rawValue, shouldWait: shouldWait)
+    }
+
+    func tapInterfaceFor(_ interface: CompositeSampleInterface) {
```
AmyL219, line 88: Uhhh. what if these methods
```
tapInterfaceFor
tapMeetingType
```
 are called after the composite launch?

Could refactr them later though

AzureCommunicationUI/AzureCommunicationUIDemoAppUITests/Utilities/XCUITestBase.swift
```diff
@@ -4,7 +4,104 @@
...
+            switch self {
+            case .groupCall:
+                return "Group Call"
+            case .teamsCall:
+                return "Teams Meeting"
+            }
+        }
+    }
+
+    var app: XCUIApplication?
```
vhuseinova-msft, line 39: I'm wondering if we can use implicitly unwrapped optional here instead?

AzureCommunicationUI/AzureCommunicationUIDemoAppUITests/AzureCommunicationUIDemoAppCallTests.swift
kevinyulu: this is to test if the cancel works first before ending call

AzureCommunicationUI/AzureCommunicationUIDemoAppUITests/Utilities/XCUITestBase.swift
kevinyulu: yes I think we'd need one app at least to run the ui test. updated.

AzureCommunicationUI/AzureCommunicationUIDemoAppUITests/Utilities/XCUITestBase.swift
kevinyulu: I think it would fail the test case as those button will be nil when queried at other screen. I created these helper function to make the uitkst more readable and put it in the base case as pretty much all the tests will use these methods before entering the composite. I can add comment to make sure those needs only to be called before entering composite.

AzureCommunicationUI/AzureCommunicationUIDemoAppUITests/AzureCommunicationUIDemoAppCallTests.swift
AmyL219: Okey, that makes sense. But do we need to duplicately test it every time though? Wouldn't this increase the whole testing time though? I think having one test to verify the function should be good enough.

AzureCommunicationUI/AzureCommunicationUIDemoAppUITests/Utilities/XCUITestBase.swift
AmyL219: I feel like it's better to make these two methods as a `makeSut` method which would include the `launchApp` inside each UITests class

AmyL219 (dismissed):
We could having a refactoring ticket to optimize code

vhuseinova-msft (dismissed):

AmyL219 (dismissed):

vhuseinova-msft (dismissed):

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/CallingViewComponent/ConfirmLeaveOverlayView.swift
```diff
@@ -42,9 +41,11 @@ struct ConfirmLeaveOverlayView: View {
 
     var leaveCallButton: some View {
         return PrimaryButton(viewModel: viewModel.getLeaveCallButtonViewModel())
+            .accessibility(identifier: LocalizationKey.leaveCallAccssibilityLabel.rawValue)
```
vhuseinova-msft, line 44: Could you update `.accessibility(identifier` to `accessibilityIdentifier(`as .accessibility(identifier will be deprecated in iOS 15.4?

vhuseinova-msft (changes requested):
Could you update `.accessibility(identifier:)` to `accessibilityIdentifier(:)` as `.accessibility(identifier:)` will be deprecated in iOS 15.4?

palatter (approved):
Looks good.

palatter:
> Could you update `.accessibility(identifier:)` to `accessibilityIdentifier(:)` as `.accessibility(identifier:)` will be deprecated in iOS 15.4?

Is the alternative available in our current version? I know some of the newer accessibility method signatures were only available in 15+

vhuseinova-msft:
> > Could you update `.accessibility(identifier:)` to `accessibilityIdentifier(:)` as `.accessibility(identifier:)` will be deprecated in iOS 15.4?
> 
> Is the alternative available in our current version? I know some of the newer accessibility method signatures were only available in 15+

yes,methods like `.accessibility(<ParameterName>:)` that we used because of iOS 13, can be changed to `accessibility<ParameterName>(:)` as they are available for iOS 14. For example: https://developer.apple.com/documentation/swiftui/view/accessibilityidentifier(_:)

kevinyulu:
> .accessibility(identifier:)

discussed with @vhuseinova-msft, the accessibility function code change will be addressed in the other PR https://github.com/Azure/communication-ui-library-ios/pull/129.

vhuseinova-msft (approved):

PR #95, opened 2022-03-29
[Feature] Added basic XCUITests for leave call

kevinyulu:
## Purpose
Added basic XCUITests for leave call

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[x] Other... Please describe: XCUITests
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code
Run the AzureCommunicationUIDemoAppCallTests
```
```

## What to Check
* check if all the test cases pass.

### DIFF
--- a/AzureCommunicationUI/AzureCommunicationUIDemoAppUITests/AzureCommunicationUIDemoAppCallTests.swift
+++ b/AzureCommunicationUI/AzureCommunicationUIDemoAppUITests/AzureCommunicationUIDemoAppCallTests.swift
@@ -1,0 +1,85 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import XCTest
+
+class AzureCommunicationUIDemoAppCallTests: XCUITestBase {
+
+    override func setUp() {
+        super.setUp()
+        app = XCUIApplication()
+        app?.launch()
+    }
+
+    // MARK: End call tests
+
+    func testCallCompositeEndCallGroupCallSwiftUI() {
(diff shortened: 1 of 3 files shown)


AzureCommunicationUI/AzureCommunicationUIDemoAppUITests/AzureCommunicationUIDemoAppCallTests.swift
```diff
@@ -0,0 +1,85 @@
...
+        toggleLeaveCallOverlay(leaveCall: true)
+    }
+
+    func testCallCompositeEndCallTeamsCallUIKit() {
+        guard app != nil else {
+            XCTFail("No App launch")
+            return
+        }
+
+        tapInterfaceFor(.uiKit)
```
AmyL219, line 61: Do we need to both test UIkit and SwiftUI demo app? They are testing the same thing fundamentally?
I wonder if this would increase the testing time though. .

kevinyulu: yeah it would increase the testing time. If SwiftUI is the main one we'd test, we can skip the same tests on UIKit but only to keep the UITest for the .uiKit button

kevinyulu: I am going to merge this PR into the feature branch and address this change.

PR #103, opened 2022-04-01
[Feature] Bumping minimum target to iOS 14

JoshuaLai:
## Purpose
Updating minimum taget to iOS 14

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## What to Check
Full regression test is needed


palatter (commented):
We should update the podspec in root as well with new min version.

AmyL219:
Should we start a thread with Jorge about the update in the channel, to make sure we are admitted for minimum version bumping

JoshuaLai:
> We should update the podspec in root as well with new min version.

Good catch. Updated!

JoshuaLai:
> Should we start a thread with Jorge about the update in the channel, to make sure we are admitted for minimum version bumping

For sure. Maybe I can revert this to a draft for now so it is not accidentally merged

vhuseinova-msft (changes requested):
Could you also update accessibility functions usage from
`accessibility(<parameterName>:)` to `accessibility< parameterName >`? [As Apple has added updated functions](https://developer.apple.com/documentation/swiftui/modifiedcontent/accessibilityhidden(_:)-3g7oh) with iOS 14 and as we don't need to support iOS 13 anymore we can update those methods. 
For example:
`accessibility(hidden:)` -> `accessibilityHidden(_:)`

vhuseinova-msft:
> > Could you also update accessibility functions usage from `accessibility(<parameterName>:)` to `accessibility< parameterName >`? [As Apple has added updated functions](https://developer.apple.com/documentation/swiftui/modifiedcontent/accessibilityhidden(_:)-3g7oh) with iOS 14 and as we don't need to support iOS 13 anymore we can update those methods. For example: `accessibility(hidden:)` -> `accessibilityHidden(_:)`
> 
> I can update the hidden methods. Do you know which other ones there are? Also do we need to update these methods? I don't see Xcode complaining about deprecated methods so not sure what the value is for doing all this.

you can search with "accessibility(" in our project and all of those methods should be updated. I think we neeed a they were used only because of support of iOS 13 (they deprecated in iOS 15) and the new one were introduced in iOS 14.  Also, we can create a separate task for updating those if you want.

JoshuaLai:
@palatter @vhuseinova-msft we are going to need to regression the accessibility too when we do the bug bash.

vhuseinova-msft (approved):

AmyL219 (approved):

PR #122, opened 2022-04-06
[Bugfix] Fix accessibility for localVideoFullscreenView

vhuseinova-msft:
## Purpose

* Hide videoGridView for accessibility when the current user's avatar is shown and no name is provided

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

## What to Check
When current user is the only participant on the call:
* when camera is off and no user's name, VoiceOver skips avatar
* when camera is off and user's name isn't empty, VoiceOver uses display name
* when camera is on, switch camera button is accessible


jimchou-dev (dismissed):

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/CallingViewModel.swift
```diff
@@ -10,6 +10,7 @@ class CallingViewModel: ObservableObject {
     @Published var isLobbyOverlayDisplayed: Bool = false
     @Published var isConfirmLeaveOverlayDisplayed: Bool = false
     @Published var isParticipantGridDisplayed: Bool
+    @Published var isVideoGridViewAccessibilityAvailable: Bool = false
```
AmyL219, line 13: Is that possible to remove the default value `false`? since it's going to update the value in the init method

vhuseinova-msft: @AmyL219 no, if the `false` value is removed, there will be a buildtime error as `isVideoGridViewAccessibilityAvailable` is initializad in a separate method

AmyL219: Would it work with force unwrap mark? `@Published var isVideoGridViewAccessibilityAvailable: Bool!`

vhuseinova-msft: @AmyL219 yes, updated

palatter (approved):
Looks good and tested!

AmyL219 (approved):

PR #125, opened 2022-04-07
[Feature] Create AvatarManager in Dependancy Container

JoshuaLai:
## Purpose
Create the `AvatarManager` and inject the manager into the `SetupView` and `CallingView` to be used. 
Introducing a avatar picker. 
Screen Shot 2022-04-14 at 3 45 29 PM
Screen Shot 2022-04-14 at 3 45 35 PM
Screen Shot 2022-04-14 at 3 45 41 PM
Screen Shot 2022-04-14 at 3 47 03 PM

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```


vhuseinova-msft (approved):

AzureCommunicationUI/AzureCommunicationUI/CallComposite.swift
```diff
@@ -44,6 +44,11 @@ public class CallComposite {
 
         dependencyContainer.registerDependencies(callConfiguration)
         let localizationProvider = dependencyContainer.resolve() as LocalizationProvider
+        let avatarManager = dependencyContainer.resolve() as AvatarManager
+        if let localAvatar = localOptions?.localPersona.avatar {
```
AmyL219, line 48: It's better to keep consistent with `setupColorTheming/setupLocalization ` by abstracting 
```
if let localAvatar = localOptions?.localPersona.avatar {
}
```
to another function, `setupLocalOptions()`

AzureCommunicationUI/AzureCommunicationUI/Presentation/Manager/AvatarManager.swift
```diff
@@ -7,7 +7,12 @@ import Foundation
 import UIKit
 import AzureCommunicationCommon
 
-public class AvatarManager {
+protocol AvatarManager {
```
AmyL219, line 10: Would it be a good idea to rename those manager who get access to view layer with `ViewManager`?
Something like `VideoViewManager`/`AvatarViewManager`, to distinguish the common managers relying on Store and the view managers which get access to view layer

AzureCommunicationUI/AzureCommunicationUIDemoApp/Views/SwiftUIDemoView.swift
```diff
@@ -143,26 +145,32 @@ extension SwiftUIDemoView {
...
-                                                                displayName: envConfigSubject.displayName))
+                                                                displayName: envConfigSubject.displayName),
+                                         localOptions: localOptions)
                 }
             case .teamsMeeting:
                 if envConfigSubject.displayName.isEmpty {
                     callComposite.launch(with: TeamsMeetingOptions(credential: credential,
-                                                                   meetingLink: link))
+                                                                   meetingLink: link),
+                                         localOptions: localOptions)
```
AmyL219, line 168: nick pick: indention

AzureCommunicationUI/AzureCommunicationUIDemoApp/Views/SwiftUIDemoView.swift
```diff
@@ -143,26 +145,32 @@ extension SwiftUIDemoView {
...
-                    callComposite.launch(with: GroupCallOptions(credential: credential,
-                                                                groupId: uuid))
+                    callComposite.launch(with: GroupCallOptions(credential: credential, groupId: uuid),
+                                         localOptions: localOptions)
                 } else {
                     callComposite.launch(with: GroupCallOptions(credential: credential,
                                                                 groupId: uuid,
-                                                                displayName: envConfigSubject.displayName))
+                                                                displayName: envConfigSubject.displayName),
+                                         localOptions: localOptions)
```
AmyL219, line 162: nick pick: indention

JoshuaLai: This indentation is actually correct. It is a second parameter so it should align with the `with:`

AzureCommunicationUI/AzureCommunicationUIDemoApp/Views/SwiftUIDemoView.swift
JoshuaLai: This indentation is actually correct. It is a second parameter so it should align with the `with:`

AzureCommunicationUI/AzureCommunicationUI/CallCompositeOptions/CommunicationULocalDataOptions.swift
```diff
@@ -7,10 +7,20 @@ import Foundation
...
-   let localPersona: CommunicationUIPersonaData
+    let localPersona: CommunicationUIPersonaData
+
+    public init(_ localPersona: CommunicationUIPersonaData) {
+        self.localPersona = localPersona
+    }
 }
 
 public struct CommunicationUIPersonaData {
     let avatar: UIImage?
```
iaulakh, line 18: I think, we can say avatarImage, as avatar is whole component in Fluent?

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Setup/SetupViewComponent/PreviewAreaView.swift
```diff
@@ -23,6 +24,7 @@ struct PreviewAreaView: View {
 
     var localVideoPreviewView: some View {
         return LocalVideoView(viewModel: viewModel.localVideoViewModel,
+                              avatarManager: avatarManager,
```
iaulakh, line 27: As AvatarManager is going to have both remote and local participant data options. In my opinion, we can inject only localDataOptions as this view does not need remote data.

AzureCommunicationUI/AzureCommunicationUI/Presentation/Manager/AvatarViewManager.swift
```diff
@@ -7,7 +7,12 @@ import Foundation
 import UIKit
 import AzureCommunicationCommon
 
-public class AvatarManager {
+protocol AvatarViewManager {
```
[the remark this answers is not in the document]
waynemo: Since the custom avatar image and display name override are only related to presentation, I think we can keep it out of the redux layer.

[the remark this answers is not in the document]
JoshuaLai: > So are you going to remove the `displayName: String?` from `LocalUserState`? Or keep `localUserState` and `CommunicationUIPersonaData` at same time? AvatarViewManager also manage the display name of the localUser? If so, would that make sense call it AvatarViewManager?

I think this is a good point. So my first thought it maybe we should revisit the name `LocalUserState` to something like `HardwareDeviceState` because it only represents what the hardware device state is (camera / microphone / video stream). And yes I do think we should now remove the `displayName` from there and now only rely on the new `PersonaData`. 
I suggest we create a refactor task to track all this. @AmyL219 @iaulakh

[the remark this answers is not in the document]
waynemo: We can still have `displayName` from the server/CallingSDK in the redux state. It can be the default value used if there isn't anything specified for custom display name rendering.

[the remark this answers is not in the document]
iaulakh: I agree with keeping persona injection outside redux layer (I do not think bitmap should be part of redux state because of raw data size). For group call options `displayName`, we can move outside redux state, however, currently it is optional and we do not need to refactor working/tested code as part of this PR. 

We can create future refactoring ticket if required.

AmyL219 (approved):

AzureCommunicationUI/AzureCommunicationUIDemoApp/Assets.xcassets/Avatars/fox.imageset/Contents.json
```diff
@@ -0,0 +1,23 @@
...
+      "idiom" : "universal",
+      "scale" : "1x"
+    },
+    {
+      "filename" : "fox4x-1.png",
+      "idiom" : "universal",
+      "scale" : "2x"
+    },
+    {
+      "filename" : "fox4x.png",
```
AmyL219, line 14: For images, shouldn't we use PDF?

JoshuaLai: Yes we should be using SVG. I am going to follow up with Alex on this. Since its just going into the demo app for the time being I didn't put a priority on it. I will have a follow up task

PR #126, opened 2022-04-07
[Feature] Localization: auto OS language detection

jimchou-dev:
## Purpose

* Automatically detect OS language if OS language is supported by Contoso app localization and supported by Composite
  1. Will check languageCode ll-CC (language-region) if supported
  2. If ll-CC not supported, fall back to check languageCode ll (language) if supported
  3. Default language is `en` (English)
* When specified languageCode in `LocalizationConfiguration` will override the auto detected OS language
* Added localization for Demo app with languages: `en`, `zh-Hant`, `fr`, `fr-CA` (currently `fr-CA` is not yet supported by Composite so will read Contoso `fr-CA` override string, but fallback Composite to `fr`)

```
[ ] Yes
[ x ] No
```

## Pull Request Type

```
[ ] Bugfix
[ x ] Feature
[ ] Other... Please describe:
```

## What to Check
* Switch system OS language to French or Traditional Chinese, then open the app, go too `Settings` page select `Language` as `auto` (Detect OS locale)
* Should see `Join Call` label in setup screen to show `Join Call (override <language>)` for French and Traditional Chinese because both Contoso (Demo app) and Composite supports French and Traditional Chinese


AzureCommunicationUI/AzureCommunicationUI/Presentation/Style/LocalizationProvider.swift
```diff
@@ -23,6 +23,26 @@ class AppLocalizationProvider: LocalizationProvider {
...
+                  self.isLanguageSupportedByContoso(collatorIdentifier
+                    .replacingOccurrences(of: "-\(regionCode)", with: "")) {
+            let languageRegionCode = collatorIdentifier
+                .replacingOccurrences(of: "-\(regionCode)", with: "")
+            self.languageCode = languageRegionCode
+
+        } else if let systemLanguageCode = Locale.current.languageCode,
+                  self.isLanguageSupportedByContoso(systemLanguageCode) {
+            self.languageCode = systemLanguageCode
+        }
```
waynemo, line 45: Could we use this instead to find the intersection of app supported localizations and iOS preferences?
https://developer.apple.com/documentation/foundation/bundle/1413220-preferredlocalizations

waynemo (dismissed):

AzureCommunicationUI/AzureCommunicationUI/Presentation/Style/LocalizationProvider.swift
```diff
@@ -77,4 +97,9 @@ class AppLocalizationProvider: LocalizationProvider {
                                  value: key,
                                  comment: key)
     }
+
+    private func isLanguageSupportedByContoso(_ systemLanguage: String) -> Bool {
```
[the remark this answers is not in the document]
jimchou-dev: I think using "language" is consistent with Apple as the documentation for preferred language calls it "language id" and not locale.

AzureCommunicationUI/AzureCommunicationUI/Presentation/Style/LocalizationProvider.swift
jimchou-dev: We can apply a for loop to contain the similar logic, but i think it will be 2 separate for-loops:
- First loop: Check preferred language with region code in order
- Second loop: Check preferred language without region code
@waynemo Does this logic make sense? If so I'll go ahead with the implementation

waynemo: I'm thinking we could just use `Bundle.main.preferredLocalizations.first` to get the preferred localization based on supported localizations and iOS settings. I think this would return the language identifier that the app would be using. From this value we can check if it is in the library supported localizations.

AzureCommunicationUI/AzureCommunicationUI/Presentation/Style/LocalizationProvider.swift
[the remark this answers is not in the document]
waynemo: In that case I think we need to differentiate the following to prevent ambiguity:

1. Language Identifier (includes Language Code + optional Region Code)
2. Language Code only

Perhaps we can explicitly use "languageIdentifier" where we are dealing with (1) and "languageCode" where we are dealing with (2).

AzureCommunicationUI/AzureCommunicationUI/Presentation/Style/LocalizationProvider.swift
jimchou-dev: Yeah we can use `Bundle.main.preferredLocalizations.first` instead of `Locale.current.collatorIdentifier`, but based on some testing it's the same thing. So we're just going to be using the first language, not checking all the language in the preferredLocalizations list

waynemo: I'm not sure if it makes sense to check all the preferredLocalization values. 

For example, if the first preferredLocalization is `ar-ar`. That would mean the app should be using `ar-ar` based on iOS settings and app supported localizations. If `ar-ar` is not supported by the library, we need a fallback. 

There might be a second preferredLocalization `fr-fr`, which is supported by the library. But it's unclear if the app would want to use it even though the rest of the app is using `ar-ar`. I'm thinking the logic is simpler if we just try to detect based on the app's current language, and if that doesn't match have a single fallback (English).

AmyL219 (commented):
There is some inconsistency in reading the customized string from System and our LocalizationPrvoider. Is this scenario expected?
I add a new language `Hebrew` inside the demo App. 
I set my device preferredLanguageId to `he-CA`. 
Add localization string of hebrew for demo app's `Start Experience` button and `Join call` button. 

Result: The demo app's `Start Experience` button would be replaced with word in the Localizable.strings. 
The composite `Join Call` button wouldn't read the string from the `Hebrew` Localizable.strings but reading the default english string

jimchou-dev:
> There is some inconsistency in reading the customized string from System and our LocalizationPrvoider. Is this scenario expected? I add a new language `Hebrew` inside the demo App. I set my device preferredLanguageId to `he-CA`. Add localization string of hebrew for demo app's `Start Experience` button and `Join call` button.
> 
> Result: The demo app's `Start Experience` button would be replaced with word in the Localizable.strings. The composite `Join Call` button wouldn't read the string from the `Hebrew` Localizable.strings but reading the default english string

Remove the logic to check if language is also supported by Composite, makes more sense if whatever the system language is, as long as Contoso supports, they could be directly providing key-value in associated Localizable.strings file for overriding strings.

kevinyulu (dismissed):

vhuseinova-msft (dismissed):
Looks good, tested

AzureCommunicationUI/AzureCommunicationUI/Presentation/Style/LocalizationProvider.swift
```diff
@@ -99,4 +113,18 @@ class AppLocalizationProvider: LocalizationProvider {
         }
         return nil
     }
+
+    private func removeLastLanguageComponent(_ languageId: String) -> String {
```
AmyL219, line 117: Would it be clearer to rename the function as `removeRegionCode`?

AmyL219 (dismissed):

vhuseinova-msft (approved):

AmyL219 (approved):

PR #128, opened 2022-04-11
[BugFix] Fix the issue that iphone pro max doesn't show screen share in landscape mode

AmyL219:
## Purpose

* The iphone pro max didn't show screen share in landscape mode is because the rendererView width is around `1480` pt after `initialScale` of scrollView set to 0.5 (`the scrollView frame.width/0.5 ~= 1480`). The CallingSDK rendererView wouldn't be able to render the screen share view once the rendererView width is large around 1400. Will report to CallingSDK.
* This quick fix is to update the `initialScale` of scrollView to `1` when the scrollView is loaded in a screen with one side larger than `850` pt. This would update renderView to load width of `scrollView.frame.size.width/1 ~= 740` in landscape mode for iphone pro max. 
* The workaround limitation is the rendererView would be blur because the rendererView size become smaller and having less pixels for zoom in 
* Other workaround is we fetch the maximum rendererView size (1300*600) for all the device type. And calculate the initialMinZoomScale based on the (1300/scrollView.bounds.size.width). This would have a better user experience when user zoom the picture. But this solution need to confirm with CallingSDK team what's the maximum quality they support as well as a lot of refactoring work in the ZoomableRendererView. Will discuss with @kevinyulu and optimize later. 

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*** Test iphone pro max, iphone pro, iphone 8, iphone Se, ipad with different screen size. See if there is any weired behavior**


kevinyulu (dismissed):
Tested in three devices in teams call and groups call, looks good to merge.
* iPhone 13 Pro max iOS 15 (simulator) both orientations
* iPhone SE 2nd gen iOS 15 (simulator) both orientations 
* iPhone XR iOS 15 (test device) both orientations 
* iPad Pro 12.9 inch (5th gen) iOS 15 (simulator) both orientations

kevinyulu (approved):

JoshuaLai (approved):

PR #138, opened 2022-04-21
[Feature] Calculate initial PIP position correctly based on orientation

palatter:
## Purpose

1. Simplified getting size of pip
2. Calculate the correct position when rotating
3. Show pip under infoheader

* ...

```
[ ] Yes
[X ] No
```

## Pull Request Type

```
[ ] Bugfix
[ X] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...


AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/CallingView.swift
```diff
@@ -59,20 +58,16 @@ struct CallingView: View {
...
-                        .accessibilityHidden(viewModel.isLobbyOverlayDisplayed)
                     if viewModel.isParticipantGridDisplayed {
                         draggableVideoPipView
-                            .padding(.horizontal, -12)
-                            .padding(.vertical, -12)
                             .onAppear {
-                                self.pipPosition = getInitialPipPosition(
-                                    containerBounds: geometry.frame(in: .local),
-                                    pipSize: self.pipSize)
+                                self.pipPosition = getInitialPipPosition(containerBounds: geometry.frame(in: .local))
```
jimchou-dev, line 64: On appear seem to have animation moving tom top left to bottom right corner

AzureCommunicationUI/Podfile.lock
```diff
@@ -61,4 +61,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 1ec5e4fa755a18606e1be3d2a9331f80c3ca3c84
 
-COCOAPODS: 1.11.3
+COCOAPODS: 1.11.2
```
jimchou-dev, line 64: nitpick: Probably don't need to commit this file

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/CallingView.swift
```diff
@@ -234,6 +218,11 @@ struct CallingView: View {
         return boundedPipPosition
     }
 
+    private func getPipSize() -> CGSize {
+        let isPortraitMode = getSizeClass() != .iphoneLandscapeScreenSize
+        return CGSize(width: isPortraitMode ? 72 : 104, height: isPortraitMode ? 104 : 72)
```
jimchou-dev, line 223: Make these value in to variables/constants, and should these size stay the same for iPad?

palatter: Moved into lets. I reached out to @pereiralex regarding this, just waiting on a response.

jimchou-dev (approved):
Tested, LGTM

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/CallingView.swift
palatter: Alex said there are no plans for different size for iPad yet.

kevinyulu:
1. Simplified getting size of pip [OK]
2. Calculate the correct position when rotating  
3. Show pip under infoheader [OK]

For 2) I noticed that the position of the pip view is off the screen when I rotate orientation on i**Pad Pro 12.9 inch simulator** iOS 15.2. video attached

https://user-images.githubusercontent.com/90668345/165395413-e450ff9c-b8a1-4145-ad16-dc98a1fa87e2.mov

palatter:
> 1. Simplified getting size of pip [OK]
> 2. Calculate the correct position when rotating
> 3. Show pip under infoheader [OK]
> 
> For 2) I noticed that the position of the pip view is off the screen when I rotate orientation on i**Pad Pro 12.9 inch simulator** iOS 15.2. video attached
> 
>  rotation.pip.position.issue.iPad.mov

Good catch, I will take a look!

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/CallingView.swift
```diff
@@ -59,20 +58,16 @@ struct CallingView: View {
...
-                            .padding(.horizontal, -12)
-                            .padding(.vertical, -12)
-                            .onAppear {
-                                self.pipPosition = getInitialPipPosition(
-                                    containerBounds: geometry.frame(in: .local),
-                                    pipSize: self.pipSize)
-                            }
-                    }
+                }
+                .onAppear {
```
AmyL219, line 69: onAppear could be called multiple called during the view lifecycle. For example, like device rotate, this on Appear could be called and pipPosition would be reset. Is this what we want?

palatter: We do want it to be called for rotation, as this reset in position mirrors what teams does.

You're right though, it would be called during resume from the background, which is not what we want. I will resolve this.

palatter: I resolved the issue but I discovered an additional issue on iPad. There is a use case where we put the app into split screen mode. Which has 1/3, 1/2 and 2/3 sizes for the app. We need to be able to detect when this change happens so we can reset the pip position. If could detect the container size changing, it could actually be used for both rotation and split screen, but I wasn't yet able to get that working.

I am going to keep working on this issue but in a new PR.

palatter: It now resets on screen size change. This covers rotation and iPad split-screen size changes.

kevinyulu (approved):
tested, lgtm

PR #140, opened 2022-04-25
[BugFix] Fix the audio session not being active after the interruption ended

AmyL219:
## Purpose

* Handle the audio session interruption with the limitation. - The mircophone won't work after the interruption. -The system notification of the interruption ended could be missed by system, so we still need to rely on Callkit integration in future. 
* Technical details:
1. When the interruption is ended, resume the audio session by `audioSession.setCategory` and `audioSession.setActive`. Simply setting audio session active won't work. Need to `setCategory` first and then `overrideOutputAudioPort` after a delay. This would reactive the audio session for the composite.
2. But after the audio session is reactive, the `audioInput` of the call won't work anymore because the CallingSDK `audioInput` wouldn't be effective after the `audioSession` switches back from inactive to reactive. The CallingSDK `audioInput` need to be recreated but there is no api internal/external for this purpose. CallingSDK has the plan to adding the api in future to handle this issue. 
* Refactoring details:
1. Add the local cache of `localUserAudioDeviceState` to filter duplicated states handling
2. Fix the notification handler `handleRouteChange`. Dispatch action `AudioDeviceChangeSucceeded` rather than `AudioDeviceChangeRequested`, because `AVAudioSession.routeChangeNotification` is the notification that `outputAudioPort` already being override successfully. We shouldn't send `AudioDeviceChangeRequested` to `switchAudioDevice` and then `overrideOutputAudioPort` again. Also adding a filter to make sure only the new output is different from current state, an action would be sent

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Awake an incoming call/alarm during the ACS Call


kevinyulu:
Tested awake an incoming call/alarm during the ACS Call. Mostly work as expected. One case I'd like to bring up is that the `localUserState.audioState.device` seems to be overridden by the app state change in one specific scenario, and its competing with the currently selected audio state:

* User enters a call with speaker selected
* In the calling view, change audio to iPhone (`receiver`)
* awake an incoming call during the ACS Call, and stop the incoming call
* see the audio selection changes values a few times from receiver to speaker, and finally set to speaker, which is not the receiver from step 2.

kevinyulu:
Tested awake an incoming call/alarm during the ACS Call. Mostly work as expected. One case I'd like to bring up is that the `localUserState.audioState.device` seems to be overridden by the app state change in one specific scenario, and its competing with the currently selected audio state:

* User enters a call with speaker selected
* In the calling view, change audio to iPhone (`receiver`)
* awake an incoming call during the ACS Call, and stop the incoming call
* see the audio selection changes values a few times from receiver to speaker, and finally set to speaker, which is not the receiver from step 2.

AmyL219:
> Tested awake an incoming call/alarm during the ACS Call. Mostly work as expected. One case I'd like to bring up is that the `localUserState.audioState.device` seems to be overridden by the app state change in one specific scenario, and its competing with the currently selected audio state:
> 
> * User enters a call with speaker selected
> * In the calling view, change audio to iPhone (`receiver`)
> * awake an incoming call during the ACS Call, and stop the incoming call
> * see the audio selection changes values a few times from receiver to speaker, and finally set to speaker, which is not the receiver from step 2.

@kevinyulu, thanks for the testing! I have noticed the similar issue but feel there is limitation on fixing this one. Typically, the system automatically switch to `speaker` when the call is coming and inform us by the notification `handleRouteChange`. 

- When we receive `handleRouteChange`, we know the output has been changed. But we can't distinguish if it's because of incoming call in system, or user switching output device. So simply override the Composite output device in the Redux State and update in the UI to let user know the current device output is the process here.

kevinyulu (dismissed):
tested, lgtm

palatter (dismissed):
I tested and it looks good. I just had the one comment about the delay, as it could be inconsistent?

AzureCommunicationUI/AzureCommunicationUI/Presentation/Manager/AudioSessionManager.swift
```diff
@@ -60,11 +99,33 @@ class AppAudioSessionManager: AudioSessionManager {
...
+                .bluetoothRequested:
+            audioDeviceType = .bluetooth
+        default:
+            audioDeviceType = .receiver
+        }
+
+        activateAudioSessionCategory()
+
+        // Add a delay of setting audioDeviceType, to override the default port from setAudioSessionCategory.
+        DispatchQueue.main.asyncAfter(deadline: .now() + 1) { [weak self] in
```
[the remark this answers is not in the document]
AmyL219: Kind of reliable. See most of the cases handling this scenario need to adding the delay
https://stackoverflow.com/a/33380735

vhuseinova-msft (dismissed):

AzureCommunicationUI/AzureCommunicationUI/Redux/State/LocalUserState.swift
```diff
@@ -95,6 +95,18 @@ class LocalUserState: ReduxState {
...
+            case (.speakerSelected, .speaker),
+                (.bluetoothSelected, .bluetooth),
+                (.headphonesSelected, .headphones),
+                (.receiverSelected, .receiver):
+                return true
+            default:
+                return false
+            }
+        }
+
```
waynemo, line 109: Could we move this logic to the `AudioSessionManager`? The redux layer shouldn't really reference any types from the presentation layer.

waynemo (dismissed):

AzureCommunicationUI/AzureCommunicationUI/Redux/State/LocalUserState.swift
AmyL219: Moved!

palatter:
Looks good!

palatter (approved):

vhuseinova-msft (approved):

PR #147, opened 2022-04-27
[Refactor] ZoomableVideoRenderView optimization and refactoring

AmyL219:
## Purpose

Based on our tests on Pinch and Zoom, Double Tap, Rotation, Background the app, did some refactoring based on these technical details:
* Remove the `private var scrollView = UIScrollView()` inside `ZoomableVideoRenderView`. The `ZoomableVideoRenderView` is a struct and it generates a new UIScrollView whenever it being passed/called. So the `scrollView` may not refer to the UIScrollView being displayed in the view hierarchy. (If needed, the correct way to get reference to the displaying scrollView is from the function `updateUIView` which has reference to the scrollView being displayed.)
* Remove actions inside `updateUIView` because currently there is no data driving UIViewRepresentable to trigger the function, like properties assigned with @State, @Binding, @ObservableObject, etc. All the scrollView zoom in/out are triggered by scrollView itself inside `ZoomableVideoRenderView`, like user action, etc. No external data drive the scrollView update, so no need to setup the `updateUIView` function since it not being called to change scrollView in my testing. 
* For device rotation, the `GridCellView` would call `makeUIView`, which would totally setup the UIViewRepresentable from scratch and return a new uiscrollView for displaying. 

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```


kevinyulu:
@AmyL219  I checked on iPhone XR iOS 15.4 device. I noticed that on the screenshare video is blurry. This is likely due to that `UIScreen.isScreenSmall` returns false for iPhone XR. can we revisit this function or increase the `smallScreenMaxLength` value to a larger value (currently 850)?

jimchou-dev (approved):
Tested, LGTM

kevinyulu (approved):
Discussed with Amy on the screen share blurriness due to initial scrollview scaling for iPhone XR and iPhone 11. As majority of the iOS devices will not have the blurriness issue and this PR passed all the other testing. We will create a new task to accommodate the screen share blurriness issue iPhone XR and iPhone 11 in the future sprint.

AmyL219:
> Discussed with Amy on the screen share blurriness due to initial scrollview scaling for iPhone XR and iPhone 11. As majority of the iOS devices will not have the blurriness issue and this PR passed all the other testing. We will create a new task to accommodate the screen share blurriness issue iPhone XR and iPhone 11 in the future sprint.

We use `850 pt` to distinguish if the device is smaller screen device (use `minimalScale 0.5`) or larger screen device (use `minimalScale 1`). For larger screen device, it would have a bad resolution because of CallingSDK limitation. See [screen size reference](https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/adaptivity-and-layout/). 
However, we found for same point size screen, part of them can have a better resolution: 
- iphone XR/ 11: `414x896 pt (828x1792 px @2x)` display the rendererView with `minimalScale 0.5`
- iphone 11/XS Pro Max: `414x896 pt (1242x2688 px @3x)` **wouldn't** display the rendererView with `minimalScale 0.5`

So we change the distinguish criteria to pixels and use `2550 px` to determine if we should use `minimalScale 0.5`. However, for iphone 12 Pro (1170x2532 px), it displays the rendererView, and for `11" iPad Pro (1668x2388)`, it doesn't display the rendererView. 
So we would keep using `850 pt` to distinguish if we should use `minimalScale 0.5` for now. Until we informed by CallingSDK about the best practice of displaying screenshare view

vhuseinova-msft (approved):

PR #148, opened 2022-04-28
[Feature] Add an event for notifying when participants joined a call

vhuseinova-msft:
## Purpose

* Added a new completion block in CallCompositeEventsHandler for notifying when participants joined a call
* Introduced CompositeRemoteParticipantsManager
* Fixed UIKit typo in the Demo application

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

## What to Check
* when participants are joined or a user joined a call with other participants, didRemoteParticipantsJoin is called in the Demo application

### DIFF
--- a/AzureCommunicationUI/AzureCommunicationUI/CallComposite.swift
+++ b/AzureCommunicationUI/AzureCommunicationUI/CallComposite.swift
@@ -59,7 +59,7 @@ public class CallComposite {
                                                                     isRightToLeft: localizationProvider.isRightToLeft)
         setupManagers(store: dependencyContainer.resolve(),
                       logger: dependencyContainer.resolve(),
-                      callingSDKEventsHandling: dependencyContainer.resolve())
+                      callingSDKWrapper: dependencyContainer.resolve())
         present(toolkitHostingController)
     }
 
@@ -93,7 +93,7 @@ public class CallComposite {
 
     private func setupManagers(store: Store<AppState>,
                                logger: Logger,
-                               callingSDKEventsHandling: CallingSDKEventsHandling) {
+                               callingSDKWrapper: CallingSDKWrapper) {
         let errorManager = CompositeErrorManager(store: store,
                                                  callCompositeEventsHandler: callCompositeEventsHandler)
         self.errorManager = errorManager
@@ -111,7 +111,7 @@ public class CallComposite {
         let remoteParticipantsManager = CompositeRemoteParticipantsManager(
             store: store,
             callCompositeEventsHandler: callCompositeEventsHandler,
-            callingSDKEventsHandling: callingSDKEventsHandling)
+            callingSDKWrapper: callingSDKWrapper)
         self.remoteParticipantsManager = remoteParticipantsManager
     }
 
--- a/AzureCommunicationUI/AzureCommunicationUI/CallComposite.swift
+++ b/AzureCommunicationUI/AzureCommunicationUI/CallComposite.swift
@@ -8,6 +8,9 @@ import SwiftUI
 import FluentUI
 import AzureCommunicationCalling
 
+public typealias CompositeErrorHandler = (CommunicationUIErrorEvent) -> Void
+public typealias RemoteParticipantsJoinedHandler = ([CommunicationIdentifier]) -> Void
+
 /// The main class representing the entry point for the Call Composite.
 public class CallComposite {
     private var logger: Logger?
@@ -31,8 +34,8 @@ public class CallComposite {
     /// Assign closures to execute when events occur inside Call Composite.
     /// - Parameter didFailAction: The closure returning the error thrown from Call Composite.
     /// - Parameter participantsJoinedAction: The closure returning identifiers for joined remote participants.
-    public func setTarget(didFail didFailAction: ((CommunicationUIErrorEvent) -> Void)?,
-                          didRemoteParticipantsJoin participantsJoinedAction: (([CommunicationIdentifier]) -> Void)?) {
+    public func setTarget(didFail didFailAction: CompositeErrorHandler?,
+                          didRemoteParticipantsJoin participantsJoinedAction: RemoteParticipantsJoinedHandler?) {
         callCompositeEventsHandler.didFail = didFailAction
         callCompositeEventsHandler.didRemoteParticipantsJoin = participantsJoinedAction
     }
--- a/AzureCommunicationUI/AzureCommunicationUI/CallComposite.swift
+++ b/AzureCommunicationUI/AzureCommunicationUI/CallComposite.swift
@@ -13,7 +13,7 @@ public class CallComposite {
     private var logger: Logger?
     private let themeConfiguration: ThemeConfiguration?
     private let localizationConfiguration: LocalizationConfiguration?
-    private let callCompositeEventsHandler = CallCompositeEventsHandler()
+    private let callCompositeEventsHandler: CallCompositeEventsHandling
     private var errorManager: ErrorManager?
     private var lifeCycleManager: UIKitAppLifeCycleManager?
     private var permissionManager: AppPermissionsManager?
@@ -22,14 +22,18 @@ public class CallComposite {
     /// Create an instance of CallComposite with options.
     /// - Parameter options: The CallCompositeOptions used to configure the experience.
     public init(withOptions options: CallCompositeOptions? = nil) {
+        callCompositeEventsHandler = CallCompositeEventsHandler()
         themeConfiguration = options?.themeConfiguration
         localizationConfiguration = options?.localizationConfiguration
     }
 
-    /// Assign closure to execute when an error occurs inside Call Composite.
-    /// - Parameter action: The closure returning the error thrown from Call Composite.
-    public func setTarget(didFail action: ((CommunicationUIErrorEvent) -> Void)?) {
-        callCompositeEventsHandler.didFail = action
+    /// Assign closures to execute when events occur inside Call Composite.
+    /// - Parameter didFailAction: The closure returning the error thrown from Call Composite.
+    /// - Parameter participantsJoinedAction: The closure returning identifiers for joined remote participants.
+    public func setTarget(didFail didFailAction: ((CommunicationUIErrorEvent) -> Void)?,
+                          didRemoteParticipantsJoin participantsJoinedAction: (([CommunicationIdentifier]) -> Void)?) {
+        callCompositeEventsHandler.didFail = didFailAction
+        callCompositeEventsHandler.didRemoteParticipantsJoin = participantsJoinedAction
     }
 
     deinit {
@@ -42,7 +46,9 @@ public class CallComposite {
         logger = dependencyContainer.resolve() as Logger
         logger?.debug("launch composite experience")
 
-        dependencyContainer.registerDependencies(callConfiguration, localDataOptions: localOptions)
+        dependencyContainer.registerDependencies(callConfiguration,
+                                                 localDataOptions: localOptions,
+                                                 eventsHandler: callCompositeEventsHandler)
         let localizationProvider = dependencyContainer.resolve() as LocalizationProvider
         setupColorTheming()
         setupLocalization(with: localizationProvider)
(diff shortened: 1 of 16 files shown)


AzureCommunicationUI/AzureCommunicationUI/Presentation/Manager/CompositeRemoteParticipantsManager.swift
```diff
@@ -0,0 +1,56 @@
...
+import Combine
+import AzureCommunicationCommon
+
+protocol RemoteParticipantsManager {
+}
+
+class CompositeRemoteParticipantsManager: RemoteParticipantsManager {
+    private let store: Store<AppState>
+    private let eventsHandler: CallCompositeEventsHandling
+    private let callingSDKEventsHandling: CallingSDKEventsHandling
```
iaulakh, line 16: Andorid is using only redux state https://github.com/Azure/communication-ui-library-android/blob/feature/remote_user_persona_injection/azure-communication-ui/azure-communication-ui/src/main/java/com/azure/android/communication/ui/handlers/RemoteParticipantJoinedHandler.kt

vhuseinova-msft: We have a little bit different implementation of checks. I use sets instead of filtering arrays as it should take less time and to decrease amount of map operations, `participantsIds` propery is needed. And `participantsLastUpdateTimeStamp` is needed to check if there were updates in remoteParticipantsState (in the same was as in the Android implementation)

AmyL219: `callingSDKEventsHandling` is something coupled with CallingSDK. It should be converted into Redux State and then pass to Managers and viewModels. Could we add Redux state and map the value from `callingSDKEventsHandling` and use the Redux State here??

AzureCommunicationUI/AzureCommunicationUI/Presentation/Manager/CompositeErrorManager.swift
```diff
@@ -11,13 +11,13 @@ protocol ErrorManager {
 
 class CompositeErrorManager: ErrorManager {
     private let store: Store<AppState>
-    private weak var eventsHandler: CallCompositeEventsHandler?
+    private let eventsHandler: CallCompositeEventsHandling
```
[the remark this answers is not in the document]
vhuseinova-msft: The only instance that holds a link to CompositeErrorManager is CallComposite and both of them and CallCompositeEventsHandling are dealocated when a user navigates back to their app from the composite.

AzureCommunicationUI/AzureCommunicationUI/Presentation/Manager/CompositeRemoteParticipantsManager.swift
vhuseinova-msft: @AmyL219 I'll change functionality to use CallingSDKWrapper instead (it looks like we use this approach in VideoManager).

vhuseinova-msft: @AmyL219 updated!

AzureCommunicationUI/AzureCommunicationUI/Presentation/Manager/CompositeErrorManager.swift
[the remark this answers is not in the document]
JoshuaLai: I also did a quick check. And it looks like all the objects are being deallocated properly. As a follow up lets just make sure we run instruments to make sure nothing new pops up.

AzureCommunicationUI/AzureCommunicationUI/CallCompositeOptions/CallCompositeEventsHandler.swift
```diff
@@ -6,6 +6,12 @@ import Foundation
 import UIKit
 import AzureCommunicationCalling
 
-class CallCompositeEventsHandler {
+protocol CallCompositeEventsHandling: AnyObject {
+    var didFail: ((CommunicationUIErrorEvent) -> Void)? { get set }
+    var didRemoteParticipantsJoin: (([CommunicationIdentifier]) -> Void)? { get set }
```
JoshuaLai, line 11: Small nitpick: It might be beneficial for us to organize these closures a little better by introducing typealias for 
`(CommunicationUIErrorEvent) -> Void` and `([CommunicationIdentifier]) -> Void)`
I think I envision `public typealias CompositeErrorHandler = (CommunicationUIErrorEvent) -> Void` and `public typealias ParticipantJoinedHandler = ([CommunicationIdentifier]) -> Void` it will give us more context as to what the role of the closure is.

AzureCommunicationUI/AzureCommunicationUI/Presentation/Manager/RemoteParticipantsManager.swift
```diff
@@ -0,0 +1,56 @@
...
+    var cancellables = Set<AnyCancellable>()
+
+    init(store: Store<AppState>,
+         callCompositeEventsHandler: CallCompositeEventsHandling,
+         callingSDKWrapper: CallingSDKWrapper) {
+        self.store = store
+        self.eventsHandler = callCompositeEventsHandler
+        self.callingSDKWrapper = callingSDKWrapper
+        participantsIds = Set(store.state.remoteParticipantsState.participantInfoList.map { $0.userIdentifier })
+        participantsLastUpdateTimeStamp = store.state.remoteParticipantsState.lastUpdateTimeStamp
```
JoshuaLai, line 29: I don't know if this will be an issue or not, but I don't think this is correct here. Technically when we init the `RemoteParticipantsManager` there can never be any remote participants in the store. And if there is I think that means there is something wrong in the store. So I think you instead initialize with a new Set and the current date and time as a default.

AzureCommunicationUI/AzureCommunicationUI/Presentation/Manager/CompositeErrorManager.swift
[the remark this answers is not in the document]
vhuseinova-msft: I checked this with Memory graph and Leaks profiling today in the morning, and I didn't notice changes in memory usage with and without the weak modifier. 
@JoshuaLai do you want to check this once more?

AzureCommunicationUI/AzureCommunicationUI/Presentation/Manager/RemoteParticipantsManager.swift
vhuseinova-msft: @JoshuaLai I've used the state values here to have the only source of truth but there won't be any problems in logic to change this to the default values.

vhuseinova-msft: Updated as discussed to make this more readable

AzureCommunicationUI/AzureCommunicationUI/CallCompositeOptions/CallCompositeEventsHandler.swift
vhuseinova-msft: Updated!

JoshuaLai (approved):
LGTM

AzureCommunicationUI/AzureCommunicationUI/CallComposite.swift
```diff
@@ -8,28 +8,36 @@ import SwiftUI
 import FluentUI
 import AzureCommunicationCalling
 
+public typealias CompositeErrorHandler = (CommunicationUIErrorEvent) -> Void
+public typealias RemoteParticipantsJoinedHandler = ([CommunicationIdentifier]) -> Void
```
AmyL219, line 12: Have we discussed those new apis with Azure Review Board

AzureCommunicationUI/AzureCommunicationUI/Presentation/Manager/RemoteParticipantsManager.swift
```diff
@@ -0,0 +1,54 @@
...
+    var cancellables = Set<AnyCancellable>()
+
+    init(store: Store<AppState>,
+         callCompositeEventsHandler: CallCompositeEventsHandling,
+         callingSDKWrapper: CallingSDKWrapper) {
+        self.store = store
+        self.eventsHandler = callCompositeEventsHandler
+        self.callingSDKWrapper = callingSDKWrapper
+        store.$state
+            .receive(on: RunLoop.main)
```
AmyL219, line 29: Does the state process need to run in the main loop though?

AzureCommunicationUI/AzureCommunicationUI/Service/Calling/CallingSDKEventsHandler.swift
```diff
@@ -101,7 +101,9 @@ class CallingSDKEventsHandler: NSObject, CallingSDKEventsHandling {
                     $0.identifier.stringValue == infoModel.userIdentifier
                 })
             }
-        participantsInfoListSubject.send(remoteParticipantsInfoList)
+        if !remoteParticipants.isEmpty {
```
AmyL219, line 104: What if there is one remote participant, and this participant leave, and then the `remoteParticipants` is empty, but we need to update Redux state store that this participant has left by sending an empty list

vhuseinova-msft: In this situation `remoteParticipantsInfoList` will be empty as a result of filtering on line 99, but `remoteParticipants` array will still contain 1 participant and `participantsInfoListSubject.send(remoteParticipantsInfoList`) will be called as expected

AzureCommunicationUI/AzureCommunicationUI/Presentation/Manager/RemoteParticipantsManager.swift
vhuseinova-msft: I think it can be changed to DispatchQueue.main as in the receive function we will cal a completion for `didRemoteParticipantsJoin` event.

AzureCommunicationUI/AzureCommunicationUI/CallComposite.swift
vhuseinova-msft: I think we haven't yet(@JoshuaLai please confirm), but there are a couple of examples around the same in AzureCommunicationCommon sdk. Also, we can discuss this change on our next review

AzureCommunicationUI/AzureCommunicationUI/Service/Calling/CallingSDKEventsHandler.swift
AmyL219: Oh, got you. would that have a better performance to add `guard !remoteParticipants.isEmpty` in the function entry point of `removeRemoteParticipants`/`addRemoteParticipants`/`removeRemoteParticipantsInfoModel`/`addRemoteParticipantsInfoModel` ?

vhuseinova-msft: @AmyL219 Good catch! I think we can move these checks to the beggining of `removeRemoteParticipantsInfoModel` and `addRemoteParticipantsInfoModel` as there contain filtering in the beginning (in case we use the functions somewhere else) and to `call(_ call: Call, didUpdateRemoteParticipant args: ParticipantsUpdatedEventArgs)`, so we won't start the process at all. 
But I think we can skip `removeRemoteParticipants/addRemoteParticipants` as the `for` cycles won't be an issue

vhuseinova-msft: Updated!

AzureCommunicationUI/AzureCommunicationUI/CallComposite.swift
JoshuaLai: Yes that s a good point. We can bring it up in the formal Azure Board Review. I think it should be ok since its a typealias and should not be considered a breaking change

AmyL219 (approved):

PR #149, opened 2022-04-28
[Refactor] Rename protocol with -Protocol suffix

jimchou-dev:
## Purpose

* Rename protocol with `-Protocol` suffix
* Rename class/struct inherieting protocol to remove `ACS-` and `App-` prefix

```
[ ] Yes
[ x ] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## What to Check
* All unit test passing, existing functionality remains the same

### DIFF
--- a/AzureCommunicationUI/AzureCommunicationUI/CallComposite.swift
+++ b/AzureCommunicationUI/AzureCommunicationUI/CallComposite.swift
@@ -16,8 +16,8 @@ public class CallComposite {
     private let callCompositeEventsHandler = CallCompositeEventsHandler()
     private var errorManager: ErrorManager?
     private var lifeCycleManager: UIKitAppLifeCycleManager?
-    private var permissionManager: AppPermissionsManager?
-    private var audioSessionManager: AppAudioSessionManager?
+    private var permissionManager: PermissionsManager?
+    private var audioSessionManager: AudioSessionManager?
 
     /// Create an instance of CallComposite with options.
     /// - Parameter options: The CallCompositeOptions used to configure the experience.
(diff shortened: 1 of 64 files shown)


AzureCommunicationUI/AzureCommunicationUI/CallComposite.swift
```diff
@@ -16,8 +16,8 @@ public class CallComposite {
     private let callCompositeEventsHandler = CallCompositeEventsHandler()
     private var errorManager: ErrorManager?
     private var lifeCycleManager: UIKitAppLifeCycleManager?
-    private var permissionManager: AppPermissionsManager?
-    private var audioSessionManager: AppAudioSessionManager?
+    private var permissionManager: PermissionsManager?
+    private var audioSessionManager: AudioSessionManager?
```
vhuseinova-msft, line 20: Should we use protocols for properties instead?

jimchou-dev: I think we can keep this the same concrete type as before but with the renaming, since these are just private variables

kevinyulu (dismissed):

AzureCommunicationUI/AzureCommunicationUI/CallComposite.swift
AmyL219: We should use protocols in properties definition though, we used to not having those protocols, so directly use Manager class. But if we have protocol, we should use protocol for decoupling purpose

vhuseinova-msft (approved):

AmyL219 (approved):

PR #151, opened 2022-04-29
[Feature] Draggable PiP

palatter:
## Purpose
Allow the PiP to be draggable around the call screen container.

```
[ ] Yes
[ x] No
```

## Pull Request Type

```
[ ] Bugfix
[ X] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...


AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/CallingView.swift
```diff
@@ -49,50 +51,77 @@ struct CallingView: View {
...
+                        .accessibilitySortPriority(1)
+                        .accessibilityHidden(viewModel.isLobbyOverlayDisplayed)
+                }
+                .onAppear {
+                    if self.pipPosition == nil {
+                        self.pipPosition = getInitialPipPosition(containerBounds: geometry.frame(in: .local))
+                    }
+                }
+                .onChange(of: geometry.size) { _ in
+                    self.pipPosition = getInitialPipPosition(containerBounds: geometry.frame(in: .local))
```
jimchou-dev, line 72: When pip is dragged and rotated back to portrait the pip seems to reset to the location middle right of the screen, is this intentional or is it suppose to go back to bottom right of the screen?
image

palatter: Looks like the onSizeChanged isn't triggered when rotating that specific way on an iphone. So I have added back in my on Rotation solution. With both of these, it should cover all cases.

jimchou-dev (approved):
Tested, LGTM

JoshuaLai (approved):
Tested with a bunch of different scenarios and different outage scenarios as well and it looks good. Checked the memory as well looks like no new leaks

PR #152, opened 2022-05-02
[Refactor] Move determineErrorType to ACSCallEndReasonExtension

AmyL219:
## Purpose

* Add extension of CallEndReason, to convert it into `CallCompositeErrorCode` String
* `Question`, I am thinking, why not make `CallCompositeErrorCode` as a protocol/subclass of String, therefore, we could limit the CommunicationUIErrorEvent only having the string code defined inside our library. Otherwise, it would be any string. 
* For example, for the function, `toCompositeErrorCodeString()->String`, it could return any string, and I couldn't make it return `CallCompositeErrorCode`, because `CallCompositeErrorCode` is a struct,  which may not friendly for Community contributor. cc @JoshuaLai
 
```
public struct CommunicationUIErrorEvent {
    public let code: String -> CallCompositeErrorCode
    public var error: Error?
}
``` 

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

### DIFF
--- a/AzureCommunicationUI/AzureCommunicationUI/Service/Calling/CallingSDKEventsHandler.swift
+++ b/AzureCommunicationUI/AzureCommunicationUI/Service/Calling/CallingSDKEventsHandler.swift
@@ -150,8 +150,8 @@ extension CallingSDKEventsHandler: CallDelegate,
 
     func call(_ call: Call, didChangeState args: PropertyChangedEventArgs) {
         let currentStatus = call.state.toCallingStatus()
-        let errorCode = determineErrorType(previousStatus: self.previousCallingStatus,
-                                           callEndReason: call.callEndReason)
+        let isCallConnected = previousCallingStatus == .connected
+        let errorCode = call.callEndReason.toCompositeErrorCodeString(isCallConnected)
 
         let callInfoModel = CallInfoModel(status: currentStatus, errorCode: errorCode)
         callInfoSubject.send(callInfoModel)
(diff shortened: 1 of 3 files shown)


palatter (dismissed):
Looks good and tested!

AzureCommunicationUI/AzureCommunicationUI/Service/Calling/CallingSDKEventsHandler.swift
```diff
@@ -150,8 +150,8 @@ extension CallingSDKEventsHandler: CallDelegate,
 
     func call(_ call: Call, didChangeState args: PropertyChangedEventArgs) {
         let currentStatus = call.state.toCallingStatus()
-        let errorCode = determineErrorType(previousStatus: self.previousCallingStatus,
-                                           callEndReason: call.callEndReason)
+        let isCallConnected = previousCallingStatus == .connected
```
kevinyulu, line 153: PR lktm, one thing about this name though, when at this line the current call state may already be disconnected. would we use names like `previousCallConnected` to reflect the previous calling status?

kevinyulu (dismissed):

vhuseinova-msft (dismissed):

AzureCommunicationUI/AzureCommunicationUI/Service/Calling/CallingSDKEventsHandler.swift
AmyL219: Yeah, rename to `wasCallConnected`

kevinyulu (approved):
LGTM

palatter:
Looks good!

palatter (approved):

PR #153, opened 2022-05-02
[Refactor] Refactored call eviction error handling not to be propagated to contoso

kevinyulu:
## Purpose
Refactor call eviction error handling not to be propagated to contoso
Make a protocol for CallCompositeErrorCode and have two new type (one public and one internal) conform to the protocol

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code
* Launch app, join a group call / teams call
* On another desktop, join a group call / teams call, and remove the remote participant from the call
* Go back to app, see if the user is removed from the call (setup screen shown with error message)

## What to Check
* Check if the user is removed from the call in the app and app navigates user back to setup screen
* Check if the error message shows correctly in the setup screen, displaying a banner with call eviction message.

### DIFF
--- a/AzureCommunicationUI/AzureCommunicationUI/CallCompositeOptions/CommunicationUIErrorEvent.swift
+++ b/AzureCommunicationUI/AzureCommunicationUI/CallCompositeOptions/CommunicationUIErrorEvent.swift
@@ -5,20 +5,37 @@
 
 import UIKit
 
-/// Call Composite runtime error types.
-public struct CallCompositeErrorCode {
+protocol BaseCallCompositeErrorCode {
+    static var callJoin: String { get }
+    static var callEnd: String { get }
+    static var tokenExpired: String { get }
+}
 
-    /// Error when local user fails to join a call.
-    public static let callJoin: String = "callJoin"
+/// Call Composite runtime error types.
+public struct CallCompositeErrorCode: BaseCallCompositeErrorCode {
+
+    /// Error when local user fails to join a call. 
+    public private(set) static var callJoin: String = "callJoin"
 
     /// Error when a call disconnects unexpectedly or fails on ending.
-    public static let callEnd: String = "callEnd"
+    public private(set) static var callEnd: String = "callEnd"
 
     /// Error when the input token is expired.
-    public static let tokenExpired: String = "tokenExpired"
+    public private(set) static var tokenExpired: String = "tokenExpired"
+}
+
+struct InternalCallCompositeErrorCode: BaseCallCompositeErrorCode { // TODO: check name InternalCallCompositeErrorCode
+    /// Error when local user fails to join a call.
+    static var callJoin: String = "callJoin"
+
+    /// Error when a call disconnects unexpectedly or fails on ending.
+    static var callEnd: String = "callEnd"
+
+    /// Error when the input token is expired.
+    static var tokenExpired: String = "tokenExpired"
 
     /// Error when a participant is evicted from the call by another participant
-    public static let callEvicted: String = "callEvicted"
+    static var callEvicted: String = "callEvicted"
 }
 
 /// The error thrown after Call Composite launching.
--- a/AzureCommunicationUI/AzureCommunicationUI/CallCompositeOptions/CommunicationUIErrorEvent.swift
+++ b/AzureCommunicationUI/AzureCommunicationUI/CallCompositeOptions/CommunicationUIErrorEvent.swift
@@ -24,7 +24,7 @@ public struct CallCompositeErrorCode: BaseCallCompositeErrorCode {
     public private(set) static var tokenExpired: String = "tokenExpired"
 }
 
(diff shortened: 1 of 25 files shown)


jimchou-dev (dismissed):

AzureCommunicationUI/AzureCommunicationUI/CallCompositeOptions/CommunicationUIErrorEvent.swift
```diff
@@ -5,20 +5,37 @@
...
+    public private(set) static var callJoin: String = "callJoin"
+
+    /// Error when a call disconnects unexpectedly or fails on ending.
+    public private(set) static var callEnd: String = "callEnd"
+
+    /// Error when the input token is expired.
+    public private(set) static var tokenExpired: String = "tokenExpired"
+}
 
+struct InternalCallCompositeErrorCode: BaseCallCompositeErrorCode {
```
AmyL219, line 27: Why not make `CallCompositeErrorCode: String` to be a String protocol. We can have 
```
public static let callJoin: CallCompositeErrorCode = "callJoin"

    public  static var callEnd: CallCompositeErrorCode = "callEnd"

    public static let tokenExpired: CallCompositeErrorCode = "tokenExpired"

    static let callEvicted: CallCompositeErrorCode = "callEvicted"
```

Right now `InternalCallCompositeErrorCode` has a lot of duplicated error code with `CallCompositeErrorCode`. And it's kind of confused for reading

AmyL219:
* `Question`, I am thinking, why not make `CallCompositeErrorCode` as a protocol/subclass of String, therefore, we could limit the CommunicationUIErrorEvent only having the string code defined inside our library. Otherwise, it would be any string. 
* For example, for the function, `toCompositeErrorCodeString()->String`, it could return any string, and I couldn't make it return `CallCompositeErrorCode`, because `CallCompositeErrorCode` is a struct,  which may not friendly for Community contributor. cc @JoshuaLai
 
```
public struct CommunicationUIErrorEvent {
    public let code: String -> CallCompositeErrorCode
    public var error: Error?
}
```

JoshuaLai:
@AmyL219 So the reason it is `struct CallCompositeErrorCode` with static variables is so developers will have ease of discoverability in the IDE. But I really like the suggestion of having a base conform or subclass String. I think that makes a lot of sense. Let me work with Kevin on this and see how the implementation will look

AzureCommunicationUI/AzureCommunicationUI/CallCompositeOptions/CommunicationUIErrorEvent.swift
```diff
@@ -5,20 +5,37 @@
 
 import UIKit
 
+protocol BaseCallCompositeErrorCode {
```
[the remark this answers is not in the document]
JoshuaLai: No this does not need to be public, the `CallCompositeErrorCode` is already public and that should be all developers need

AzureCommunicationUI/AzureCommunicationUI/CallCompositeOptions/CommunicationUIErrorEvent.swift
JoshuaLai: I really like this suggestion and actually makes a lot more sense and simpler too.

AzureCommunicationUI/AzureCommunicationUI/CallCompositeOptions/CommunicationUIErrorEvent.swift
```diff
@@ -5,20 +5,37 @@
...
+    static var callJoin: String { get }
+    static var callEnd: String { get }
+    static var tokenExpired: String { get }
+}
+
 /// Call Composite runtime error types.
-public struct CallCompositeErrorCode {
+public struct CallCompositeErrorCode: BaseCallCompositeErrorCode {
```
vhuseinova-msft, line 15: If we use only static values, we can use classes instead of structs and inheritance

AzureCommunicationUI/AzureCommunicationUI/CallCompositeOptions/CommunicationUIErrorEvent.swift
```diff
@@ -5,20 +5,37 @@
...
+
+    /// Error when local user fails to join a call. 
+    public private(set) static var callJoin: String = "callJoin"
```
vhuseinova-msft, line 18: Can public(get) be used here?

AzureCommunicationUI/AzureCommunicationUI/CallCompositeOptions/CommunicationUIErrorEvent.swift
kevinyulu: will add this in the next commit

AmyL219 (approved):

vhuseinova-msft (approved):

PR #154, opened 2022-05-03
[Feature] Add functionality for setting remote participant persona data

vhuseinova-msft:
## Purpose

* Renamed CommunicationUIPersonaData to PersonaData
* Added functionality to set and store persona data for remote participants

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## Other Information

The functionality for remote participants persona data check will be added in the next task for grid update.
The errors handling will be added in the next PRs.


AzureCommunicationUI/AzureCommunicationUI/Presentation/Manager/AvatarViewManager.swift
```diff
@@ -5,27 +5,53 @@
...
+            .sink { [weak self] state in
+                self?.receive(state: state)
+            }.store(in: &cancellables)
     }
 
-    func getLocalPersonaData() -> CommunicationUIPersonaData? {
+    private func receive(state: AppState) {
+        guard state.callingState.status == .disconnected ||
+                state.errorState.errorCategory == .callState
+        else { return }
```
AmyL219, line 35: Should we move `else` to the trail of the statement, to keep the whole project coding style consistent? 
We could create another PR/ticket for this one
Coding style reference: https://docs.swift.org/swift-book/LanguageGuide/ControlFlow.html
```
guard state.callingState.status == .disconnected ||
                state.errorState.errorCategory == .callState else { 
return 
}
```

```
        guard let state = state as? AppState else {
            return
        }

```

AzureCommunicationUI/AzureCommunicationUI/CallComposite.swift
```diff
@@ -94,35 +93,52 @@ public class CallComposite {
...
-                               callingSDKWrapper: CallingSDKWrapper) {
-        let errorManager = CompositeErrorManager(store: store,
+    @discardableResult
+    /// Set persona data for the remote participant
+    /// - Parameters:
+    ///   - identifier: The communication identifier for the remote participant
+    ///   - personaData: PersonaData used to set the user participants information for the call.
+    ///   This is data is not sent up to ACS.
+    /// - Returns: The `Result` enum value with either a `Void` or an `Error`
+    public func setRemoteParticipantPersonaData(for identifier: CommunicationIdentifier,
```
AmyL219, line 103: Any reference of similiar ACS Api using `Result<Void, Error>`? would that be more Consistent to CallingSDK by defining the error in a completion handler?

AzureCommunicationUI/AzureCommunicationUI/CallComposite.swift
```diff
@@ -94,35 +93,52 @@ public class CallComposite {
...
+                                                personaData: PersonaData) -> Result<Void, Error> {
+        guard let avatarManager = avatarViewManager
+        else { return .failure(CompositeError.callCompositeNotLaunched) }
+
+        return avatarManager.setRemoteParticipantPersonaData(for: identifier,
+                                                             personaData: personaData)
+    }
+
+    private func setupManagers(dependencyContainer: DependencyContainer) {
```
AmyL219, line 112: Yeah, we should be able to move the Managers initialization to Dependency Container. As long as make `CallComposite` keep reference to those managers. 
[Task 2843967](https://skype.visualstudio.com/SPOOL/_workitems/edit/2843967): [iOS][P4] Move managers setup from CallComposite to DependencyContainer

AzureCommunicationUI/AzureCommunicationUI/CallComposite.swift
AmyL219: I feel `Result<Void, Error>` is not popular api in public ACS interface. Correct me if I am wrong

AzureCommunicationUI/AzureCommunicationUI/CallComposite.swift
vhuseinova-msft: Yep, I've created that task to move the managers and not wait for the customdatamodel branch to be merged. For now, avatarViewManager is created here for consistency

AzureCommunicationUI/AzureCommunicationUI/CallComposite.swift
JoshuaLai: No I don't think you are wrong this is something brand new that I think we should discuss with the Azure Review Board. I do believe that ChatSDK may be doing something similar, but they actually have a body response in the case of success

vhuseinova-msft: The Result was used because it's a native enum and, in this case, there is no async code, so the func return result will be okay. I think ChatSDK uses [HTTPResultHandler](https://azure.github.io/azure-sdk-for-ios/AzureCore/Typealiases.html#/s:9AzureCore13ResultHandlera)

AzureCommunicationUI/AzureCommunicationUI/Presentation/Manager/AvatarViewManager.swift
vhuseinova-msft: Updated!

AmyL219 (approved):

JoshuaLai (approved):

PR #157, opened 2022-05-03
[Bugfix] Accessibility VoiceOver for screen share and join call

jimchou-dev:
## Purpose

* Added accessibility VoiceOver for screen share start and end
* Added accessibility VoiceOver for when joined a call (both group call and teams call)

```
[ ] Yes
[ x ] No
```

## Pull Request Type

```
[ x ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## What to Check
* Test voice over will announce when screen share starts, and ends
* Test voice over will announce when joined a call

### DIFF
--- a/AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/CallingViewModel.swift
+++ b/AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/CallingViewModel.swift
@@ -13,7 +13,7 @@ class CallingViewModel: ObservableObject {
     @Published var isVideoGridViewAccessibilityAvailable: Bool = false
     let isRightToLeft: Bool
     @Published var appState: AppStatus = .foreground
-    private var isCallConnected: Bool = false
+    private var callHasConnected: Bool = false
 
     private let compositeViewModelFactory: CompositeViewModelFactoryProtocol
     private let logger: Logger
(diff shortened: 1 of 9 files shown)


kevinyulu (dismissed):
tested, LGTM

AzureCommunicationUI/AzureCommunicationUI/Presentation/SwiftUI/Calling/CallingViewModel.swift
```diff
@@ -13,6 +13,7 @@ class CallingViewModel: ObservableObject {
     @Published var isVideoGridViewAccessibilityAvailable: Bool = false
     let isRightToLeft: Bool
     @Published var appState: AppStatus = .foreground
+    private var isCallConnected: Bool = false
```
kevinyulu, line 16: very minor: it looks like this `isCallConnected` is to make sure voice over only gets announced once. if the call has already not been connected and the call status is connected, then the voice over should be announced. This variable can be renamed as `callHasConnected`

AzureCommunicationUI/AzureCommunicationUITests/Presentation/Calling/CallingViewModelTests.swift
```diff
@@ -149,7 +149,7 @@ class CallingViewModelTests: XCTestCase {
     func test_callingViewModel_receive_when_statusUpdated_then_bannerViewModelUpdated() {
         let expectation = XCTestExpectation(description: "BannerViewModel is updated")
         let appState = AppState(callingState: CallingState(status: .connected))
-        let sut = makeSUT(updateBannerViewModel: { callingState in
+        let sut = makeSUT(state: appState, updateBannerViewModel: { callingState in
```
vhuseinova-msft, line 152: Does this broke the test logic, as the same appState is used for the initial state and later for updating the state?

jimchou-dev: Yes agree, created refactoring ticket ADO item 2852111

vhuseinova-msft: Agreed with @jimchou-dev that this will be addressed later in a separate task

vhuseinova-msft (dismissed):
looks good and tested

AmyL219 (dismissed):

vhuseinova-msft (approved):

AmyL219 (approved):

PR #167, opened 2022-05-09
[Refactor] Event Handler refactor as per Azure Board suggestions

JoshuaLai:
## Purpose
As per conversations with suggestions brought forward by Azure board 
New unilateral properties suggestion for how to inject events. 

```
[x] Yes
[ ] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

### DIFF
--- a/AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/CallComposite.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/CallComposite.swift
@@ -10,10 +10,10 @@ import AzureCommunicationCalling
 
 /// The main class representing the entry point for the Call Composite.
 public class CallComposite {
+    public var callCompositeEventsHandler = CallCompositeEvents()
     private var logger: Logger?
     private let themeConfiguration: ThemeConfiguration?
     private let localizationConfiguration: LocalizationConfiguration?
-    private let callCompositeEventsHandler = CallCompositeEventsHandler()
     private var errorManager: ErrorManagerProtocol?
     private var lifeCycleManager: LifeCycleManagerProtocol?
     private var permissionManager: PermissionsManagerProtocol?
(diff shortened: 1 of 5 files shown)


AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/CallComposite.swift
```diff
@@ -10,10 +10,10 @@ import AzureCommunicationCalling
 
 /// The main class representing the entry point for the Call Composite.
 public class CallComposite {
+    public var callCompositeEventsHandler = CallCompositeEvents()
```
JoshuaLai, line 13: Should this be `lazy`?

AmyL219: would it make more sense to call it `eventsHandler` rather than `callCompositeEventsHandler`?

AmyL219 (commented):
I don't have too much concern to move to the univerProperties event handler. Please be free to discuss this without me.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/CallComposite.swift
waynemo: I think the recommendation was just to use `events` since it would be events on the `CallComposite`.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/CallCompositeOptions/CallCompositeEventsHandler.swift
```diff
@@ -6,6 +6,7 @@ import Foundation
 import UIKit
 import AzureCommunicationCalling
 
-class CallCompositeEventsHandler {
-    var didFail: ((CommunicationUIErrorEvent) -> Void)?
+public typealias CompositeErrorHandler = (CommunicationUIErrorEvent) -> Void
+public struct CallCompositeEvents {
```
vhuseinova-msft, line 10: I'm wondering if there can be a problem when we already sent copies of this struct to our objects and  a user changes `CallCompositeEvents's` properties later. I think we will end up with outdated version of `CallCompositeEvents ` objects in our objects

PR #170, opened 2022-05-09
[Feature] Updated the size and UX of the pip view on iPad layout in calling screen

kevinyulu:
## Purpose
Updated the size and UX of the pip view on iPad layout in calling screen

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code
* Join a call, turn on video, using iPad and iPhone, 
* test the pip view on different type of device, with both orientations and two UI modes

## What to Check
* Make sure the pip view has correct UX in calling view in different type of device, with both orientations and two UI modes


kevinyulu:
The size of the PIP view is to be confirm by UX

PR #171, opened 2022-05-09
[Feature] Display remote participants persona data in the grid view

vhuseinova-msft:
## Purpose

* Introduce functionality to set a remote participant persona data
* Display avatars and updated names for remote participants (all data is saved and set locally)
* Add an option for turning on remote persona data injection

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to test
Please turn on "Inject avatars" option in Settings

## What to Check
* images and name can be set for remote participants
* preset image and name are shown for a participant who has identifier's last symbol in 0...5 range

## Other Information

You may need to add a couple of users to the call to see avatars and updated names as some participants can have identifier's last symbol equal >5 or a letter.


AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Presentation/SwiftUI/Calling/Grid/Cell/ParticipantGridCellView.swift
```diff
@@ -10,6 +10,9 @@ import Combine
 struct ParticipantGridCellView: View {
     @ObservedObject var viewModel: ParticipantGridCellViewModel
     let rendererViewManager: RendererViewManager?
+    let avatarViewManager: AvatarViewManager
+    @State var renderDisplayName: String?
+    @State var avatarImage: UIImage?
```
jimchou-dev, line 15: Would it make sense to have an optional `PersonaData` instead? And some of the logic like `getName(personaData: PersonaData?)` can be extracted to the ParticipantGridCellViewModel

vhuseinova-msft: In this case PersonaData properties should be changed from constants to variables (it is a public struct) and also a custom binding should be created for images, as in case when personadata = nil, the binding to the image will be broken. So it looks like it's better to use separate states for images and names

jimchou-dev (approved):
Tested, LGTM

AzureCommunicationUI/AzureCommunicationUIDemoApp/AzureCommunicationUIDemoApp/RemoteParticipantAvatarHelper.swift
```diff
@@ -0,0 +1,43 @@
...
+            if let lastSymbol = id?.last,
+               let index = Int(String(lastSymbol)),
+               index < avatars.count {
+                let selectedAvatar = avatars[index]
+                let image = UIImage(named: selectedAvatar)
+                let nameIdValue = id != nil ? " \(id?.suffix(4) ?? "")" : ""
+                let personaData = PersonaData(avatar: image,
+                                              renderDisplayName: selectedAvatar + nameIdValue)
+                callComposite.setRemoteParticipantPersonaData(for: identifier,
+                                                              personaData: personaData)
```
JoshuaLai, line 39: In this case, I think regardless of whether or not the last character is an `Int` we still update the `renderDisplayName` for the remote participant. This will help with debugging so we know WHY they don't have an avatar displayed

vhuseinova-msft: Updated!

JoshuaLai (approved):

PR #174, opened 2022-05-11
[Feature] Change how we set event handler

JoshuaLai:
## Purpose
Second approach to having developers set event handlers. 
Have a new method per event handler so developers can only set but not get a reference to them .

What they will see in the IDE:
Screen Shot 2022-05-11 at 12 08 24 PM

If developers choose to use trailing closure syntax:
Screen Shot 2022-05-11 at 12 08 49 PM

```
[x] Yes
[ ] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

### DIFF
--- a/AzureCommunicationUI/AzureCommunicationUIDemoApp/AzureCommunicationUIDemoApp/Views/SwiftUIDemoView.swift
+++ b/AzureCommunicationUI/AzureCommunicationUIDemoApp/AzureCommunicationUIDemoApp/Views/SwiftUIDemoView.swift
@@ -151,6 +151,7 @@ extension SwiftUIDemoView {
             : Theming(envConfigSubject: envConfigSubject),
             localization: localizationConfig)
         let callComposite = CallComposite(withOptions: callCompositeOptions)
+
         let didRemoteParticipantsJoin: ([CommunicationIdentifier]) -> Void = { [weak callComposite] identifiers in
             guard let composite = callComposite else {
                 return
@@ -158,7 +159,9 @@ extension SwiftUIDemoView {
 
             self.didRemoteParticipantsJoin(to: composite, identifiers: identifiers)
         }
-        callComposite.setTarget(didFail: didFail, didRemoteParticipantsJoin: didRemoteParticipantsJoin)
+        callComposite.setEventHandler(didFail: didFail)
+        callComposite.setEventHandler(didRemoteParticipantJoin: didRemoteParticipantsJoin)
+
         let renderDisplayName = envConfigSubject.renderedDisplayName.isEmpty ?
                                 nil:envConfigSubject.renderedDisplayName
         let persona = PersonaData(avatar: UIImage(named: envConfigSubject.avatarImageName),
--- a/AzureCommunicationUI/AzureCommunicationUIDemoApp/AzureCommunicationUIDemoApp/Views/SwiftUIDemoView.swift
+++ b/AzureCommunicationUI/AzureCommunicationUIDemoApp/AzureCommunicationUIDemoApp/Views/SwiftUIDemoView.swift
@@ -159,8 +159,8 @@ extension SwiftUIDemoView {
 
             self.didRemoteParticipantsJoin(to: composite, identifiers: identifiers)
         }
-        callComposite.setEventHandler(didFail: didFail)
-        callComposite.setEventHandler(didRemoteParticipantJoin: didRemoteParticipantsJoin)
+        callComposite.setDidFailHandler(didFail: didFail)
+        callComposite.setRemoteParticipantJoinHandler(didRemoteParticipantJoin: didRemoteParticipantsJoin)
 
         let renderDisplayName = envConfigSubject.renderedDisplayName.isEmpty ?
                                 nil:envConfigSubject.renderedDisplayName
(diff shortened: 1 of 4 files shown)


AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/CallComposite.swift
```diff
@@ -32,12 +32,15 @@ public class CallComposite {
...
-    /// - Parameter participantsJoinedAction: The closure returning identifiers for joined remote participants.
-    public func setTarget(didFail didFailAction: CompositeErrorHandler?,
-                          didRemoteParticipantsJoin participantsJoinedAction: RemoteParticipantsJoinedHandler?) {
+    public func setEventHandler(didFail didFailAction: CompositeErrorHandler?) {
         callCompositeEventsHandler.didFail = didFailAction
+    }
+
+    /// Assign closures to execute when participant has joined a call  inside Call Composite.
+    /// - Parameter participantsJoinedAction: The closure returning identifiers for joined remote participants.
+    public func setEventHandler(didRemoteParticipantJoin participantsJoinedAction: RemoteParticipantsJoinedHandler?) {
```
waynemo, line 43: Let's check out what happens when we have multiple events with the exact same closure signature.

waynemo (dismissed):

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/CallComposite.swift
vhuseinova-msft: But in this case we will still have different function signature, won't we?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/CallComposite.swift
```diff
@@ -32,12 +32,15 @@ public class CallComposite {
         localizationConfiguration = options?.localizationConfiguration
     }
 
-    /// Assign closures to execute when events occur inside Call Composite.
+    /// Assign closures to execute when error event  occurs inside Call Composite.
     /// - Parameter didFailAction: The closure returning the error thrown from Call Composite.
-    /// - Parameter participantsJoinedAction: The closure returning identifiers for joined remote participants.
-    public func setTarget(didFail didFailAction: CompositeErrorHandler?,
-                          didRemoteParticipantsJoin participantsJoinedAction: RemoteParticipantsJoinedHandler?) {
+    public func setEventHandler(didFail didFailAction: CompositeErrorHandler?) {
```
vhuseinova-msft, line 37: Maybe we can rename the parameter for setEventHandler functions to be in `on<event>` format, e.g. `onFail`?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/CallComposite.swift
JoshuaLai: In the case of trailing closure syntax developers will see the `Ambiguous use of 'setEventHandler'` error in the IDE. Which makes sense now that I think about it. 
image

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/CallComposite.swift
JoshuaLai: The original intent was follow swift idioms a little bit more.  
But I think if this concept raises ambiguity in the trailing syntax closure we may fallback to this idea

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/CallComposite.swift
JoshuaLai: Updating the method signature to be more specific will resolve this ambiguity.

jimchou-dev (approved):

AzureCommunicationUI/AzureCommunicationUIDemoApp/AzureCommunicationUIDemoApp/Views/SwiftUIDemoView.swift
```diff
@@ -151,14 +151,17 @@ extension SwiftUIDemoView {
...
+
         let didRemoteParticipantsJoin: ([CommunicationIdentifier]) -> Void = { [weak callComposite] identifiers in
             guard let composite = callComposite else {
                 return
             }
 
             self.didRemoteParticipantsJoin(to: composite, identifiers: identifiers)
         }
-        callComposite.setTarget(didFail: didFail, didRemoteParticipantsJoin: didRemoteParticipantsJoin)
+        callComposite.setDidFailHandler(didFail: didFail)
```
vhuseinova-msft, line 162: I think we can skip parameter name in this case as it's a part of function name

vhuseinova-msft: Minor: I think we can even skip `with` parameter name as there is no new info added and we just set a handler

vhuseinova-msft: Minor: I think we can even skip `with` parameter name as there is no new info added and we just set a handler

vhuseinova-msft (approved):

PR #218, opened 2022-05-27
[Feature] Add onHold state for remote participant with UI/UX update

AmyL219:
## Purpose

* Add onHold state for remote participant with UI/UX update
Simulator Screen Shot - iPhone 11 - 2022-05-27 at 10 12 03

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

### DIFF
--- a/AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Presentation/SwiftUI/Calling/Grid/Cell/ParticipantGridCellViewModel.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Presentation/SwiftUI/Calling/Grid/Cell/ParticipantGridCellViewModel.swift
@@ -22,6 +22,7 @@ class ParticipantGridCellViewModel: ObservableObject, Identifiable {
     @Published var displayName: String?
     @Published var isSpeaking: Bool
     @Published var isMuted: Bool
+    @Published var isHold: Bool
     @Published var participantIdentifier: String
     private var isScreenSharing: Bool = false
     private var participantName: String
@@ -35,6 +36,7 @@ class ParticipantGridCellViewModel: ObservableObject, Identifiable {
         self.participantName = participantModel.displayName
         self.displayName = participantModel.displayName
         self.isSpeaking = participantModel.isSpeaking
+        self.isHold = participantModel.status == .hold
         self.participantIdentifier = participantModel.userIdentifier
         self.isMuted = participantModel.isMuted
         self.videoViewModel = getDisplayingVideoStreamModel(participantModel)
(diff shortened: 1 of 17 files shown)


AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Service/Calling/Extension/ACSParticipantStateExtension.swift
```diff
@@ -0,0 +1,31 @@
...
+        case .hold:
+            return .hold
+        case .inLobby:
+            return .inLobby
+        case .disconnected:
+            return .disconnected
+        case .ringing:
+            return .ringing
+        default:
+            return .idle
```
kevinyulu, line 28: out of curiosity, what the rule from SDK that would determine if a participant is `.idle` vs `.connected`?

kevinyulu: also would we default the state to .connected?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Presentation/SwiftUI/Calling/Grid/Cell/ParticipantGridCellViewModel.swift
```diff
@@ -76,6 +78,12 @@ class ParticipantGridCellViewModel: ObservableObject, Identifiable {
         if self.isMuted != participantModel.isMuted {
             self.isMuted = participantModel.isMuted
         }
+
+        let isOnHold = participantModel.status == .hold
+
+        if self.isHold != isOnHold {
```
kevinyulu, line 84: Tested with having a teams call with 2 participants: (Kevin Guo (host) and User 2 (guest)) with the following two scenarios:
Scenario 1 **Kevin Guo is on hold**: (tested OK) host will show on hold UX in participant grid and participant list drawer. 

Scenario 2 **User 2 is on hold**: when this case happens, the app shows an on hold status on the host _Kevin Guo_ in the participant drawer. Should the guest User 2 be labeled as on hold instead? Screenshot attached
IMG_0090

kevinyulu (commented):
tested the app in iPhone XR using Teams and left a couple of comments on some test cases.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Presentation/SwiftUI/Calling/Grid/Cell/ParticipantGridCellViewModel.swift
AmyL219: Yeah, the CallingSDK updates all remote participants with `.hold` state. Let me dismiss the participant list when the local user is update to `onHold` state which would be consistent with Android side

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Service/Calling/Extension/ACSParticipantStateExtension.swift
AmyL219: I think `.idle` is more equal to the localUserState `.none`. I think we would only update the remote participant state to .connected when he has the .connected status from CallingSDK

PR #222, opened 2022-05-31
[Bug] Fix the composite thrown error code when user failed to resume 

AmyL219:
## Purpose

* Update the function `wasCallConnected`, so the error code would be update to `EndCall` from `JoinCall`

```
[ ] Yes
[x] No
```


palatter (approved):

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Service/Calling/CallingSDKEventsHandler.swift
```diff
@@ -140,6 +140,16 @@ class CallingSDKEventsHandler: NSObject, CallingSDKEventsHandling {
...
     }
+
+    private func wasCallConnected() -> Bool {
+        guard previousCallingStatus != .connected,
+              previousCallingStatus != .localHold,
+              previousCallingStatus != .remoteHold else {
+            return true
+        }
+
+        return false
```
jimchou-dev, line 151:
```suggestion
        return previousCallingStatus == .connected || previousCallingStatus == .localHold || previousCallingStatus == .remoteHold
```

AmyL219: I feel for the readability and extensibility, the original one is better.

AmyL219: I see. From code clarity and performance side, `guard` is more recommended than using `if` in the swift. See this most popular Swift best practice: https://github.com/raywenderlich/swift-style-guide#golden-path

Typically `guard` would make the code cleaner and with a better performance in the running time. Because `guard` would directly return for `unhappy path`. The `unhappy condition` wouldn't go through the whole function. I feel like we should use `guard` in the whole project for consistent purpose.

jimchou-dev: From the swift guide those `guard` example make sense because it's dealing with errors and exceptions which is control flow related.

The evaluation of the statement results in a Bool, and it would be redundant to explicitly return `true` and `false` when function serves no other purpose then returning the evaluated result.

If it's for consistency purpose in the project, for function that returns Bool with simple logic currently just directly return the conditional evaluation, eg. `isCameraDisabled()`, but I agree there are some inconsistencies, some have the word `return` while other only has the evaluation for simplicity.
eg.
```
SetupControlBarViewModel
 func isCameraDisabled() -> Bool {
        return isJoinRequested || cameraPermission == .denied
    }
```
versus
```
ControlBarViewModel
 func isCameraDisabled() -> Bool {
        cameraPermission == .denied || cameraState.operation == .pending
    }
```

AmyL219: The concern of this approach  you mentioned is , what if in future, there are a lot of conditions need to add. I feel like this is not friendly for reading. 
```
  func isCameraDisabled() -> Bool {
        return isJoinRequested || isCallHold || isCallConnected ..... || cameraPermission == .denied
    }
```
Let me discuss this in tomorrow's parking lot. To finalized the consistent approach we want to take.
Possibly I would merge this PR first to unblock the callOnHold feature merging. And do the refactoring work later.

jimchou-dev (approved):

PR #236, opened 2022-06-03
[Bugfix] fixed one retain cycle of call composite

kevinyulu:
## Purpose

* Fixed one retain cycle of call composite

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

1. Start a group call with lcoal video on
2. End the call 
3. re-launch call 
4. You would find the local video view is empty

## What to Check
* check if the local video view will show camera scene in setup view

### DIFF
--- a/AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Presentation/SwiftUI/Calling/CallingViewModel.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Presentation/SwiftUI/Calling/CallingViewModel.swift
@@ -123,7 +123,10 @@ class CallingViewModel: ObservableObject {
         }
 
         if callHasConnected != newIsCallConnected && newIsCallConnected {
-            DispatchQueue.main.asyncAfter(deadline: .now() + 2.5) {
+            DispatchQueue.main.asyncAfter(deadline: .now() + 2.5) { [weak self] in
+                guard let self = self else {
+                    return
+                }
                 self.accessibilityProvider.postQueuedAnnouncement(
                     self.localizationProvider.getLocalizedString(.joinedCallAccessibilityLabel))
             }
--- a/AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Presentation/SwiftUI/Container/ContainerUIHostingController.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Presentation/SwiftUI/Container/ContainerUIHostingController.swift
@@ -62,10 +62,10 @@ class ContainerUIHostingController: UIHostingController<ContainerUIHostingContro
             .$supportedOrientations
             .receive(on: RunLoop.main)
             .removeDuplicates()
-            .sink(receiveValue: { _ in
+            .sink(receiveValue: { [weak self]_ in
                 switch containerView.router.currentView {
                 case .setupView:
-                    guard self.traitCollection.userInterfaceIdiom == .phone else {
+                    guard let self = self, self.traitCollection.userInterfaceIdiom == .phone else {
                         return
                     }
                     if UIDevice.current.isGeneratingDeviceOrientationNotifications {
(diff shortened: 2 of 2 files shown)


jimchou-dev (dismissed):

vhuseinova-msft (dismissed):

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Presentation/SwiftUI/Container/ContainerUIHostingController.swift
```diff
@@ -62,10 +62,10 @@ class ContainerUIHostingController: UIHostingController<ContainerUIHostingContro
             .$supportedOrientations
             .receive(on: RunLoop.main)
             .removeDuplicates()
-            .sink(receiveValue: { _ in
+            .sink(receiveValue: { [weak self]_ in
```
AmyL219, line 65: one space after `[weak self] `

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Presentation/SwiftUI/Container/ContainerUIHostingController.swift
```diff
@@ -62,10 +62,10 @@ class ContainerUIHostingController: UIHostingController<ContainerUIHostingContro
...
                 switch containerView.router.currentView {
                 case .setupView:
-                    guard self.traitCollection.userInterfaceIdiom == .phone else {
+                    guard let self = self, self.traitCollection.userInterfaceIdiom == .phone else {
```
AmyL219, line 68: switch another line

palatter (dismissed):
Tested.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Presentation/SwiftUI/Container/ContainerUIHostingController.swift
kevinyulu: @AmyL219  thanks for fixing. (ps: Im surprised that SwiftLint didn't catch this one)

AmyL219 (approved):

vhuseinova-msft (approved):

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Presentation/SwiftUI/Calling/CallingViewModel.swift
```diff
@@ -123,7 +123,10 @@ class CallingViewModel: ObservableObject {
         }
 
         if callHasConnected != newIsCallConnected && newIsCallConnected {
-            DispatchQueue.main.asyncAfter(deadline: .now() + 2.5) {
+            DispatchQueue.main.asyncAfter(deadline: .now() + 2.5) { [weak self] in
```
vhuseinova-msft, line 126: Does it create any problems? we don't have a link to the closure and there shouldn't be a need for weak self

vhuseinova-msft (approved):

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Presentation/SwiftUI/Calling/CallingViewModel.swift
vhuseinova-msft: I think in this case we are safe as we don't have links to the closure ([here](https://www.swiftbysundell.com/questions/is-weak-self-always-required/) is a good article on this).

AmyL219: Yeah, typically it would keep the strong self for 2.5 second while wait period, if the composite is dismissed. I feel it's better to keep using [weak self] in all closure. It's more consistent and easy to remember. 
This is similar to the timers' retain cycle issue. If we don't invalidate timer, it would keep retain cycle after the composite is dismissed. 
But once timer is invalidated because of interruption end, those retain cycles would be removed outside the callComposite

PR #238, opened 2022-06-06
[BugFix] Fix the issue that local video status is inconsistent after call resume from background

AmyL219:
## Purpose

* When Siri is triggered, the composite would go to `background` mode first, we would trigger the `CameraPausedSucceeded` action for background mode, which update the camera status to `paused`. 
* Update the localVideoStream id to nil. This would prevent the local video pip displaying the video when camera is `.off` icon
* Then interruption ended would be triggered, call become `onHold` even in the background mode. 
* When back to foreground, add the `callingState == .localHold` check, resume the local video stream even the call is inHold state

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```


palatter (approved):
Looks good and tested.

jimchou-dev (approved):
LGTM

PR #241, opened 2022-06-06
[BugFix] Fix the issue that local pip view doesn't show up after removing local video stream Id after video pasued

AmyL219:
## Purpose

* There is a bug in the localVideoView that if we update the localvideoStream quickly from `string` to `nil` back to `string`, the video stream wouldn't show up. Revert the code to prevent the issue. 
* Update the localVideoView to check the `CameraOperationalStatus`. Make the video button in the control bar and localVideoView consistently when determine if we should display the local video stream

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```


vhuseinova-msft (approved):

jimchou-dev (approved):
LGTM

kevinyulu (approved):

PR #249, opened 2022-06-09
[Refactor] Separate CallCompositeError with internal error code  

AmyL219:
## Purpose

* Rename CompositeError with CallCompositeInternalError. Use CallCompositeInternalError inside the composite's Calling service layer, Redux layer and viewModel layer

*  [ErrorState](AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Redux/State/ErrorState.swift) 

1. `internalError`: the internalErrorType defined from CallingComposite
2. `error`: the error throw out from `CallingSDKWrapper`. Note:`CallingSDKWrapper` could also throw out `CallCompositeInternalError`, but if so, we would assign the internal error to `ErrorState.internalError` rather than `ErrorState.error`
3. `errorCategory`, this part I feel it could be optimized in future.  we can separate `CallJoin`, `CallEnd` further. Right now CallJoin and CallEnd could be either fatal and callingStatus. See the new created ticket: [Task 2887925](https://skype.visualstudio.com/SPOOL/_workitems/edit/2887925): [iOS] Fatal Error category optimization

*  [CompositeErrorManager.swift](https://github.com/Azure/communication-ui-library-ios/pull/249/files#diff-2950f3b6adb169d9eae80865597d985db159533c8f758d89256bf25dc89e0f69) would 

1. only update ErrorHandler when the internalErrorType could be translated into `CallCompositeErrorCode`, and then throw out to Contoso User
2. only exit the composite when the `internalErrorType` is fatal and `errorState.category` is fatal

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code


AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Redux/Reducer/ErrorReducer.swift
```diff
@@ -11,24 +11,29 @@ struct ErrorReducer: Reducer {
             return state
         }
 
-        var error = state.error
+        var errorType = state.internalError
+        var callingError = state.error
```
AmyL219, line 15: should rename `callingError` to `error`

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Utilities/CallCompositeInternalError.swift
```diff
@@ -0,0 +1,54 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import Foundation
+
+enum CallCompositeInternalError: String, LocalizedError, Equatable {
```
JoshuaLai, line 8: If this is internal does it need to be a `LocaliazedError`? 
Should we just conform to `Error`? I bring this up cause having `localizedDescription` return a `NSLocalizedString` will do nothing cause I can't imagine we would ever want to localize these strings

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Presentation/Manager/CompositeErrorManager.swift
```diff
@@ -28,44 +28,39 @@ class CompositeErrorManager: ErrorManagerProtocol {
...
         }
+        previousError = errorState.internalError
+        updateEventHandler(state.errorState)
+        updateFatalError(state.errorState)
     }
 
-    private func update(error: CallCompositeError) {
-        guard self.error != error else {
+    private func updateEventHandler(_ errorState: ErrorState) {
+        guard let compositeError = getCallCompositeError(errorState: errorState),
```
JoshuaLai, line 41: The tiniest bit optimization, we can do a check for `didFail` first cause if it doesn't exist we don't nee to bother trying to create the `compositeError`

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Utilities/CallCompositeInternalError.swift
AmyL219: Yeah, we can replace `localizedError` with just `Error`.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Presentation/Manager/CompositeErrorManager.swift
AmyL219: Yeah, that's true

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Redux/Action/CallingAction.swift
```diff
@@ -36,11 +36,14 @@ struct ParticipantListUpdated: Action {
...
 struct ErrorAction: Action {
     struct FatalErrorUpdated: Action {
-        let error: CallCompositeError
+        let internalError: CallCompositeInternalError
+        let error: Error?
     }
 
     struct StatusErrorAndCallReset: Action {
-        let error: CallCompositeError
+        let internalError: CallCompositeInternalError
```
JoshuaLai, line 44: I am wondering if we still need both these `Action` now. The `internalError` now has a new method to return whether or not it is a fatal error so we might be able to just get away with having a single error action and check inside the internal error. Thoughts?

AmyL219: Yeah, I'd like to, but need to refactor further and discuss with Android side. There is a limitation here. The `callJoin` and `callEnd` reason is used both `fatal` and `statusUpdate`, for example, 
- like when we receiving `error` inside `endCall(state: ReduxState?, dispatch: @escaping ActionDispatch)` , we would mark the internalError with `callEnd` and `fatal`, which immediately quit the call. 
- like internet time out, we would use `callEnd`. which is a status error. 
- So I created  ([Task 2887925](https://skype.visualstudio.com/SPOOL/_workitems/edit/2887925): [iOS] Fatal Error category optimization), to refactor this part. We could have different internal error type for `fatal` `callEnd` and `statusUpdate` `callEnd`.

JoshuaLai: I understand the theory here. The thought process on the Android side is a little different for how they use the internal error if memory serves. 
It gets treated as "internal" today but in the future this "internal" will be refactored to an event that will be notified to developers in a new event handler. I think this is fine for now though so happy to approve this PR.

JoshuaLai (approved):

palatter (approved):
Looks good, just a question about how we return from call failed

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Service/Calling/CallingSDKWrapper.swift
```diff
@@ -76,7 +76,7 @@ class CallingSDKWrapper: NSObject, CallingSDKWrapperProtocol {
 
             self.callAgent?.join(with: joinLocator, joinCallOptions: joinCallOptions) { [weak self] (call, error) in
                 guard let self = self else {
-                    return promise(.failure(CompositeError.invalidSDKWrapper))
+                    return promise(.failure(CallCompositeInternalError.callJoinFailed))
                 }
 
                 if let error = error {
```
palatter, line 82: We can return here without callJoinFailed type error?

AmyL219: Yeah, after GA we could introduce a new internal type, which is different from callJoinFailed. 
But this function either returns .success or failure with an error
```
func joinCall(isCameraPreferred: Bool, isAudioPreferred: Bool) -> Future<Void, Error>
```

PR #258, opened 2022-06-15
[Bugfix] Add one fix for testLaunch() failure in AppCenter by removing the unneeded testLaunch function

kevinyulu:
## Purpose

Add fix for two UI tests issues in AppCenter
* Removed testLaunch in UItest, it's already tested in setup() function in XCUITestBase

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
(Locally) Run the XCUITest in the project locally
(Pipeline & App center) trigger an run job in UITest app center pipeline
```

## What to Check
* see test the test code passes in both local and app center


jimchou-dev (dismissed):

AzureCommunicationUI/AzureCommunicationUIDemoApp/AzureCommunicationUIDemoAppUITests/Utilities/XCUIElementExtension.swift
```diff
@@ -10,6 +10,10 @@ extension XCUIElement {
         tap()
         UIPasteboard.general.string = text
         guard UIPasteboard.general.hasStrings else {
+            // Workaround for a few iOS test devices, to be sure tap happens on the text field
+            sleep(1)
```
JoshuaLai, line 14: Can you give a short explanation to why we need this `sleep(1)`

AzureCommunicationUI/AzureCommunicationUIDemoApp/AzureCommunicationUIDemoAppUITests/Utilities/XCUIElementExtension.swift
```diff
@@ -10,6 +10,10 @@ extension XCUIElement {
...
+            tap()
+            tap()
```
JoshuaLai, line 16: Should we just use `doubleTap()` here? Or `tap(withNumberOfTaps: 2 numberOfTouches: 1)`

kevinyulu: @JoshuaLai  I was referring to the approach taken in this post https://github.com/mozilla-mobile/firefox-ios/issues/5826. The error is pretty similar and I think it's related to UI element not getting focused. I think doubleTap() or `tap(withNumberOfTaps: 2)` would also work. The concern is that these error is from App Center and they are not reproducible (or not very likely to repro) on local environment and to get it tested, we'll need to see how it works in App Center.

JoshuaLai: I'm not really sure I understand. The post linked here doesn't really reference anything about putting in `sleep` or `tap()` `tap()`

kevinyulu: this post has the fix they did for the similar issue we have: https://github.com/mozilla-mobile/firefox-ios/pull/5831/files

JoshuaLai (dismissed):

AzureCommunicationUI/AzureCommunicationUIDemoApp/AzureCommunicationUIDemoAppUITests/Utilities/XCUIElementExtension.swift
AmyL219: Does this can fix the UI test? Could we run 10 times to make sure this change would fixed the UI test issue

kevinyulu: yes I am trying these new tests in app center now. will get an result in a bit

kevinyulu: it looks like this is not fixing the issue, as it appears that the test can still fail in app center tests. I am going to remove this code change. https://appcenter.ms/orgs/Azure-Communication-Services/apps/Azure-Communication-Services-Call-UI-Library-Sample-iOS-Internal/test/runs/a1c66a08-dbf0-4d9d-b15f-dbf35ed2fea3

JoshuaLai (approved):

AmyL219 (approved):

jimchou-dev (approved):

PR #262, opened 2022-06-15
[Bug] Accessibility hold overlay should be solid colour 

JoshuaLai:
## Purpose
As per figma: 
Screen Shot 2022-06-15 at 3 00 40 PM
The hold overlay should have a solid colour and not have the same background colour like the lobby overlay. 
Introducing new way to set background via the protocol

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```


jimchou-dev (approved):

AmyL219 (commented):
Approve. But 
Does accessibility team mention why `onLobby` view could be transparent?
Does this also need to be applied in Android design
Does Alex know the update?

AmyL219 (approved):

JoshuaLai:
> Approve. But Does accessibility team mention why `onLobby` view could be transparent? 
Thats a good point we will need to bring up with the Accessibility team
> Does this also need to be applied in Android design 
Let me chat with them to make sure. This design is actually what was originally intended and in figma
> Does Alex know the update?
Yes, he was the one that called this out, this is as per figma design.

palatter (approved):
Looks good!

PR #268, opened 2022-06-16
[Bug] Fix the issue that resume a call without internet would stop user from exit

AmyL219:
## Purpose

* The `.remoteHold` would happen randomly when try to resume a call without internet. The critical scenario is after the the internet is connected again, the user wouldn't be able to leave the composite even the user tap the call end button. This is a CallingSDK bug need to be fixed in server side. 
* Confirmed with Sanath that .remoteHold is applied only on 1:1 meeting which our UI library doesn't have, so it's safe to quit the composite when the status become .remoteHold
* Add the current workaround to directly leave the call when the calling status become `.remoteHold`. 

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```


JoshuaLai:
The code change is fine but lets update the changelog with this fix.

kevinyulu (dismissed):

JoshuaLai (approved):

kevinyulu (approved):

jimchou-dev (approved):

PR #289, opened 2022-06-27
[Refactor] Switch actions to an enumerated types

emlynmac:
## Purpose
Currently, the redux implementation is using an open protocol. This allows any object to become an action and forces run time dynamic resolution of actions and functionality.

This PR changes the actions to be a enumeration hierarchy, which enables the compiler to see types at compile time and as a result can ensure that we have the correct actions being handled properly.

The need for class checking at run time is removed, replaced with compile-time checking of enumeration instances with associated data. This removes much of the `if let <value> as? <type>` and associated tests.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
Follow the usual testing procedure; code functionality should be unchanged.

## What to Check
All functionality should be unchanged.

## Other Information
I will be discussing this PR and related ideas at the meeting tomorrow.


AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/DI/DependancyContainer.swift
```diff
@@ -72,18 +72,11 @@ final class DependencyContainer {
...
-            callingMiddlewareHandler: middlewaresHandler)]
-        let appStateReducer = AppStateReducer(permissionReducer: PermissionReducer(),
-                                              localUserReducer: LocalUserReducer(),
-                                              lifeCycleReducer: LifeCycleReducer(),
-                                              audioSessionReducer: AudioSessionReducer(),
-                                              callingReducer: CallingReducer(),
-                                              navigationReducer: NavigationReducer(),
-                                              errorReducer: ErrorReducer())
+        let middlewares: [Middleware] = [ liveCallingMiddleware(callingMiddlewareHandler: middlewaresHandler)
+        ]
```
jimchou-dev, line 76: We have it as an array previously also, but can someone help to explain why does this needs to be an array?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Redux/Middleware/CallingMiddleware.swift
```diff
@@ -5,54 +5,74 @@
 
 import Combine
 
-struct CallingMiddleware: Middleware {
-    private let actionHandler: CallingMiddlewareHandling
+extension Middleware {
+    static func liveCallingMiddleware(callingMiddlewareHandler actionHandler: CallingMiddlewareHandling)
```
JoshuaLai, line 9: nitpick: I think we can think up of a better name. I am not sure if `liveCallingMiddleware` gives us context to know what this method does. 
Maybe `applyActionHandler`?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Redux/Action/Actions.swift
```diff
@@ -0,0 +1,18 @@
...
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import Foundation
+
+///
+/// Actions for the entire library. All actions are defined here as a heirarchy of enum types
+/// 
+enum Actions: Equatable {
```
JoshuaLai, line 11: minor nitpick: `Actions` I personally would like to see as `Action` this aligns closer with its predecessors name and aligns with the Android implementations name. 
Also IMO it just looks odd cause plural implies its a list so I feel it would look odd if it was `[Actions]`. 
Super minor though

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Redux/Middleware/CallingMiddleware.swift
```diff
@@ -5,54 +5,74 @@
...
+                            case .callStartRequested:
+                                actionHandler.startCall(state: getState(), dispatch: dispatch)
+                            case .callEndRequested:
+                                actionHandler.endCall(state: getState(), dispatch: dispatch)
+                            case .holdRequested:
+                                actionHandler.holdCall(state: getState(), dispatch: dispatch)
+                            case .resumeRequested:
+                                actionHandler.resumeCall(state: getState(), dispatch: dispatch)
+                            default:
+                                break
```
JoshuaLai, line 29: I know we discussed about breaking this up into smaller helper methods, but now that this is in a static methodit might not be any more elegant. Just some food for thought here.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Redux/Store.swift
```diff
@@ -6,40 +6,37 @@
...
         self.state = state
         self.dispatchFunction = middlewares
             .reversed()
             .reduce({ [unowned self] action in
                 self._dispatch(action: action)
             }, { nextDispatch, middleware in
-                let dispatch: (Action) -> Void = { [weak self] in self?.dispatch(action: $0) }
-                let getState = { [weak self] in self?.state }
-                return middleware.apply(dispatch: dispatch, getState: getState)(nextDispatch)
+                let dispatch: (Actions) -> Void = { [unowned self] in self.dispatch(action: $0) }
```
JoshuaLai, line 27: I believe using `unowned` here is ok cause self should still be there. However as general best practice I think I do still prefer us use `weak` and not risk the off chance of a run time crash.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICallingTests/Redux/Reducer/ErrorReducerTests.swift
```diff
@@ -12,120 +12,84 @@ class ErrorReducerTests: XCTestCase {
...
-    func test_handleErrorReducer_reduce_when_notErrorState_then_return() {
-        let state = StateMocking()
-        let action = ErrorAction.FatalErrorUpdated(internalError: .callJoinFailed,
-                                                   error: nil)
-        let sut = getSUT()
-
-        let resultState = sut.reduce(state, action)
-        XCTAssert(resultState is StateMocking)
-    }
+    // No longer possible with compile time state checking
```
JoshuaLai, line 15: Yep this makes sense. In this case lets just delete this test since it is no longer relevant

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICallingTests/Redux/Reducer/ErrorReducerTests.swift
```diff
@@ -12,120 +12,84 @@ class ErrorReducerTests: XCTestCase {
...
 
     func test_handleErrorReducer_reduce_when_statusErrorCallDeniedAndCallReset_then_returnErrorState_categoryCallState() {
-        let state = ErrorState(internalError: .callDenied,
-                               error: nil,
-                               errorCategory: .callState)
-
-        let action = ErrorAction.StatusErrorAndCallReset(internalError: .callDenied,
-                                                         error: nil)
+        let state = ErrorState()
+        // Should set the initial state to something other than the expected state so we know the reducer did something
```
JoshuaLai, line 62: I assume with this comment you are suggesting a change to this specific unit test to use a different error state? Or perhaps no error state at all? And then update the error state to be `.callDenied`

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Redux/Middleware/CallingMiddleware.swift
emlynmac: The live version of the calling middleware implementation here was the logic for the name.

I could go with  `Middleware.callHandling` maybe?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Redux/Action/Actions.swift
emlynmac: Yes, completely agree. I deferred the thinking about name until discussion with the group. 
My preference would be something like, `LibraryAction` or `ActionType` just because `Action` is used to specify generics elsewhere.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Redux/Middleware/CallingMiddleware.swift
emlynmac: I personally like seeing all the cases clearly.
If we remove the `default` case and implement every un-used case, then it will be a _lot_ of cases and I can see value in breaking out the handlers to other functions.

I'll have a look at both and check in the the one I think looks better, removing the default handlers.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Redux/Store.swift
emlynmac: The store must always outlive the dispatch. This makes it safe to call unowned and as a result remove the possibility of an optional to handle.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICallingTests/Redux/Reducer/ErrorReducerTests.swift
emlynmac: Will clean up

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICallingTests/Redux/Reducer/ErrorReducerTests.swift
emlynmac: Some of the tests were setting the initial state the same as the expected state, which means you don't get a real test. I think this is a comment I wrote to call that out - I'll clean it up.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Redux/Action/LocalUserAction.swift
```diff
@@ -6,59 +6,57 @@
...
-    struct CameraOnTriggered: Action {}
-    struct CameraOnSucceeded: Action {
-        var videoStreamIdentifier: String
-    }
-    struct CameraOnFailed: Action {
-        var error: Error
-    }
+enum LocalUserAction: Equatable {
+    static func == (lhs: LocalUserAction, rhs: LocalUserAction) -> Bool {
+        guard type(of: lhs) == type(of: rhs) else {
```
vhuseinova-msft, line 11: I'm wondering if this will always return true as `type(of: lhs)` and `type(of: rhs)` will return `LocalUserAction`

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Redux/Action/LocalUserAction.swift
```diff
@@ -6,59 +6,57 @@
...
-    }
+            return (lErr as NSError).code == (rErr as NSError).code
 
-    struct CameraSwitchTriggered: Action {}
-    struct CameraSwitchSucceeded: Action {
-        var cameraDevice: CameraDevice
-    }
-    struct CameraSwitchFailed: Action {
-        var error: Error
+        default:
```
vhuseinova-msft, line 27: It would be good to add checks for other cases with Associated Values

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Redux/State/RemoteParticipantsState.swift
```diff
@@ -5,7 +5,7 @@
 
 import Foundation
 
-class RemoteParticipantsState: ReduxState {
+class RemoteParticipantsState {
```
vhuseinova-msft, line 8: Should `RemoteParticipantsState` be updated to a struct?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Presentation/SwiftUI/Calling/CallingViewComponent/ControlBarViewModel.swift
```diff
@@ -108,15 +108,15 @@ class ControlBarViewModel: ObservableObject {
         }
 
         isCameraStateUpdating = true
-        let action: Action = cameraState.operation == .on ?
-            LocalUserAction.CameraOffTriggered() : LocalUserAction.CameraOnTriggered()
-        dispatch(action)
+        let action: LocalUserAction = cameraState.operation == .on ?
+            LocalUserAction.cameraOffTriggered : LocalUserAction.cameraOnTriggered
```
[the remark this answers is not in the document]
emlynmac: Updated

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Redux/Action/LocalUserAction.swift
emlynmac: Good catch! 
Exploring this in a playground does not work as I thought it would. I will rework.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Redux/Action/LocalUserAction.swift
emlynmac: Have an update for this as a more general case.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Redux/State/RemoteParticipantsState.swift
emlynmac: Updated

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Redux/Reducer/Reducer.swift
```diff
@@ -6,6 +6,6 @@
 import Combine
 import Foundation
 
-protocol Reducer {
-    func reduce(_ state: ReduxState, _ action: Action) -> ReduxState
+struct Reducer<State, Action> {
+    var reduce: (_ state: State, _ action: Action) -> State
```
waynemo, line 10: Wondering why we have this as a var instead of an initializer. Do we need to update the `reduce` property?

waynemo (dismissed):

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Redux/Reducer/Reducer.swift
emlynmac: The var defines the interface for the struct. 
There is no explicit initializer as it depends on context from the generics.
When we use the implicit initializer, it creates the same interface.

emlynmac: Updated after discussion; we're not using the override ability currently so this can be changed to a `let` for now

vhuseinova-msft (dismissed):

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Redux/Action/Actions.swift
emlynmac: Ah ha - ok consistency. Will rename to `Action` and change the generic to `ActionType`

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Redux/Reducer/CallingReducer.swift
```diff
@@ -5,28 +5,42 @@
...
 import Combine
 
-struct CallingReducer: Reducer {
-    func reduce(_ state: ReduxState, _ action: Action) -> ReduxState {
-        guard let callingState = state as? CallingState else {
-            return state
-        }
+extension Reducer where State == CallingState,
+                        Actions == Action {
+    static var liveCallingReducer: Self = Reducer { callingState, action in
```
AmyL219, line 10: could Reducer be `non-static`?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Redux/Store.swift
```diff
@@ -6,27 +6,27 @@
...
 
-class Store<T: ReduxState>: ObservableObject {
+class Store<State>: ObservableObject {
 
-    @Published var state: T
+    @Published var state: State
 
     private var dispatchFunction: ActionDispatch!
-    private let reducer: Reducer
+    private let reducer: Reducer<State, Action>
```
AmyL219, line 14: This `Action` is still not generic, correct? And we couldn't reuse the Store in ChatComposite with fully flexibility
What if we package the `Store` to a package? I think we have to also package `Action` into the package. otherwise it would throw the warning. But `Action` is only specific to `CallingComposite` in it's definition

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Redux/Reducer/AppStateReducer.swift
```diff
@@ -5,88 +5,64 @@
...
+            var callingState = state.callingState
+            var remoteParticipantState = state.remoteParticipantsState
+            var navigationState = state.navigationState
+            var errorState = state.errorState
+            var audioSessionState = state.audioSessionState
 
-        if let newLifeCycleState = lifeCycleReducer.reduce(state.lifeCycleState, action) as? LifeCycleState {
-            lifeCycleState = newLifeCycleState
-        }
+            if case let .permissionAction(permAction) = action {
```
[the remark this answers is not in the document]
emlynmac: Case let is essentially a switch but then resolving that type to the action.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Redux/Reducer/LifeCycleReducer.swift
```diff
@@ -5,19 +5,16 @@
 
 import Foundation
 
-struct LifeCycleReducer: Reducer {
-    func reduce(_ state: ReduxState, _ action: Action) -> ReduxState {
-        guard let appLifeCycleCurrentState = state as? LifeCycleState else {
-            return state
-        }
+extension Reducer where State == LifeCycleState,
+                        Actions == LifecycleAction {
```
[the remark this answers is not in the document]
emlynmac: We always want to limit the possible values allowed to a function to the minimal types that it needs. 
Scoping the actions appropriately ensures that we don't need to write unnecessary handlers and also makes it very clear what the scope of the reducing function is.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/DI/DependancyContainer.swift
AmyL219: Because we could have more than one middleware in future. 
We could have callMiddleware, we could have audioSessionMiddleware.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Redux/Reducer/AppStateReducer.swift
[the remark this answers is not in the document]
AmyL219: `if case let ` should be equal to `if`. So the action would check all those if statement. The performance is same with casting

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Redux/Reducer/LocalUserReducer.swift
```diff
@@ -80,32 +78,32 @@ struct LocalUserReducer: Reducer {
...
-        case .speaker:
-            return .speakerRequested
-        case .receiver:
-            return .receiverRequested
-        case .bluetooth:
-            return .bluetoothRequested
-        case .headphones:
-            return .headphonesRequested
-        }
+private func getRequestedDeviceStatus(for audioDeviceType: AudioDeviceType)
```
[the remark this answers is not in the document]
AmyL219: I don't feel it's a good practice

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Redux/Reducer/AppStateReducer.swift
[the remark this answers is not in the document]
emlynmac: Yes, we can do it that way to avoid a `case let`; but the whole method cannot be wrapped up in a switch as the control flow can mean multiple reducers handle a single action.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Redux/Store.swift
vhuseinova-msft: We would be able to reuse `Store` in `ChatComposite` even without generics but when `ChatComposite` will have its own implementation of `Action`.

vhuseinova-msft (approved):

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Redux/Reducer/LocalUserReducer.swift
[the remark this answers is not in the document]
emlynmac: In general, global writable variables are evil and should be avoided at all times.

This is a file private function, which gives it the same scope as a private method on a class. 
The function is pure and is only applicable to the file.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Redux/Middleware/CallingMiddleware.swift
emlynmac: Task created for follow up https://skype.visualstudio.com/SPOOL/_sprints/taskboard/ACS%20Mobile%20UI%20-%20iOS/SPOOL/CY2022-Q3/Sprint%2066%20(Ends%20July%2010)?workitem=2912031

emlynmac:
I've created a task for follow up of some of the discussion points here in https://skype.visualstudio.com/SPOOL/_workitems/edit/2912031

jimchou-dev (approved):

JoshuaLai (approved):

kevinyulu (approved):

AzureCommunicationUI/sdk/AzureCommunicationUICalling/AzureCommunicationUICalling/Redux/Reducer/LocalUserReducer.swift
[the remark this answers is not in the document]
emlynmac: To be discussed as part of https://skype.visualstudio.com/SPOOL/_workitems/edit/2912031

PR #290, opened 2022-06-28
[Refactor] improved UI test to use new acs token every time joining a call instead of reusing the same asc token

kevinyulu:
## Purpose

* Improved UI test to use new acs token every time joining a call instead of reusing the same asc token

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* Run the XCUItest to see if all tests are passed.

### DIFF
--- a/AzureCommunicationUI/AzureCommunicationUIDemoApp/AzureCommunicationUIDemoApp/Views/SettingsView.swift
+++ b/AzureCommunicationUI/AzureCommunicationUIDemoApp/AzureCommunicationUIDemoApp/Views/SettingsView.swift
@@ -52,6 +52,15 @@ struct SettingsView: View {
         Section(header: Text("Local Participant Settings")) {
             Toggle("Use expired token", isOn: $envConfigSubject.useExpiredToken)
                 .accessibilityIdentifier(AccessibilityId.expiredAcstokenToggleAccessibilityID.rawValue)
+            Button("Generate a new token") {
+                let url = URL(string: envConfigSubject.acsTokenUrl)!
+                AuthenticationHelper.getCommunicationToken(tokenUrl: url)() { token, error in
+                    if let token = token {
+                        debugPrint("hay! token has been manually updated: \(token), error is \(error)")
+                        envConfigSubject.acsToken = token
+                    }
+                }
+            }.accessibilityIdentifier(AccessibilityId.getNewAcstokenButtonAccessibilityID.rawValue)
         }
     }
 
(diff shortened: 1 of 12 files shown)


AzureCommunicationUI/AzureCommunicationUIDemoApp/AzureCommunicationUIDemoApp/Views/SettingsView.swift
```diff
@@ -52,6 +52,15 @@ struct SettingsView: View {
         Section(header: Text("Local Participant Settings")) {
             Toggle("Use expired token", isOn: $envConfigSubject.useExpiredToken)
                 .accessibilityIdentifier(AccessibilityId.expiredAcsTokenToggleAccessibilityID.rawValue)
+            Button("Generate a new token") {
+                let url = URL(string: envConfigSubject.acsTokenUrl)!
```
AmyL219, line 56: Why do we want to have the acsTokenUrl? Which acsTokenUrl we would like to use though? 
The acsTokenUrl we want to use in the E2E test need to be secured. A self host token url  ( like `https://xxxxx.azurewebsites.net/api/Auth?code=...` without authentication is not allowed.

kevinyulu: Would we use this acsTokenUrl in the E2E test?
Yes it's used in some UI tests:

```
testCallCompositeEndCallGroupCallSwiftUI()
testCallCompositeEndCallTeamsCallSwiftUI()
testCallCompositeEndCallGroupCallUIKit()
testCallCompositeEndCallTeamsCallUIKit()
```
Which acsTokenUrl we would like to use though?
I think currently I test it with a self host one in my local test. we can use a secured one in the info.plist from app center / local. it should work.

palatter: So if we use a secured url, which user is used when running the pipeline?

AmyL219: Secured one would need user to login to get the access token. 
The pipeline has one step to fetch the token every run inside. 
See this step `Get access token in pip3` in this pipeline: https://skype.visualstudio.com/SCC/_apps/hub/ms.vss-ciworkflow.build-ci-hub?_a=edit-build-definition&id=12410

kevinyulu:
closed pull request as the the team has not made a decision of the approach to use acs token url in the ui test

PR #297, opened 2022-07-25
[Bugfix] Fixed Participant Drawer Height Calculation

jpeng-ms:
## Purpose
The calculation of participant drawer height needs to be fixed so that user would not have content being cutoff when having large font enabled in accessibility.

## Context
The drawer container calculation was previously based on visible cells which is fine in the most of cases. 

But in some cases, the visible cells could be empty due to the fact that we had been calling resizeDrawer() multiple times which results in a race condition where tableview could be not ready and having no visible cells. Then, since there are no visible cells, current logic is to use a fixed default height for each cell and this height becomes too low (causing cutoff) when user had large fonts enabled in accessibility.

## Solution

To resolve this issue, we need to:
1. eliminate repeated calls to resizeDrawer() while table reload is in progress
2. use drawing area for each row to get cell height instead

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code
*  enable large font in accessibility
*  enable zooms in display setting 
*  open the app
*  enter information such as token, display name and teams link
*  join the call
*  tap on participant list, see if drawer container shows up properly
*  tap on end call button, see if drawer container shows all content in the screen without being cutting off
*  tap on audio button, see if drawer shows up properly

## What to Check
* drawer should not be cutoff
* drawer contents are visible to contoso (except in landscape mode where user might need to scroll)

### How to reproduce original issue:
1. enable zoom and set font size to large in accessibility
2. open the app and join a call
3. rotate the device 360 degrees
4. tap on end call button, you would noticed that drawer got cutoff5. 

OR

1. enable zoom and set font size to large in accessibility
2. open the app and join a call
3. tap on end call button and hold it for one second
4. release the button, drawer should now show up and got cutoff

### Before - resizeDrawer being called multiple time which caused a race condition:
image
image

### After - resizeDrawer being called exactly once for each tap:
image

### Before:
image

### After:
image

### Uneven row height test:
image

### DIFF
--- a/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/FluentUI/Wrapper/DrawerContainerViewController.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/FluentUI/Wrapper/DrawerContainerViewController.swift
@@ -54,18 +54,12 @@ class DrawerContainerViewController<T>: UIViewController, DrawerControllerDelega
         }
     }
 
-    override func viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator) {
-        super.viewWillTransition(to: size, with: coordinator)
-        resizeDrawer()
-    }
-
     func dismissDrawer(animated: Bool = false) {
         self.controller?.dismiss(animated: animated)
     }
 
     func updateDrawerList(items: [T]) {
         self.items = items
-        resizeDrawer()
     }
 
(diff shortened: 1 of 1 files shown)


palatter:
This looks like a good solution! Just testing all the cases now.

palatter (approved):
Tested and looks good!

JoshuaLai (approved):
LGTM

PR #301, opened 2022-07-27
[Feature] Device Manager failure throws new unknown error

JoshuaLai:
## Purpose
Error Handling: In the rare case that device manager fails,  we want to throw an error to our error event handler. 

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

### DIFF
--- a/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Service/Calling/CallingSDKWrapper.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Service/Calling/CallingSDKWrapper.swift
@@ -250,10 +250,10 @@ extension CallingSDKWrapper {
                 guard let self = self else {
                     return
                 }
-                if let error = error {
-                    self.logger.error("Failed to get device manager instance")
-                    return promise(.failure(error))
+                guard let deviceManager = deviceManager, error == nil else {
+                    return promise(.failure(CallCompositeInternalError.deviceManagerFailed(error)))
                 }
+
                 self.deviceManager = deviceManager
                 self.deviceManager?.delegate = self
                 return promise(.success(()))
--- a/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Utilities/CallCompositeInternalError.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Utilities/CallCompositeInternalError.swift
@@ -58,25 +58,16 @@ enum CallCompositeInternalError: Error, Equatable {
 extension CallCompositeInternalError {
     static func == (lhs: CallCompositeInternalError, rhs: CallCompositeInternalError) -> Bool {
         switch(lhs, rhs) {
-        case (.deviceManagerFailed, .deviceManagerFailed):
-            return true
-        case (.callTokenFailed, .callTokenFailed):
-            return true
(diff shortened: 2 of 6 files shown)


Leah-Xia-Microsoft (dismissed):
LGTM

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Utilities/CallCompositeInternalError.swift
```diff
@@ -50,3 +54,32 @@ enum CallCompositeInternalError: String, Error, Equatable {
         }
     }
 }
+
+extension CallCompositeInternalError {
+    static func == (lhs: CallCompositeInternalError, rhs: CallCompositeInternalError) -> Bool {
+        switch(lhs, rhs) {
+        case (.deviceManagerFailed, .deviceManagerFailed):
```
vhuseinova-msft, line 61: `cases` that return `true` can be cobined in 1 `case`

vhuseinova-msft (dismissed):

vhuseinova-msft (dismissed):

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Utilities/CallCompositeInternalError.swift
```diff
@@ -50,3 +54,23 @@ enum CallCompositeInternalError: String, Error, Equatable {
         }
     }
 }
+
+extension CallCompositeInternalError {
+    static func == (lhs: CallCompositeInternalError, rhs: CallCompositeInternalError) -> Bool {
```
emlynmac, line 59: We need the explicit equality operator for this enum now because `Error` is not equatable. 

With swift 5.7, we get existential types, which means we can start this sort of thing:

```
func equals(_ lhs: Any, _ rhs: Any) -> Bool {
  func unpack<A: Equatable>(_ lhs: A, _ rhs: Any) -> Bool {
    lhs == (rhs as? A)
  }

  guard let lhs = lhs as? any Equatable else { 
    return false
  }

  return unpack(lhs, rhs)
}
```

As long as the type of the Error is Equatable, we can use that but otherwise it fails.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Service/Calling/CallingSDKWrapper.swift
```diff
@@ -250,10 +250,10 @@ extension CallingSDKWrapper {
                 guard let self = self else {
                     return
                 }
-                if let error = error {
-                    self.logger.error("Failed to get device manager instance")
-                    return promise(.failure(error))
+                guard let deviceManager = deviceManager, error == nil else {
+                    return promise(.failure(CallCompositeInternalError.deviceManagerFailed(error)))
```
AmyL219, line 254: Any specific reason why do we want to have this error as `unknownError`, rather than `callJoin` error? since this error only could be happen when we try to join the call

emlynmac (dismissed):
LGTM.
We have some refactoring coming to address some of the other discussions

emlynmac (approved):
Reapproved after adding changelog

jimchou-dev (approved):

PR #306, opened 2022-08-08
[Bug] End Ongoing Call if User Force Quits

jpeng-ms:
## Purpose
The goal of this PR is to end the ongoing call if user closes the app unexpectedly.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## How to Test
0. launch a call on your side
1. run the demo app in simulator and join the call
2. swipe from the bottom to go to app switch view
3. swipe up to close demo app
4. check to see if demo app participant has been removed from the app

## What to Check
1. when demo app closes, its caller should be removed from the call

## Other Information
1. the end call has a delay of up to few seconds (may takes few seconds for participant to be removed). This is consistent with ending the call explicitly using the end call button as well as Teams' iOS app.
2. if app got terminated from Xcode, then lifecycle event won't be sent to our app to initiate end call request. Participant would still remain in the call


Leah-Xia-Microsoft (commented):
Tested this PR, the caller is removed after about 20 -30 seconds, is it expected?

Leah-Xia-Microsoft:
Also the PR type should be bug instead of feature?

jpeng-ms:
> Tested this PR, the caller is removed after about 20 -30 seconds, is it expected?

let's have a chat

PR #307, opened 2022-08-09
[Task] Optimization token fetch refresher to avoid passing in nil

Leah-Xia-Microsoft:
## Purpose

* This refactor is based on the suggestion from the calling SDK team
* for CommunicationTokenRefreshOptions, it's recommended not passing pass initialToken as nil
```
CommunicationTokenRefreshOptions(initialToken: nil, refreshProactively: true, tokenRefresher: tokenRefresher)
```
* The reason being that "the problem is create call agent is blocked unnecessarily if you do it before that you know whether contoso has properly working token refresh url or not instead of trying to debug why createcallagent is stuck"

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

## What to Check
* Verify both UIKit demo app and SwiftUI demo app can still start a call


AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/UIKitDemoViewController.swift
```diff
@@ -208,8 +210,11 @@ class UIKitDemoViewController: UIViewController {
         case .tokenUrl:
             if let url = URL(string: acsTokenUrlTextField.text!) {
                 let tokenRefresher = AuthenticationHelper.getCommunicationToken(tokenUrl: url)
-                let refreshOptions = CommunicationTokenRefreshOptions(initialToken: nil, refreshProactively: true,
-                                                                                        tokenRefresher: tokenRefresher)
+                let initialToken = await TokenManager().fetchInitialToken(with: tokenRefresher)
```
JoshuaLai, line 213: Nitpick: Same comment for both UIKit and SwiftUI implementation. But now that we are making a service call we may need to disable to "Start Experience" button so users don't tap on it multiple times while we are waiting for this call. 
Not the biggest thing cause I think we are cleaning up properly.

Leah-Xia-Microsoft: > 

Good point. I'll just use the system activity indicator and disable button still for now? Or should i reach out to Alex for this?

JoshuaLai: No need to reach out to Alex since this is just for internal dev use. I think just disabling / reenable the button is fine

JoshuaLai (dismissed):

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Utilities/TokenManager.swift
```diff
@@ -0,0 +1,20 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import Foundation
+import AzureCommunicationCalling
+
+final class TokenManager {
```
kevinyulu, line 9: there seems to be a `AuthenticationHelper` that has token related function. Can we consider refactor out the function to be one helper/manager class?

kevinyulu (dismissed):

kevinyulu (approved):

JoshuaLai (approved):

PR #312, opened 2022-08-17
[Refactor] Use async functions in the Calling SDK Components

emlynmac:
## Purpose
Combine has been used for wrapping the calls to the underlying calling SDK, using `Future` to convert to a publisher.
These must then be consumed as such.

With Swift's latest concurrency tools, we can simplify this to use `async throwing` functions in a majority of places and set up the base for further asynchronous contexts.

```
[ ] Yes
[x] No
```

## Pull Request Type
```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
* Clone the repository
* Run the demo app and notice that there is no change in functionality.

## What to Check
* All unit tests pass
* Functionality is unaffected

### DIFF
--- a/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Service/Calling/CallingSDKWrapper.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Service/Calling/CallingSDKWrapper.swift
@@ -276,7 +276,6 @@ extension CallingSDKWrapper {
         }
     }
 
-    @MainActor
     private func change(_ videoStream: LocalVideoStream, source: VideoDeviceInfo) async throws {
         do {
             try await videoStream.switchSource(camera: source)
--- a/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Service/Calling/CallingSDKWrapper.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Service/Calling/CallingSDKWrapper.swift
@@ -7,6 +7,7 @@ import Foundation
 import Combine
 import AzureCommunicationCalling
 
+// swiftlint:disable all
 class CallingSDKWrapper: NSObject, CallingSDKWrapperProtocol {
     let callingEventsHandler: CallingSDKEventsHandling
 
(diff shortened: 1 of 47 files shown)


AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Service/Calling/CallingSDKWrapper.swift
```diff
@@ -299,38 +259,31 @@ extension CallingSDKWrapper {
...
-                    if error != nil {
-                        self?.logger.error( "Local video failed to switch camera. \(error!)")
-                        promise(.failure(error!))
-                        return
-                    }
-                    self?.logger.debug("Local video switched camera successfully")
-                    promise(.success(()))
-                }
-            }
+    @MainActor
```
emlynmac, line 279: I think this can be removed.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Tests/CallingMiddlewareHandlerMocking.swift
```diff
@@ -7,83 +7,114 @@ import Foundation
...
+    var requestCameraSwitchCalled: ((Bool) -> Void)?
+    var requestMicMuteCalled: ((Bool) -> Void)?
+    var requestMicUnmuteCalled: ((Bool) -> Void)?
+    var requestHoldCalled: ((Bool) -> Void)?
+    var requestResumeCalled: ((Bool) -> Void)?
 
-    func startCall(state: AppState, dispatch: @escaping ActionDispatch) {
-        startCallWasCalled = true
+    func setupCall(state: AppState, dispatch: @escaping ActionDispatch) -> Task<Void, Never> {
+        Task {
```
jpeng-ms, line 27: good work! but I was wondering if you think there's a case where we need to use detached task vs normal task?

Leah-Xia-Microsoft (approved):

vhuseinova-msft (approved):

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Tests/CallingMiddlewareHandlerMocking.swift
emlynmac: Let's chat about this offline.

PR #314, opened 2022-08-18
[Refactor] Updated App to Use FluentUI Colours

jpeng-ms:
## Purpose
This PR is meant to replace custom colours we defined in the app with FluentUI colours.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[X] Refactoring (no functional changes, no api changes)
[ ] Other... Please describe:
```

## How to Test
1. open digital color picker app on your Mac
2. open the test picture attached
3. compare before and after screenshots

## What to Check
1. make sure colours are identical in before and after.

## Other Information
1. This PR is not meant to fix UI bugs (colours that are not aligned with UI spec). A separate ticket has been raised for this.
2. Confirmed with UI that it's an intentional design decision to use colours across components (for example, using text palette for icons).
3. It's also a design decision to not follow with FluentUI dark/light mode colour combos, so I had made a custom function for this.
4. for colours that do not exist in FluentUI, no refactor was done in this case.   
image
5. comparing between test pictures is better test for this PR because it eliminates the colour differences in display's color profile. If you measure directly from simulator vs the test picture, the colour would not be the same unless you save the screenshot and paste it into Figma file of the test picture, then you would noticed that colours are actually matched up. Also notice that colour measured on external display would be different from built-in display. for example: 

image

#### (wait for few seconds for picture to be loaded.)
calling views


emlynmac (dismissed):
LGTM

kevinyulu (dismissed):

jimchou-dev (approved):

vhuseinova-msft (approved):

PR #317, opened 2022-08-22
[Feature] Show Alert When App is Offline

jpeng-ms:
## Purpose
The goal is to alert Contoso user they cannot join the call if their internet is down. 

## Technical details

This PR introduces a network manager which listens to system event on network condition and also responsible to dispatch actions to update network state in AppState, which would be used by our view accordingly. 

The overall flow is the following:
```
iOS notifies network manager on network changes ->
manager dispatch an network action(.lost or .restored) ->
action then gets consumed by network reducer ->
a new state (.online or .offline) gets produced -> 
set up view model which listens to state changes received new network state ->
ask error info view model to render/dismiss error banner
```

Previous implement was to use network action to produce an error state so that error banner can be updated automatically. But PM explicitly wanted the alert to be shown only when user taps on "Join Call" button, therefore, I am introducing network state to accommodate this request.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## How to Test
1. clone the repo
2. set target to demo app
3. select a physical device
4. click run 

## What to Check
1. when there's no active internet connection, an alert should be shown when you tap on "Join Call"

## Other Information
1. The scope of this ticket is limited to set up view only and we should not handle anything on calling view as it's calling SDK's task to let us know about call conditions.
2. Please use physical device for testing. The network changes on your laptop does not propagate to simulator properly and you might notice unexpected behaviour due to this. 

## Screenshots
image


AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/Manager/NetworkManager.swift
```diff
@@ -0,0 +1,32 @@
...
+    private let logger: Logger
+    private let monitor = NWPathMonitor()
+    private let networkMonitorQueue = DispatchQueue(label: "NetworkMonitorQueue")
+
+    init(store: Store<AppState>,
+         logger: Logger) {
+        self.store = store
+        self.logger = logger
+        monitor.pathUpdateHandler = { path in
+            if path.status != .satisfied {
```
kevinyulu, line 22: would this  handler be triggered when a wifi connection has been disconnected and network has been switched to cellular?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Redux/State/NetworkState.swift
```diff
@@ -0,0 +1,18 @@
...
+import Foundation
+
+enum NetworkStatus {
+    case online, offline
+}
+
+struct NetworkState {
+    let status: NetworkStatus
+
+    init(_ status: NetworkStatus = .online) {
```
kevinyulu, line 15: would this initial value be a safe assumption? how about we'd avoid giving an initial value and determine the network connectivity when initializing.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/Manager/NetworkManager.swift
jpeng-ms: Hi Kevin, since I initialized monitor as NWPathMonitor(), so it would monitor both wifi and cellular network. If we just wanted to monitor a particular type of network, we would then need to initiate it as 
```
let cellMonitor = NWPathMonitor(requiredInterfaceType: .cellular)
let wifiMonitor = NWPathMonitor(requiredInterfaceType: .wifi)
let ethernetMonitor = NWPathMonitor(requiredInterfaceType: .wiredEthernet)
```

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Redux/State/NetworkState.swift
jpeng-ms: > would this initial value be a safe assumption? how about we'd avoid giving an initial value and determine the network connectivity when initializing.

Thanks, I will be introducing a new state ```.unknown``` to accommodate this.

jpeng-ms: updated, thanks!

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/Manager/NetworkManager.swift
kevinyulu: do we know what type of connection are we monitoring in this case? I am thinking app doesn't quit the call when wifi is lost but cellular is still available.

jpeng-ms: Good point! According to Apple's website (see link above), it would monitor network condition from all available interfaces (such as wifi/cellular).

I have specifically tested out the scenario you asked, here's the log: 
image
since network condition remains "online" the entire time, user would be able to join the call and no error banner would be shown

JoshuaLai:
Not for this PR. But lets create an task to introduce UI tests for these scenarios

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/Setup/SetupViewModel.swift
```diff
@@ -12,6 +12,7 @@ class SetupViewModel: ObservableObject {
     private let localizationProvider: LocalizationProviderProtocol
 
     private var callingStatus: CallingStatus = .none
+    private var isOffline: Bool = false
```
JoshuaLai, line 15: Should we not rely on `NetworkStatus` here?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/Setup/SetupViewModel.swift
```diff
@@ -134,4 +140,13 @@ class SetupViewModel: ObservableObject {
         joinCallButtonViewModel.update(isDisabled: permissionState.audioPermission == .denied)
         errorInfoViewModel.update(errorState: state.errorState)
     }
+
+    private func showOfflineError() {
+        errorInfoViewModel.update(errorState: .init(internalError: .connectionFailed,
```
JoshuaLai, line 145: I don't think you want this method in here. Ideally once the network manager has determined the state is `offline` we can have our error reducer listen of the network action and create this error state for us. So you won't need to manually call show here.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/Setup/SetupViewModel.swift
jpeng-ms: hi josh, are you suggesting to use NetworkState rather than Bool here? If not, this variable is indeed relying on network state (at line 131)

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/Setup/SetupViewModel.swift
jpeng-ms: thanks josh, as I mentioned in this pr, this is exactly how I implemented originally. But PM explicit asks the error to be shown only when user tap on join call button, so I cannot reduce them to error state directly. Doing this would cause error to be shown as soon as composite launches.

JoshuaLai: But we should be able to start off the network status in `unknown` state, and when we tap on the joinCall button do the actual check for the network state, we would determine the network is actually offline and dispatch the error right?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/Setup/SetupViewModel.swift
JoshuaLai: Yes I think we rely on status instead of the boolean. Its a truer representation of the redux state

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/Setup/SetupViewModel.swift
jpeng-ms: that's not the case, In the network manager, we have set up the closure, it would be invoked as soon as network manager finishes initialization which would overwrite the network state to .online right away. The iOS decides when/how many network events it dispatches to us.

kevinyulu (approved):

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/Setup/SetupViewModel.swift
jpeng-ms: thanks will update it

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/Setup/SetupViewModel.swift
JoshuaLai: I see I think this still feel incorrect flow. Cause right now you are saying 
- Update error state
- force error to show
I actually don't think the error view model show should even be public method. 
We ideally want the state to determine whether or not the error banner should be displayed. Let me think to see if there is a better way to resolve this

PR #319, opened 2022-08-23
[Bugfix] Removed applying a global layout direction setting and added setting …

kevinyulu:
## Purpose

* Replaced global layout direction setting to setting it for the needed screens. We don't need to set it globally. Setting it globally will make the navigation items in demo view controller / demo ui view flip position unexpectedly after quitting call composite.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

Testing Navigation item in UIKitDemoViewController or SwiftUIDemoView may not be apparent. Try adding navigation items in these two screens and follow the test steps.

```
Run the app, choose Swift UI or UI Kit
In Settings, turn on is Right to left, 
Join call
Follow What to Check section below
```

## What to Check
* Make sure all the call composite screens follows layout direction settings. 
* Make sure Demo screens does not follow layout direction settings


AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/FluentUI/Wrapper/DrawerContainerViewController.swift
```diff
@@ -24,13 +24,16 @@ class DrawerContainerViewController<T>: UIViewController, DrawerControllerDelega
...
-         showHeader: Bool = false
+         showHeader: Bool = false,
+         isRightToLeft: Bool = false
     ) {
         self.items = items
         self.sourceView = sourceView
         self.showHeader = showHeader
         self.headerName = headerName
         super.init(nibName: nil, bundle: nil)
+        UIView.appearance().semanticContentAttribute = isRightToLeft ?
```
vhuseinova-msft, line 35: Should it be done for DrawerController or DrawerContainerViewController view instead?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/Container/ContainerUIHostingController.swift
```diff
@@ -32,7 +32,7 @@ class ContainerUIHostingController: UIHostingController<ContainerUIHostingContro
         self.callComposite = callComposite
         self.environmentProperties = environmentProperties
         super.init(rootView: environmentRoot)
-        UIView.appearance().semanticContentAttribute = isRightToLeft ?
+        self.view.semanticContentAttribute = isRightToLeft ?
```
jimchou-dev, line 35: Does this mean using view here instead of UIView will not cause issue to the app settings when returning to Contoso app?  And I haven't tested it for react native/Xamarin, but this also fixes the issue for cross platform correct?

jimchou-dev (commented):
Does this

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/Container/ContainerUIHostingController.swift
kevinyulu: yes, UIView.appearance().semanticContentAttribute sets the property globally and in this case we would want to set it just on this view controller and its views. I used a dummy navigation view in the demoView screen to and this change would not incur view layout switch on the demoView when semanticContentAttribute is changed.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/FluentUI/Wrapper/DrawerContainerViewController.swift
kevinyulu: I think we'd have the control of setting semanticContentAttribute if we put the logics in `DrawerContainerViewController` rather than `DrawerController`. `DrawerController`. I corrected the logics so the semanticContentAttribute will be updated on the `contentView` from this class. not at UIView as it will apply globally in the app.

vhuseinova-msft (approved):

Leah-Xia-Microsoft (approved):

jimchou-dev (approved):
LGTM

PR #322, opened 2022-08-25
[Bug] refactor code for xcode13.3.1 test failure

Leah-Xia-Microsoft:
## Purpose

* Passing only the error handler instead the whole self to accommodate Xcode13.3.1

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code and run tests on Xcode 13.3.1 and Xcode 13.4.x

## What to Check
* All tests pass

### DIFF
--- a/AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/UIKitDemoViewController.swift
+++ b/AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/UIKitDemoViewController.swift
@@ -170,8 +170,12 @@ class UIKitDemoViewController: UIViewController {
         }
 
         callComposite.events.onError = { [weak self] error in
-            Task { @MainActor in self?.onError(error) }
+            guard let errorHandler = self?.onError else {
+                return
+            }
+            Task { @MainActor in errorHandler(error) }
         }
+
         callComposite.events.onRemoteParticipantJoined = onRemoteParticipantJoinedHandler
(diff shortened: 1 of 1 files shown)


AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/UIKitDemoViewController.swift
```diff
@@ -170,8 +170,12 @@ class UIKitDemoViewController: UIViewController {
         }
 
         callComposite.events.onError = { [weak self] error in
-            Task { @MainActor in self?.onError(error) }
```
jpeng-ms, line 173: is this issue only happening in ui kit? do we need to update swiftuidemoview as well?

jimchou-dev (approved):

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/UIKitDemoViewController.swift
Leah-Xia-Microsoft: > is this issue only happening in ui kit? do we need to update swiftuidemoview as well?

The SwiftUI counterpart is a struct instead of a class so it doesn't have this problem.

jpeng-ms (approved):

PR #323, opened 2022-08-25
[Bug] End Ongoing Call if User Force Quits

jpeng-ms:
## Purpose
The goal of this PR is to end the ongoing call if user closes the app unexpectedly.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
0. launch a call on your side
1. run the demo app in simulator and join the call
2. swipe from the bottom to go to app switch view
3. swipe up to close demo app
4. check to see if demo app participant has been removed from the app

## What to Check
1. when demo app closes, its caller should be removed from the call

## Other Information
1. the end call has a delay of up to few seconds (may takes few seconds for participant to be removed). This is consistent with ending the call explicitly using the end call button as well as Teams' iOS app.
2. if app got terminated from Xcode, then lifecycle event won't be sent to our app to initiate end call request. Participant would still remain in the call


Leah-Xia-Microsoft (dismissed):
LGTM. Tested on iPhone 13 with simulator

Leah-Xia-Microsoft (dismissed):

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/Manager/AppLifeCycleManager.swift
```diff
@@ -41,4 +56,12 @@ class UIKitAppLifeCycleManager: LifeCycleManagerProtocol {
         logger.debug("Did Activate")
         store.dispatch(action: .lifecycleAction(.foregroundEntered))
     }
+
+    @objc func willTerminate(_ notification: Notification) {
+        logger.debug("Will Terminate")
+        if callingStatus == .connected {
+            store.dispatch(action: .lifecycleAction(.willTerminate))
+            CFRunLoopRun()
```
AmyL219, line 64: Why not having 
```
if callingStatus == .disconnected {
            CFRunLoopStop(CFRunLoopGetCurrent())
```

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Service/Calling/CallingSDKWrapper.swift
```diff
@@ -90,13 +90,21 @@ class CallingSDKWrapper: NSObject, CallingSDKWrapperProtocol {
...
+        // executing the following code in main thread because
+        // we need to break correct runloop (the one on main thread)
+        // otherwise, when user force close the app, the app would
+        // hang (runloop blocking main thread) and
+        // gets terminated by iOS with exit code 9
+        try await Task { @MainActor in
+            do {
+                try await call?.hangUp(options: HangUpOptions())
+                logger.debug("Call ended successfully")
+                CFRunLoopStop(CFRunLoopGetCurrent())
```
AmyL219, line 102: Could we move the `CFRunLoopStop(CFRunLoopGetCurrent())` to `UIKitAppLifeCycleManager`?
I think it makes more sense to decouple the `endCall` function with `CFRunLoop`.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/Manager/AppLifeCycleManager.swift
jpeng-ms: that's because we need to know when end call returns to break the runloop. Putting runloop here would create a deadlock and the loop would never be terminated

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Service/Calling/CallingSDKWrapper.swift
jpeng-ms: No we can't, we have to put it here to wait for end call to finish execution. The code in cycle manager is async and independent to end call flow.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/Manager/AppLifeCycleManager.swift
AmyL219: For point 1, when callingSDK signals `.disconnected`, the user would be removed from the group call successfully, remote participant would see this user is removed from the group call. The call composite actually rely on `.disconnected` (the single source of truth) to end the composite lifecycle, rather than the endCall callback return function. 
I think it's better to keep consistent in using `disconnected` as the single source of ending the whole composite

jpeng-ms: Hi Amy, hopefully the SDK team has cleared this out. If so, please resolve this conversation.

jpeng-ms:
closing this PR since Amy has taken this over and she had some pending discussions on further refactoring calling states

PR #327, opened 2022-08-30
[Feature] Add Alert When Network Lost

jpeng-ms:
## Purpose

* ...

```
[ ] Yes
[ ] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...


AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/Setup/SetupViewModel.swift
```diff
@@ -94,6 +98,10 @@ class SetupViewModel: ObservableObject {
         }.store(in: &cancellables)
     }
 
+    deinit {
+        networkManager.stopMonitor()
```
jpeng-ms, line 102: To address interpal's concerns on performance impact of having extra thread for network monitoring, we cancel the monitoring once set up view is dismissed.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/Setup/SetupViewModel.swift
```diff
@@ -134,4 +146,9 @@ class SetupViewModel: ObservableObject {
         joinCallButtonViewModel.update(isDisabled: permissionState.audioPermission == .denied)
         errorInfoViewModel.update(errorState: state.errorState)
     }
+
+    private func showOfflineError() {
+        store.dispatch(action: .errorAction(.statusErrorAndCallReset(internalError: .connectionFailed,
```
jpeng-ms, line 151: redux is still required to clear out error state so that correct error can be shown. If we don't update error state, then incorrect error would be shown when user swipe down Notification Center or change audio/video settings or enter background (any state changes).

for example, when SDK returns an error, error state registers it. if we don't update error state, when there's a state change, error banner would change back to SDK error rather than network error.

PR #339, opened 2022-09-07
[Feature] Chat Composite project skeleton structure

jimchou-dev:
## Purpose

* Added folder/project structure for chat composite
* Added entry point for starting chat composite and reading from environment variable
* Added stub for public API for chat composite

```
[ ] Yes
[ x ] No
```

## Pull Request Type

```
[ ] Bugfix
[ x ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...


AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/CredentialsView.swift
```diff
@@ -0,0 +1,280 @@
...
+    @State var unreadMessageCount: Int = 0
+
+    @ObservedObject var envConfigSubject: EnvConfigSubject
+
+    let verticalPadding: CGFloat = 5
+    let horizontalPadding: CGFloat = 10
+
+    let viewType: DemoViewType
+
+//    let callWithChatComposite = CallWithChatComposite(withOptions: CallWithChatCompositeOptions())
```
palatter, line 33: Commented code

AzureCommunicationUI/Podfile
```diff
@@ -48,5 +66,8 @@ post_install do |installer|
         config.build_settings['IPHONEOS_DEPLOYMENT_TARGET'] = '14.0'
       end
     end
+    target.build_configurations.each do |config|
```
palatter, line 69: Is the reason we need this now that we have added chat?

jimchou-dev: This is because the SDK don't support bitcode so the dependencies need to set bitcode as NO

JoshuaLai:
@jimchou-dev as part of initiative to rework some of the readme. I think it makes sense to create a new readme.md in the `sdk/AzureCommunicationUIChat` folder. I think it can be just the stub rea me for now and we can fill it in as the new public apis  are ready.

AzureCommunicationUI/Podfile
palatter: We already had calling sdk not supporting it though?

jimchou-dev: Correct, but when doing `pod install` the dependency may have bitcode enabled. This is to ensure the dependencies will also have bitcode disabled in order to run the demo app on physical device.

AzureCommunicationUI/sdk/common/BaseComposite/Sources/Logger/Logger.swift
```diff
@@ -0,0 +1,27 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import Foundation
+import os
+
+protocol Logger {
```
jpeng-ms, line 9: can we leverage protocol extension to provide a default implementation rather than having a different class for it

jimchou-dev: We can do that but there are reasons not to provided in this medium article
https://betterprogramming.pub/swift-why-you-should-avoid-using-default-implementations-in-protocols-eeffddbed46d

PR #342, opened 2022-09-09
[Refactor] Updated Remote Manager to Protocol

jpeng-ms:
## Purpose
This refactor is meant to address the inconsistency we had when creating view managers. Ideally, all of view managers should be protocol rather than its concrete class

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[X] Refactoring (no functional changes, no api changes)
[ ] Other... Please describe:
```

## How to Test
1. run the app
2. open settings to turn on remote participant injection
3. join a call
## What to Check
1. avatar does show up
2. no crash upon launching composite
3. no crash upon joining the call

## Other Information
1. AvatarViewManager remains as is since we have additional refactor coming up for it. So this PR only addresses remote participant view manager.

### DIFF
--- a/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallComposite.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallComposite.swift
@@ -100,7 +100,7 @@ public class CallComposite {
         self.permissionManager = dependencyContainer.resolve() as PermissionsManagerProtocol
         self.audioSessionManager = dependencyContainer.resolve() as AudioSessionManagerProtocol
         self.avatarViewManager = dependencyContainer.resolve() as AvatarViewManager
-        self.remoteParticipantsManager = dependencyContainer.resolve() as RemoteParticipantsManager
+        self.remoteParticipantsManager = dependencyContainer.resolve() as RemoteParticipantsManagerProtocol
     }
 
     private func cleanUpManagers() {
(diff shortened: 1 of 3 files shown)


AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallComposite.swift
```diff
@@ -100,7 +100,7 @@ public class CallComposite {
         self.permissionManager = dependencyContainer.resolve() as PermissionsManagerProtocol
         self.audioSessionManager = dependencyContainer.resolve() as AudioSessionManagerProtocol
         self.avatarViewManager = dependencyContainer.resolve() as AvatarViewManager
```
palatter, line 102: Does the avatarViewManager have the same issue?

palatter (approved):
Looks good!

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallComposite.swift
jpeng-ms: good question! we actually have a separate refactor ticket for AvatarViewManager since it's an observable object and we can't just replace it with protocol directly. I believe @AmyL219 had created a separate ticket for this task

AmyL219 (approved):

PR #353, opened 2022-09-20
[Bugfix] Join call button accessiblity text sizing fix

palatter:
## Purpose

* Resolve issue of join call button layout not fitting on the screen with very large text sizes. This is resolved in a similar way to our icon label buttons.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...

### DIFF
--- a/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/FluentUI/Wrapper/CompositeButton.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/FluentUI/Wrapper/CompositeButton.swift
@@ -33,9 +33,9 @@ struct CompositeButton: UIViewRepresentable {
     func updateUIView(_ button: FluentUI.Button, context: Context) {
         let sizeCategory = UIApplication.shared.preferredContentSizeCategory
         if sizeCategory >= UIContentSizeCategory.accessibilityMedium {
-            button.titleLabel!.font = Fonts.button1Accessibility
+            button.titleLabel?.font = Fonts.button1Accessibility
         } else {
-            button.titleLabel!.font = Fonts.button1
+            button.titleLabel?.font = Fonts.button1
         }
     }
 }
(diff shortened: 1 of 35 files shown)


AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/FluentUI/Wrapper/CompositeButton.swift
```diff
@@ -30,6 +30,12 @@ struct CompositeButton: UIViewRepresentable {
         return button
     }
 
-    func updateUIView(_ uiView: FluentUI.Button, context: Context) {
+    func updateUIView(_ button: FluentUI.Button, context: Context) {
+        let sizeCategory = UIApplication.shared.preferredContentSizeCategory
+        if sizeCategory >= UIContentSizeCategory.accessibilityMedium {
+            button.titleLabel!.font = Fonts.button1Accessibility
```
jpeng-ms, line 36: Hi, I was wondering if it makes sense if we define min scale factor rather than using a fixed font size here?

palatter: I gave this a try but because this is inside the button and wrapped with UI representable it didn't scale the text correctly.

jpeng-ms (dismissed):

JoshuaLai (dismissed):
LGTM

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/FluentUI/Wrapper/CompositeButton.swift
emlynmac, line 36: Can this be an optional unwrap? Would prevent a crash in the case that for some reason it is nil

palatter: Oh good catch, I'll fix that!

jpeng-ms (approved):

kevinyulu (approved):

JoshuaLai (approved):

PR #359, opened 2022-09-28
[Bug] Fixed Drawer Being Cutoff

jpeng-ms:
## Purpose
The drawer resize function being called when table view not ready which results the incorrect height being defined for drawer and causing the content being cutoff.

The solution is to check if table view content is ready. If it's not ready, we would not call resize() function to prevent incorrect  drawer height being set.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...

### before

https://user-images.githubusercontent.com/109105353/193646243-06cb32bd-9791-4652-a26e-df6457389235.mov

### after

https://user-images.githubusercontent.com/109105353/193640285-ec6858cc-bf9e-425f-80f4-b8281e59685d.mov

### before
Screen Shot 2022-10-03 at 11 33 18 AM

### after
image

## Testing

https://user-images.githubusercontent.com/109105353/193653494-ac83d6a4-ba6e-4f7b-b997-7c1ad3fca12b.mov


Leah-Xia-Microsoft:
Hi John, I'm testing this branch and the first thing noticed is that the drawer is not high enough to show the participants (only me in this case)
IMG_0061

PR #362, opened 2022-09-28
[Refactor] Generify the Calling SDK wrapper protocol

emlynmac:
## Purpose
The calling SDK wrapper protocol interface was designed to separate out the underlying layers.
This PR removes the interface dependency on the azure calling SDK, needing only the concrete implementations to include those objects where needed.

This is the first part of enabling automated end-to-end testing.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code
Pull and build

* Test the code
Run all unit / UI tests
Perform manual testing

## What to Check
No behavioural changes


AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/VideoViewManager.swift
```diff
@@ -73,18 +73,21 @@ class VideoViewManager: NSObject, RendererDelegate, RendererViewManager {
             return localRenderCache.rendererView
         }
 
-        guard let videoStream = callingSDKWrapper.getLocalVideoStream(videoStreamId) else {
+        guard let videoStream: LocalVideoStream<AzureCommunicationCalling.LocalVideoStream> =
+                callingSDKWrapper.getLocalVideoStream(videoStreamId) else {
             return nil
         }
-
+        let nativeVideoStream = videoStream.wrappedObject
```
AmyL219, line 80: I am pretty confused by the name `wrappedObject`. What's the object type is it, how does it wrapped?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/Manager/RemoteParticipantsManager.swift
```diff
@@ -70,8 +70,13 @@ class RemoteParticipantsManager: RemoteParticipantsManagerProtocol {
             return
         }
 
-        let joinedParticipantsCommunicationIds = joinedParticipantsIds
-            .compactMap { callingSDKWrapper.getRemoteParticipant($0)?.identifier }
+        let joinedParticipantsCommunicationIds: [CommunicationIdentifier] = joinedParticipantsIds
+            .compactMap {
+                let participant: RemoteParticipant< RemoteParticipantIdentifable,
+                                                   AnyObject>? =
```
AmyL219, line 76: Why do we want to allow `AnyObject` here? Shouldn't we narrow down the type of it which have better readness and performance?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Service/Calling/AzureCommunicationCalling/RemoteParticipantExtension.swift
```diff
@@ -27,9 +27,38 @@ extension RemoteParticipant {
                                     screenShareVideoStreamModel: screenShareVideoStreamModel,
                                     cameraVideoStreamModel: cameraVideoStreamModel)
     }
+
+    static func toUiRemoteParticipant(
```
AmyL219, line 31: Does the prefix `Ui`  consistent with any name we have? We usually use `composite`

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Service/Calling/AzureCommunicationCalling/RemoteParticipantExtension.swift
```diff
@@ -40,4 +69,15 @@ extension MediaStreamType {
             return VideoStreamInfoModel.MediaStreamType.cameraVideo
         }
     }
+
+    var asUiMediaStreamType: MediaStreamType {
```
AmyL219, line 73: same concern about prefix `Ui`, ,if we should update to `composite`

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Service/Calling/CallingSDKWrapperProtocol.swift
```diff
@@ -2,19 +2,64 @@
...
 enum CameraDevice {
     case front
     case back
 }
 
+protocol RemoteParticipantIdentifable {
+    var identifier: CommunicationIdentifier { get }
+}
+
+class RemoteParticipant<WrappedType, VideoStreamType>: RemoteParticipantIdentifable {
```
AmyL219, line 18: My main concern is we don't define the type of `WrappedType` and `VideoStreamType`, they can be `AnyObject`.  If any developer pass a `UITableView` as the `WrappedType` into this RemoteParticipant self defined wrapper, is it a valid RemoteParticipantWrapper anymore?
Any why don't we add a suffix to this wrapper rather than make this wrapper duplicated with the CallingSDK.RemoteParticipant?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Service/Calling/CallingSDKWrapperProtocol.swift
```diff
@@ -2,19 +2,64 @@
...
+        self.wrappedObject = wrappedObject
+    }
+}
+
+enum MediaStreamType {
+    case cameraVideo
+    case screenSharing
+}
+
+class RemoteVideoStream<WrappedType> {
```
AmyL219, line 37: Same concern for the WrappedType. If we want to use the generic type, we'd better have a constraints for those generic types

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/VideoViewManager.swift
emlynmac: It is a wrappedObject, in order to remove include dependency. The type depends on how you use it - have a look at the generics on the method signature.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/Manager/RemoteParticipantsManager.swift
emlynmac: We want to ensure we allow any object to be resolved here. We don't care about the second generic as it is not used in this context. If we restrict it, then the participant will be `nil` where the generics don't match, which is _not_ what we want.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Service/Calling/CallingSDKWrapperProtocol.swift
emlynmac: The whole point of the wrapped type generic _is_ that any implementor can choose their object that they pass through. Maybe we need a chat offline here?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Service/Calling/CallingSDKWrapperProtocol.swift
emlynmac: For now, this simply ensures we don't force calling SDK objects to the interface.
In the future, we might make this public and have expectations about the objects being passed through, i.e. constraining it.  We simply don't have that understanding at this point in time.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Service/Calling/AzureCommunicationCalling/RemoteParticipantExtension.swift
emlynmac: I will update all the instances of the composite objects accordingly.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Service/Calling/CallingSDKWrapperProtocol.swift
```diff
@@ -2,19 +2,64 @@
...
+}
+
+class RemoteParticipant<WrappedType, VideoStreamType>: RemoteParticipantIdentifable {
+    var identifier: CommunicationIdentifier
+    var videoStreams: [RemoteVideoStream<VideoStreamType>]
+    var wrappedObject: WrappedType
+
+    init(id: CommunicationIdentifier,
+         videoStreams: [RemoteVideoStream<VideoStreamType>],
+         wrappedObject: WrappedType) {
```
JoshuaLai, line 25: I think this `wrappedObject` name can be changed to have a little more context to what it is under the hood, like `rawRemoteParticipant` for example

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Service/Calling/CallingSDKWrapperProtocol.swift
```diff
@@ -2,19 +2,65 @@
...
+
+enum CompositeMediaStreamType {
+    case cameraVideo
+    case screenSharing
+}
+
+class CompositeRemoteVideoStream<WrappedType> {
+    var id: Int
+    var mediaStreamType: CompositeMediaStreamType = .cameraVideo
+    var wrappedObject: WrappedType
```
JoshuaLai, line 40: Same comment here

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Service/Calling/CallingSDKWrapperProtocol.swift
emlynmac: OK, I can do that.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Service/Calling/CallingSDKWrapperProtocol.swift
emlynmac: OK. I mean it is completely generic at this point, until the WrappedType is constrained to something.

JoshuaLai (dismissed):
Chatted offline with Emlyn to understand the motivation behind the PR. 
Everything makes sense to me and sounds like a move in the right direction.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Service/Calling/AzureCommunicationCalling/LocalVideoStreamExtension.swift
```diff
@@ -0,0 +1,19 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import AzureCommunicationCalling
+import Foundation
+
+extension CompositeLocalVideoStream {
+    static func nativeAccMediaStreamType(type: CompositeMediaStreamType)
```
kevinyulu, line 10: is it nativeAcsMediaStreamType?

kevinyulu (dismissed):

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Service/Calling/CallingSDKWrapperProtocol.swift
AmyL219: I think we should keep consistent with Android in naming those wrapper class with suffix `wrapper`. And also, limit the type of each variables inside those wrapper class 
https://github.com/Azure/communication-ui-library-android/tree/develop/azure-communication-ui/calling/src/main/java/com/azure/android/communication/ui/calling/service/sdk

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/VideoViewManager.swift
emlynmac: At this point in time, yes. It is enabling object passthrough.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/Manager/RemoteParticipantsManager.swift
emlynmac: Because we want this to resolve regardless of the second type. This is a lookup, based on the RemoteParticipantIdentifiable interface.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/VideoViewManager.swift
```diff
@@ -112,16 +115,19 @@ class VideoViewManager: NSObject, RendererDelegate, RendererViewManager {
...
                   return String(stream.id) == videoStreamId
               }) else {
             return nil
         }
 
+        let nativeStream = videoStream.wrappedObject
         do {
             let options = CreateViewOptions(scalingMode: videoStream.mediaStreamType == .screenSharing ? .fit : .crop)
-            let newRenderer: VideoStreamRenderer = try VideoStreamRenderer(remoteVideoStream: videoStream)
+            let newRenderer: VideoStreamRenderer = try VideoStreamRenderer(remoteVideoStream: nativeStream)
```
jpeng-ms, line 130: sorry didn't mean to be nitpicking here, but should we name it differently since it's not a native stream (aka. AVCaptureVideoDataOutput) per sec? I am okay with this if you think this won't cause confusion later on when SDK exposes native stream to us

jpeng-ms (approved):

jimchou-dev (approved):
LGTM

kevinyulu (approved):

PR #363, opened 2022-09-29
[Bug] Fix Rotation does not work until drawer opens on iOS 16

Leah-Xia-Microsoft:
## Purpose

* For iOS 16, we need to update the UIViewController that the device has rotated or else it will cause rotation issue such as:
 - Rotation does not work until drawer opens
 - After rotating to landscape mode, camera is forced back to portrait mode after 1 second

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get and run the code
* Start a video call
* Rotate devices

## What to Check
* Rotation works as expected

## Other Information

We are currently doing this step of `attemptRotationToDeviceOrientation` when dismiss drawers:
```
    override func viewDidDisappear(_ animated: Bool) {
        super.viewDidDisappear(animated)
        if isBeingDismissed || isMovingFromParent {
            sourceView.superview?.isUserInteractionEnabled = true
            sourceView.removeFromSuperview()
        }
        if UIDevice.current.userInterfaceIdiom == .phone {
            UIDevice.current.setValue(UIDevice.current.orientation.rawValue, forKey: "orientation")
            UIViewController.attemptRotationToDeviceOrientation()
        }
    }
```
This PR just migrate it to `CallingView` as well.


AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/Calling/CallingView.swift
```diff
@@ -210,6 +210,9 @@ extension CallingView {
             && (areAllOrientationsSupported || (!areAllOrientationsSupported
                                                 && newOrientation != .portraitUpsideDown)) {
             orientation = newOrientation
+            if UIDevice.current.userInterfaceIdiom == .phone {
```
vhuseinova-msft, line 213: Should the same be done for iPad?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/Calling/CallingView.swift
```diff
@@ -210,6 +210,9 @@ extension CallingView {
...
+                UIViewController.attemptRotationToDeviceOrientation()
```
vhuseinova-msft, line 214: this API is deprecated in iOS 16. Maybe for iOS 16 setNeedsUpdateOfSupportedInterfaceOrientations should be used as the issues were introduced in iOS 16

AmyL219: https://developer.apple.com/documentation/uikit/uiviewcontroller/1621400-attemptrotationtodeviceorientati
attemptRotationToDeviceOrientation is trying to attemptRotationToDeviceOrientation?
Could this fix the bug?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/Calling/CallingView.swift
Leah-Xia-Microsoft: @vhuseinova-msft I've tested on iPad simulator. These iOS 16 issue doesn't affect iPad. Would be nice if @jimchou-dev can test and confirm on iPad real device

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/Calling/CallingView.swift
Leah-Xia-Microsoft: @AmyL219 Do you mean there's some concern about attemptRotationToDeviceOrientation rotate the device immediately? Both John and I tested on iOS 16 devices ant it fixed the cannot rotate until tap on drawer issue.

emlynmac: The replacement API is iOS 16 and above only, so we'd need to have both code paths.
It might be better for now to keep the fix like this, and then remove the deprecated API use either if we have further issues or when iOS 16 becomes the minimum.

emlynmac (dismissed):

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/Calling/CallingView.swift
jpeng-ms: agree, maybe we can start to adopt the new API now or either create a refactor task yo track this

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/Calling/CallingView.swift
jimchou-dev: iPadOS does not have have iOS16 yet, until later in October or later this year I think

emlynmac: There is a beta available for anyone with an iPad to test.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/Calling/CallingView.swift
Leah-Xia-Microsoft: @vhuseinova-msft Since`attemptRotationToDeviceOrientation` is also used in a few other places as well that's not only for iOS 16, and switching to `setNeedsUpdateOfSupportedInterfaceOrientations ` is not a simple change. I tend to agree to create a refactor task when 16 becomes minimum. What do you think?

vhuseinova-msft: Sounds good!

vhuseinova-msft (approved):

jpeng-ms (approved):

PR #366, opened 2022-09-30
[Bug] Error Banner Margin Fix

jpeng-ms:
## Purpose
UI spec for the error banner shows that it should have equal width as the join call button: 
image
but in our app, we have a small margin which makes the banner appeared to be shorter than join call button:
image

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
1. disconnect the internet
2. tap on join call

## What to Check
1. make sure button has the equal width as the join call button under both dark and light mode

## Other Information
image


emlynmac (approved):
LGTM

palatter (approved):
Looks good and tested.

kevinyulu (approved):

PR #367, opened 2022-09-30
[Bug] Fix drawer orientation changes when device is on a flat surface

Leah-Xia-Microsoft:
## Purpose

* Currently when we join a call and go to landscape mode, then put device on a flat surface and open any drawer, the device rotate to a different direction, for example, from landscapeLeft to LandscapeRight.
* This PR fix that issue by catching the scenario that screenSizeClass is .iphoneLandscapeScreenSize but UIDevice.current.orientation is not landscape, which could be faceUp or faceDown.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get and run the code
* Join a call and go to landscape mode
* put device on a flat surface
* open drawer by tapping on end call button

## What to Check
* Device orientation should not change

### DIFF
--- a/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/ViewComponents/LockPhoneOrientation.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/ViewComponents/LockPhoneOrientation.swift
@@ -18,7 +18,11 @@ struct LockPhoneOrientation: ViewModifier {
         case .iphonePortraitScreenSize:
             return .portrait
         case .iphoneLandscapeScreenSize:
-            return UIDevice.current.orientation == .landscapeLeft ? .landscapeRight : .landscapeLeft
+            if UIDevice.current.orientation.isLandscape {
+                return UIDevice.current.orientation == .landscapeLeft ? .landscapeRight : .landscapeLeft
+            } else {
+                return .landscape
+            }
         default:
             return SupportedOrientationsPreferenceKey.defaultValue
         }


AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/ViewComponents/LockPhoneOrientation.swift
```diff
@@ -18,7 +18,11 @@ struct LockPhoneOrientation: ViewModifier {
         case .iphonePortraitScreenSize:
             return .portrait
         case .iphoneLandscapeScreenSize:
-            return UIDevice.current.orientation == .landscapeLeft ? .landscapeRight : .landscapeLeft
+            if UIDevice.current.orientation.isLandscape {
+                return UIDevice.current.orientation == .landscapeLeft ? .landscapeRight : .landscapeLeft
```
Leah-Xia-Microsoft, line 22: This line was for a fix in this PR (https://github.com/Azure/communication-ui-library-ios/pull/31/files) January this year. It's a bit counter-intuitive but I've consulted Kevin and revert this line introduce side effects.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/ViewComponents/LockPhoneOrientation.swift
```diff
@@ -18,7 +18,11 @@ struct LockPhoneOrientation: ViewModifier {
...
+            } else {
+                return .landscape
```
jpeng-ms, line 24: hi I am bit of confused by this, if orientation.isLandscape return .landscapeLeft or .landscapeRight otherwise return .landscape

can you explain what's going on here?

kevinyulu: The new lines takes the assumption that the `iphoneLandscapeScreenSize` is true in this switch case. In this switch case we'd already know that the user is on an interface that has size resembles iPhone landscape size. Thus we'd only return landscape orientation accordingly. For an use case that user has switch to landscape and put their phone flat, the `iphoneLandscapeScreenSize` would still be true in this case and we'd mask the orientation to .landscape by default.

kevinyulu (approved):

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/ViewComponents/LockPhoneOrientation.swift
Leah-Xia-Microsoft: @jpeng-ms 
* For `return UIDevice.current.orientation == .landscapeLeft ? .landscapeRight : .landscapeLeft` part:
  - This line was for a fix in this PR (https://github.com/Azure/communication-ui-library-ios/pull/31/files) January this year. It's a bit counter-intuitive but I've consulted Kevin and revert this line introduce side effects.
* For `return .landscape` part:
  - We need to look at the whole function:
  ```
    var orientationMask: UIInterfaceOrientationMask {
        switch screenSizeClass {
        case .iphonePortraitScreenSize:
            return .portrait
        case .iphoneLandscapeScreenSize:
            if UIDevice.current.orientation.isLandscape {
                return UIDevice.current.orientation == .landscapeLeft ? .landscapeRight : .landscapeLeft
            } else {
                return .landscape
            }
        default:
            return SupportedOrientationsPreferenceKey.defaultValue
        }
    }
```
  - This new if/else statement is under case of .iphoneLandscapeScreenSize. So if the screenSizeClass is already in landscapeScreenSize, then we check whether `UIDevice.current.orientation.isLandscape` is landscape or not. If it's not landscape in this case, we assume it's .faceUp or .faceDown which is cause the device on flat surface problem.

jpeng-ms (approved):

emlynmac (approved):

PR #382, opened 2022-10-05
[Bug] Set Speaker Audio Device as the default option

Leah-Xia-Microsoft:
## Purpose

* Set Speaker Audio Device as the default option

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Fresh install the app
*  Tap on `Start Experience` 
*  Observe whether the default audio device is Speaker
*  Connect to a bluetooth or headphone, then disconnect it
*  Observe whether it get back to Speaker instead of receiver
*  Tap on `Join call`
*  Observe whether the default audio device is Speaker
*  Connect to a bluetooth or headphone, then disconnect it
*  Observe whether it get back to Speaker instead of receiver

### DIFF
--- a/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/Manager/AudioSessionManager.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/Manager/AudioSessionManager.swift
@@ -123,7 +123,7 @@ class AudioSessionManager: AudioSessionManagerProtocol {
             case .headphones, .headsetMic:
                 return .headphones
             case .builtInSpeaker:
-                return .speaker
+                return .receiver
             default:
                 return .speaker
             }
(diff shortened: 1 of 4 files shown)


emlynmac (dismissed):
LGTM; just one question about doing this; is there a risk of feedback at call start with speaker / microphone active?

Leah-Xia-Microsoft:
> LGTM; just one question about doing this; is there a risk of feedback at call start with speaker / microphone active?

Good point. I've tested on 2 physical devices joining the call at the same time with speaker / microphone active. Seems not causing feedback.

emlynmac (approved):

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/Manager/AudioSessionManager.swift
```diff
@@ -123,12 +123,12 @@ class AudioSessionManager: AudioSessionManagerProtocol {
             case .headphones, .headsetMic:
                 return .headphones
             case .builtInSpeaker:
```
kevinyulu, line 125: should this be changed to .buildInReceiver?

Leah-Xia-Microsoft: Good catch. It also make me doubt about the scenario that after disconnect to bluetooth, do we want to fall back to receiver or speaker. I'll have to confirm with Alex on this when he's back to the office. Closing this PR for now. Thanks.

PR #384, opened 2022-10-06
[Bug] Drawer Cutoff and Bouncing Animation Fix

jpeng-ms:
## Purpose
The drawer resize function being called when table view not ready which results the incorrect height being defined for drawer and causing the content being cutoff.

The solution is to check if table view content is ready. If it's not ready, we would not call resize() function to prevent incorrect  drawer height being set.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
for bouncing animation:
1. enable slow animation
2. tap on audio list
3. make a selection
4. audio should be dismissed without janky animations
5. test the same thing for end call as well

for cutoff:
1. enable large font size in a11y
2. leave display name empty
3. join a call
4. tap on participant list button repeatedly until cutoff occurs (since it's a race condition, you might need to try multiple times for it to happen)

## What to Check
1. no bouncing animations when selections are made
2. no text being cutoff
3. participant list gets dynamically updated
4. drawer still work as before on both iPad and iPhone (note: on iPads, there's no animation when presenting the drawer, this behaviour is consistent with 1.0.0 GA release.)
5. no blank drawer shown up under any conditions

### before

https://user-images.githubusercontent.com/109105353/193646243-06cb32bd-9791-4652-a26e-df6457389235.mov

### after

https://user-images.githubusercontent.com/109105353/193640285-ec6858cc-bf9e-425f-80f4-b8281e59685d.mov

### before

https://user-images.githubusercontent.com/109105353/194431133-c07d5b28-79e1-4750-8a66-83bb7c680282.mov

### after

https://user-images.githubusercontent.com/109105353/194431043-d3571eb4-0d7f-4a48-bc66-7ccfb5946c65.mov

## Testing

https://user-images.githubusercontent.com/109105353/193653494-ac83d6a4-ba6e-4f7b-b997-7c1ad3fca12b.mov

### DIFF
--- a/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/FluentUI/Wrapper/DrawerContainerViewController.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/FluentUI/Wrapper/DrawerContainerViewController.swift
@@ -17,18 +17,22 @@ class DrawerContainerViewController<T: Equatable>: UIViewController, DrawerContr
     private let sourceView: UIView
     private let showHeader: Bool
     private let isRightToLeft: Bool
+    private weak var controller: DrawerController?
+
+    // MARK: Constants
     private var halfScreenHeight: CGFloat {
         UIScreen.main.bounds.height / 2
     }
-    private weak var controller: DrawerController?
+    private var resizeBarHeight: CGFloat {
+        UIDevice.current.userInterfaceIdiom == .phone ? 20 : 0
+    }
+    private let drawerWidth: CGFloat = 400.0
 
-    init(items: [T],
-         sourceView: UIView,
+    init(sourceView: UIView,
          headerName: String? = nil,
          showHeader: Bool = false,
          isRightToLeft: Bool = false
     ) {
-        self.items = items
         self.sourceView = sourceView
         self.showHeader = showHeader
         self.headerName = headerName
@@ -144,7 +148,7 @@ class DrawerContainerViewController<T: Equatable>: UIViewController, DrawerContr
                 isScrollEnabled = true
             }
             drawerTableView.isScrollEnabled = isScrollEnabled
-            self.controller?.preferredContentSize = CGSize(width: 400,
+            self.controller?.preferredContentSize = CGSize(width: self.drawerWidth,
                                                            height: drawerHeight)
         }
     }
@@ -154,7 +158,6 @@ class DrawerContainerViewController<T: Equatable>: UIViewController, DrawerContr
                                  showHeader: Bool,
                                  isiPhoneLayout: Bool) -> CGFloat {
         let headerHeight = self.getHeaderHeight(tableView: tableView, isiPhoneLayout: isiPhoneLayout)
-        let resizeBarHeight: CGFloat = isiPhoneLayout ? 20 : 0
         let dividerOffsetHeight = CGFloat(numberOfItems * 3)
 
         var drawerHeight: CGFloat = getTotalCellsHeight(tableView: tableView, numberOfItems: numberOfItems)
@@ -182,9 +185,8 @@ class DrawerContainerViewController<T: Equatable>: UIViewController, DrawerContr
         guard let tableView = self.drawerTableView else {
             return
         }
-        let resizeBarHeight: CGFloat = isiPhoneLayout ? 20 : 0
-        let width = UIScreen.main.bounds.width
-        tableView.frame = CGRect(x: 0.0, y: resizeBarHeight, width: width, height: 0.0)
+        let initialWidth = isiPhoneLayout ? UIScreen.main.bounds.width : drawerWidth
+        tableView.frame = CGRect(x: 0, y: resizeBarHeight, width: initialWidth, height: 0)
         tableView.setNeedsDisplay()
     }
 }
--- a/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/FluentUI/Wrapper/DrawerContainerViewController.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/FluentUI/Wrapper/DrawerContainerViewController.swift
@@ -6,7 +6,7 @@
 import FluentUI
 import UIKit
 
-class DrawerContainerViewController<T>: UIViewController, DrawerControllerDelegate {
+class DrawerContainerViewController<T: Equatable>: UIViewController, DrawerControllerDelegate {
     weak var delegate: DrawerControllerDelegate?
     lazy var drawerTableView: UITableView? = nil
     let backgroundColor: UIColor = UIDevice.current.userInterfaceIdiom == .pad
@@ -67,6 +67,9 @@ class DrawerContainerViewController<T>: UIViewController, DrawerControllerDelega
     }
 
     func updateDrawerList(items: [T]) {
+        guard self.items != items else {
+            return
+        }
         guard self.items.count != items.count else {
             self.items = items
             self.drawerTableView?.reloadData()
--- a/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/FluentUI/Wrapper/DrawerContainerViewController.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/FluentUI/Wrapper/DrawerContainerViewController.swift
@@ -20,14 +20,18 @@ class DrawerContainerViewController<T: Equatable>: UIViewController, DrawerContr
     private weak var controller: DrawerController?
 
     // MARK: Constants
-    private var halfScreenHeight: CGFloat {
-        UIScreen.main.bounds.height / 2
+    private enum Constants {
+        static var drawerWidth: CGFloat { return 400.0 }
+        static var resizeBarHeight: CGFloat {
+            UIDevice.current.userInterfaceIdiom == .phone ? 20 : 0
+        }
+        static var halfScreenHeight: CGFloat {
+            UIScreen.main.bounds.height / 2
+        }
+        static var drawerHeaderMargin: CGFloat {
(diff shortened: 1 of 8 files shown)


AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/FluentUI/Wrapper/DrawerContainerViewController.swift
```diff
@@ -162,4 +177,14 @@ class DrawerContainerViewController<T>: UIViewController, DrawerControllerDelega
             }
         }.map { index in return tableView.rectForRow(at: index).height }.reduce(0, +)
     }
+
+    private func setTableViewFrame(_ isiPhoneLayout: Bool) {
+        guard let tableView = self.drawerTableView else {
+            return
+        }
+        let resizeBarHeight: CGFloat = isiPhoneLayout ? 20 : 0
```
Leah-Xia-Microsoft, line 185: Can we add some explanation as a comment here about this magic number?

jpeng-ms: hi Leah, this is how our UI is constructed currently: 
image
I did not come up this number if this what you asking

Leah-Xia-Microsoft (dismissed):
LGTM. Tested on iPhone 13

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/FluentUI/Wrapper/DrawerContainerViewController.swift
vhuseinova-msft: Let's create a func/property for `resizeBarHeight` and/or create constants for `20` and `0`, so it will be more clear when it is updated the next time.

jpeng-ms: it seems there are more constants that needed to be refactored, can I create a separate ticket for it?

jpeng-ms: anyways, I refactored constants I used, but will open a refactor ticket anyways to see if there's a better way to use static in generic class

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/FluentUI/Wrapper/DrawerContainerViewController.swift
```diff
@@ -17,18 +17,22 @@ class DrawerContainerViewController<T>: UIViewController, DrawerControllerDelega
     private let sourceView: UIView
     private let showHeader: Bool
     private let isRightToLeft: Bool
+    private weak var controller: DrawerController?
+
+    // MARK: Constants
     private var halfScreenHeight: CGFloat {
```
jpeng-ms, line 23: I tried to refactor those constants in the form of ```enum Constant { ... }``` as we agreed previously. But since this class contains generics, so we cannot go with this. I am open to suggestions: 
image

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/FluentUI/Wrapper/DrawerContainerViewController.swift
```diff
@@ -162,4 +180,13 @@ class DrawerContainerViewController<T>: UIViewController, DrawerControllerDelega
             }
         }.map { index in return tableView.rectForRow(at: index).height }.reduce(0, +)
     }
+
+    private func setTableViewFrame(_ isiPhoneLayout: Bool) {
```
kevinyulu, line 184: minor: can we consider changing name to  `setInitialTableViewFrame` or `initializeTableViewFrame`? The reason for this is that the tableview frame is set with an initial width of height 0 and default width few lines below, which seems like  the function is used during initialization/ reset time only.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/FluentUI/Wrapper/DrawerContainerViewController.swift
```diff
@@ -17,18 +17,22 @@ class DrawerContainerViewController<T>: UIViewController, DrawerControllerDelega
...
     private let isRightToLeft: Bool
+    private weak var controller: DrawerController?
+
+    // MARK: Constants
     private var halfScreenHeight: CGFloat {
         UIScreen.main.bounds.height / 2
     }
-    private weak var controller: DrawerController?
+    private var resizeBarHeight: CGFloat {
+        UIDevice.current.userInterfaceIdiom == .phone ? 20 : 0
```
JoshuaLai, line 27: Where did we get these numbers from? Was it from Alex or was it just a choice by developers cause it looked good?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/FluentUI/Wrapper/DrawerContainerViewController.swift
```diff
@@ -67,6 +71,18 @@ class DrawerContainerViewController<T>: UIViewController, DrawerControllerDelega
     }
 
     func updateDrawerList(items: [T]) {
+        // if contents are identical, do nothing
+        guard self.items != items else {
```
JoshuaLai, line 75: I'm a little confused here. If we are never assigning the items in the `init` any more why do we even want an update method?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/FluentUI/Wrapper/DrawerContainerViewController.swift
jpeng-ms: hi josh, those numbers were hard coded at function level, since I need them for another function now, I am moving them to global level

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/FluentUI/Wrapper/DrawerContainerViewController.swift
jpeng-ms: Good question! We still need this function to update drawer content from ```[]``` to actual list. If we assign drawer content during ```init()```, this function would prevent resizeDrawer() gets called and create another animation bug where on iPad, you might noticed a blank drawer shown then gets updated to actual list immediately.

little context here, not sure if this is relevant:
This function gets called from ```updateUIViewController``` whenever there's a state change even through those state changes are irrelevant to the content of items (participant list, audio list, end call option list). The main reason we are having this is because the ```drawer component``` from ```Fluent UI``` is written in ```UIKit```, we have to wrapped it into a UIViewControllerRepresentable so that it can be used in ```SwiftUI```. Therefore, this function would be called even though ```items``` did not change. 

 If we remove ```updateDrawerList() ```, drawer content would become static.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/FluentUI/Wrapper/DrawerContainerViewController.swift
jpeng-ms: good point, I will update it

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/FluentUI/Wrapper/DrawerContainerViewController.swift
jpeng-ms: https://user-images.githubusercontent.com/109105353/194678834-c59e50cd-06e9-431e-bb84-671acaad915d.mov

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/FluentUI/Wrapper/DrawerContainerViewController.swift
jpeng-ms: updated, thanks!

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/FluentUI/Wrapper/DrawerContainerViewController.swift
palatter: May be we can include them in constants with a name?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/FluentUI/Wrapper/DrawerContainerViewController.swift
vhuseinova-msft: Not sure if it's a good solution but we can use static Computed Properties for generics. Please see `myValue` property:
Screen Shot 2022-10-12 at 10 14 59 AM

jpeng-ms: thanks I will update my PR

JoshuaLai (approved):

kevinyulu (approved):

palatter (approved):
Looks good and tested.

PR #389, opened 2022-10-11
[Feature] Update the functionality of getCurrentDiagnosticInfo

JoshuaLai:
## Purpose
Calling manager to fetch the CallId 

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

### DIFF
--- a/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallComposite.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallComposite.swift
@@ -30,6 +30,7 @@ public class CallComposite {
     private var audioSessionManager: AudioSessionManagerProtocol?
     private var remoteParticipantsManager: RemoteParticipantsManagerProtocol?
     private var avatarViewManager: AvatarViewManagerProtocol?
+    private var diagnosticsManager: DiagnosticsManagerProtocol?
 
     /// Create an instance of CallComposite with options.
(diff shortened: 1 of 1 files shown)


AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallComposite.swift
```diff
@@ -95,7 +96,14 @@ public class CallComposite {
     }
 
     public func getCurrentDiagnosticInfo​() -> DiagnosticsInfo {
-        return DiagnosticsInfo(callId: "")
+        guard let diagnosticsManager = diagnosticsManager else {
+            return DiagnosticsInfo(callId: "")
```
jimchou-dev, line 100: Maybe change this to take default value of empty string?

vhuseinova-msft (approved):

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallComposite.swift
JoshuaLai: Thinking about what a default value might look like. 
In theory the only way we would fall into this case is if something about setting up the DI goes wrong. 
That being said @vhuseinova-msft we need to discuss what would be returned in the case where developers call this method in the SetUp view. And in that case do we want to return an `""` or some kind of default value

jimchou-dev (approved):

PR #391, opened 2022-10-12
[Feature] Implemented event handlers for service layer for Chat Composite

jimchou-dev:
## Purpose

* Implemented service layer to handle TrouterEvents (realtimeNotificationConnect, realtimeNotificationDisconnect, threadDeleted, threadPropertiesUpdated, participantsAdded, participantsRemoved)

```
[ ] Yes
[ x ] No
```

## Pull Request Type

```
[ ] Bugfix
[ x ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...

### DIFF
--- a/AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Model/ChatEventModel.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Model/ChatEventModel.swift
@@ -6,6 +6,8 @@
 import Foundation
 
 enum ChatEventType: Equatable {
+    case realTimeNotificationConnected
+    case realTimeNotificationDisconnected
     case chatMessageReceived
     case chatMessageEdited
     case chatMessageDeleted
@@ -19,5 +21,10 @@ enum ChatEventType: Equatable {
 
 struct ChatEventModel {
     let eventType: ChatEventType
-    let infoModel: BaseInfoModel
+    let infoModel: BaseInfoModel?
+
+    init(eventType: ChatEventType, infoModel: BaseInfoModel? = nil) {
+        self.eventType = eventType
+        self.infoModel = infoModel
+    }
 }
--- a/AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Model/ChatThreadInfoModel.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Model/ChatThreadInfoModel.swift
@@ -10,8 +10,8 @@ struct ChatThreadInfoModel: BaseInfoModel {
     let topic: String?
     let receivedOn: Iso8601Date
 
-    init(topic: String? = nil, receivedOn: Iso8601Date?) {
+    init(topic: String? = nil, receivedOn: Iso8601Date) {
         self.topic = topic
-        self.receivedOn = receivedOn ?? Iso8601Date()
+        self.receivedOn = receivedOn
     }
 }
--- a/AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Model/ChatThreadInfoModel.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Model/ChatThreadInfoModel.swift
@@ -1,0 +1,17 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import AzureCore
+import Foundation
+
+struct ChatThreadInfoModel: BaseInfoModel {
+    let topic: String?
+    let receivedOn: Iso8601Date
+
+    init(topic: String? = nil, receivedOn: Iso8601Date?) {
+        self.topic = topic
+        self.receivedOn = receivedOn ?? Iso8601Date()
+    }
+}
--- a/AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Redux/Middleware/ChatServiceEventHandler.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Redux/Middleware/ChatServiceEventHandler.swift
@@ -64,7 +64,7 @@ class ChatServiceEventHandler: ChatServiceEventHandling {
                       let participantsInfo as ParticipantsInfoModel):
                     self.handleParticipantsRemoved(dispatch: dispatch, participantsInfo: participantsInfo)
                 default:
-                    print("ChatServiceEventHandler subscription switch: default case")
+                    self.logger.warning("ChatServiceEventHandler subscription switch: default case")
                 }
             }.store(in: cancelBag)
     }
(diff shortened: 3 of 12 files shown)


AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Model/ChatThreadInfoModel.swift
```diff
@@ -0,0 +1,17 @@
...
+import AzureCore
+import Foundation
+
+struct ChatThreadInfoModel: BaseInfoModel {
+    let topic: String?
+    let receivedOn: Iso8601Date
+
+    init(topic: String? = nil, receivedOn: Iso8601Date?) {
+        self.topic = topic
+        self.receivedOn = receivedOn ?? Iso8601Date()
```
jpeng-ms, line 15: hi I was wondering if it makes sense if we make init failable so that we don't need those default values anymore?

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Model/ChatEventModel.swift
```diff
@@ -19,5 +21,10 @@ enum ChatEventType: Equatable {
 
 struct ChatEventModel {
     let eventType: ChatEventType
-    let infoModel: BaseInfoModel
+    let infoModel: BaseInfoModel?
+
+    init(eventType: ChatEventType, infoModel: BaseInfoModel? = nil) {
```
jpeng-ms, line 26: I feel like in here, the info model should not be optional so that we can skip dispatch those event early in ChatServiceEventHandler instead of passing it down to other layers and let them handle it

jpeng-ms (dismissed):
looks good, thanks for adding those functions stubs in [ChatServiceEventHandler

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Model/ChatEventModel.swift
jimchou-dev: The case that uses the infoModel nil is with the realTimeNotificationConnected/Disconnected that it doesn't have any data passed from ChatSDK, just the event type only.

jpeng-ms (dismissed):
looks good!

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Redux/Middleware/ChatServiceEventHandler.swift
```diff
@@ -32,26 +32,97 @@ class ChatServiceEventHandler: ChatServiceEventHandling {
...
+                      let chatThreadInfo as ChatThreadInfoModel):
+                    self.handleChatThreadPropertiesUpdated(dispatch: dispatch, threadInfo: chatThreadInfo)
+                case (.participantsAdded,
+                      let participantsInfo as ParticipantsInfoModel):
+                    self.handleParticipantsAdded(dispatch: dispatch, participantsInfo: participantsInfo)
+                case (.participantsRemoved,
+                      let participantsInfo as ParticipantsInfoModel):
+                    self.handleParticipantsRemoved(dispatch: dispatch, participantsInfo: participantsInfo)
                 default:
                     print("ChatServiceEventHandler subscription switch: default case")
```
vhuseinova-msft, line 67: minor: it may be better to use logger here

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Service/Chat/Extension/TrouterEventExtension.swift
```diff
@@ -20,6 +20,41 @@ extension ChatMessageReceivedEvent {
...
+        let participants = participantsAdded.map {
+            $0.toParticipantInfoModel()
+        }
+        return ParticipantsInfoModel(
+            participants: participants)
+    }
+}
+
+extension ParticipantsRemovedEvent {
+    func toParticipantsInfo(_ participantsAdded: [SignalingChatParticipant]) -> ParticipantsInfoModel {
```
vhuseinova-msft, line 49: Should it be `participantsRemoved`?

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Tests/Redux/Middleware/ChatServiceEventHandlerTests.swift
```diff
@@ -67,4 +100,83 @@ class ChatServiceEventHandlerTests: XCTestCase {
...
+        mockChatService.chatEventSubject.send(chatEventModel)
+        wait(for: [expectation], timeout: 1)
+    }
+
+    func test_chatServiceEventHandler_subscription_when_receiveChatThreadPropertiesUpdatedEvent_then_dispatchChatTopicUpdatedAction() {
+        let expectation = XCTestExpectation(description: "Dispatch the new action")
+        func dispatch(action: Action) {
+            switch action {
+            case .chatAction(.chatTopicUpdated(_)):
+                expectation.fulfill()
```
vhuseinova-msft, line 127: Could you add check that topic value is the same that was send with `ChatThreadInfoModel`

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Tests/Redux/Middleware/ChatServiceEventHandlerTests.swift
```diff
@@ -67,4 +100,83 @@ class ChatServiceEventHandlerTests: XCTestCase {
...
+        mockChatService.chatEventSubject.send(chatEventModel)
+        wait(for: [expectation], timeout: 1)
+    }
+
+    func test_chatServiceEventHandler_subscription_when_receiveParticipantsAddedEvent_then_dispatchParticipantsAddedAction() {
+        let expectation = XCTestExpectation(description: "Dispatch the new action")
+        func dispatch(action: Action) {
+            switch action {
+            case .participantsAction(.participantsAdded(_)):
+                expectation.fulfill()
```
vhuseinova-msft, line 146: It may be a good idea to check that participants' ids are the same that were send with ChatEventModel

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Redux/Reducer/ParticipantsReducer.swift
```diff
@@ -11,8 +11,10 @@ extension Reducer where State == ParticipantsState,
     static var liveParticipantsReducer: Self = Reducer { participantsState, action in
 
         switch action {
-        case .participantsAction(.leaveChatSuccess):
-            print("ParticipantsReducer `leaveChatSuccess` not implemented")
+        case .participantsAction(.participantsAdded(let participants)):
```
palatter, line 14: Do we have a way of using the logger in static situations like this?

jimchou-dev: I don't think we have access to logger in the reducer from DI, these print statement are temporary should not be there once things are implemented.

palatter (approved):
Looks good.

jpeng-ms (approved):

PR #397, opened 2022-10-14
[Feature] Implemented send message for ChatComposite

jimchou-dev:
## Purpose

* Implemented send message for chat composite for both service layer and redux layer
* Introduced internalId when sending message and update with actual messageId from ChatSDK response
  * When receive TrouterEvent for the local user's send message will replace the message with same id in message repository
* Filter logs by `*Message` when receiving and sending message to see the logs in console

```
[ ] Yes
[ x ] No
```

## Pull Request Type

```
[ ] Bugfix
[ x ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...

### DIFF
--- a/AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Manager/MessageRepositoryManager.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Manager/MessageRepositoryManager.swift
@@ -11,6 +11,8 @@ protocol MessageRepositoryManagerProtocol {
 
     // MARK: sending local events
     func addInitialMessages(initialMessages: [ChatMessageInfoModel])
+    func addNewSentMessage(message: ChatMessageInfoModel)
+    func replaceMessageId(internalId: String, actualId: String)
 
     // MARK: receiving remote events
     func addReceivedMessage(message: ChatMessageInfoModel)
@@ -29,8 +31,27 @@ class MessageRepositoryManager: MessageRepositoryManagerProtocol {
         messages = initialMessages
     }
 
-    func addReceivedMessage(message: ChatMessageInfoModel) {
+    func addNewSentMessage(message: ChatMessageInfoModel) {
         messages.append(message)
     }
 
+    func replaceMessageId(internalId: String, actualId: String) {
+        if let index = messages.firstIndex(where: {
+            $0.internalId == internalId
+        }) {
+            var msg = messages[index]
+            msg.replace(id: actualId)
+            messages[index] = msg
+        }
+    }
+
+    func addReceivedMessage(message: ChatMessageInfoModel) {
+        if let index = messages.firstIndex(where: {
+            $0.id == message.id
+        }) {
+            messages[index] = message
+        } else {
+            messages.append(message)
+        }
+    }
 }
--- a/AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Manager/MessageRepositoryManager.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Manager/MessageRepositoryManager.swift
@@ -37,7 +37,7 @@ class MessageRepositoryManager: MessageRepositoryManagerProtocol {
 
     func replaceMessageId(internalId: String, actualId: String) {
         if let index = messages.firstIndex(where: {
-            $0.internalId == internalId
+            $0.id == internalId
         }) {
             var msg = messages[index]
             msg.replace(id: actualId)
--- a/AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Manager/MessageRepositoryManager.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Manager/MessageRepositoryManager.swift
@@ -11,7 +11,7 @@ protocol MessageRepositoryManagerProtocol {
 
     // MARK: sending local events
     func addInitialMessages(initialMessages: [ChatMessageInfoModel])
-    func addNewSentMessage(message: ChatMessageInfoModel)
+    func addNewSendingMessage(message: ChatMessageInfoModel)
     func replaceMessageId(internalId: String, actualId: String)
 
     // MARK: receiving remote events
@@ -31,7 +31,7 @@ class MessageRepositoryManager: MessageRepositoryManagerProtocol {
         messages = initialMessages
     }
 
-    func addNewSentMessage(message: ChatMessageInfoModel) {
+    func addNewSendingMessage(message: ChatMessageInfoModel) {
         messages.append(message)
     }
 
--- a/AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/SwiftUI/Chat/ChatViewComponents/MessageInputViewModel.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/SwiftUI/Chat/ChatViewComponents/MessageInputViewModel.swift
@@ -23,7 +23,9 @@ class MessageInputViewModel: ObservableObject {
         guard !message.isEmpty else {
             return
         }
-//        dispatch(.chatAction(.))
+        dispatch(.repositoryAction(.sendMessageTriggered(
+            internalId: UUID().uuidString,
+            content: message)))
         message = ""
     }
 }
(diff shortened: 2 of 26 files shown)


Leah-Xia-Microsoft (dismissed):

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Manager/MessageRepositoryManager.swift
```diff
@@ -29,8 +31,27 @@ class MessageRepositoryManager: MessageRepositoryManagerProtocol {
         messages = initialMessages
     }
 
-    func addReceivedMessage(message: ChatMessageInfoModel) {
+    func addNewSentMessage(message: ChatMessageInfoModel) {
```
palatter, line 34: Do we need 'new' in method here?

jimchou-dev: I added the 'new' because the message is just sending (not yet delivered) as a placeholder dummy message. If it's just addSentMessage` feels it's already finish sending. But let me know if you can come up with some better name to avoid the confusion I'll be happy to rename it!

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Service/Chat/ChatSDKWrapper.swift
```diff
@@ -70,6 +70,29 @@ class ChatSDKWrapper: NSObject, ChatSDKWrapperProtocol {
...
+                senderDisplayName: senderDisplayName
+            )
+            return try await withCheckedThrowingContinuation { continuation in
+                chatThreadClient?.send(message: messageRequest) { result, _ in
+                    switch result {
+                    case let .success(result):
+                        continuation.resume(returning: result.id)
+                    case .failure(let error):
+                        self.logger.error("Failed to send message: \(error)")
+                        continuation.resume(throwing: error)
```
palatter, line 86: Does this throw end up in the parent catch as well?

jimchou-dev: Line 86 will throw, and line 92 will catch and re-throw it. Yes it will be logged twice (once in line 85 once in line 91).

Currently the error actions are not yet being handled, all the error/throwing and logging will be cleaned up under the error handling User Story

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/SwiftUI/Chat/ChatViewComponents/MessageInputViewModel.swift
```diff
@@ -23,7 +23,9 @@ class MessageInputViewModel: ObservableObject {
         guard !message.isEmpty else {
             return
         }
-//        dispatch(.chatAction(.))
+        dispatch(.repositoryAction(.sendMessageTriggered(
+            internalId: UUID().uuidString,
```
JoshuaLai, line 27: I'm guessing this is our key in the message repository?

JoshuaLai: Why the difference between the message id and making our own?

jimchou-dev: The message will have 2 kind of identifier: `id` (from chatSDK) and `internalId` (randomly generated UUID). When we send an message to chatSDK, we do not know what will be the `id` yet, so need to use the `internalId` to replace/update the messageInfoModel with the actual id when sendMessage is successful to ChatSDK.

I don't think this will be the key for message repository data stricture, but probably just searching backward from back (or forward from front) since its newly added so usually will be O(1) with iterating through only a few messages.
But maybe there can be some kind of data structure that will have both `id` and `internalId` as key.

Also another note, the `id` from ChatSDK is epoch time in milliseconds, so we could use that as some kind of sorted map data structure (to be implemented/spiked with the MessageRepo user story)

JoshuaLai: Makes sense!
`Also another note, the id from ChatSDK is epoch time in milliseconds, so we could use that as some kind of sorted map data structure (to be implemented/spiked with the MessageRepo user story)`
Does it make sense instead of being a uuid we also use epoch time as well? 
Just looking at the message model its a little odd that we have ID as optional and I am guessing this is the reason. I am thinking we can clean up the model a bit by just having one id, and we fill in an epoch time of what we think it is, then when the server that confirmed received the message and given us a new id we just replace it. Will that work or just make it harder and more complex?

jimchou-dev: Yeah I think it's possible to just have one `id`, but I am not so sure if we should be using epoch time as dummy/internalId since it doesn't exactly guarantee to be unique when compared to UUID (Universally unique identifier) I'll make another commit to do this refactoring.

JoshuaLai: In theory epoch time should be unique enough (including the milliseconds?) Thats why the ChatSDK send it to us this way. But thats just a theory, if some one rapidly sends messages quickly one after another it may raise a concern

jimchou-dev: Yeah I think the probability of clashing id is low, but not guaranteed, for now I'll keep the internal/dummy id as uuid, and its a simple refactor in the future if need to change to epoch milliseconds. New commit added to remove the internalId field.

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Redux/Middleware/RepositoryMiddleware.swift
```diff
@@ -45,12 +45,34 @@ private func handleParticipantsAction(
...
+    _ actionHandler: RepositoryMiddlewareHandling,
+    _ getState: () -> AppState,
+    _ dispatch: @escaping ActionDispatch) {
         switch action {
+
+            // MARK: local events
+
         case .fetchInitialMessagesSuccess(let messages):
-            actionHandler.loadInitialMessages(messages: messages)
+            actionHandler.loadInitialMessages(messages: messages,
```
kevinyulu, line 56: minor: we can rename fetchInitialMessages to loadInitialMessages or vice versa

kevinyulu (dismissed):

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Redux/Middleware/RepositoryMiddleware.swift
jimchou-dev: The actions are a set: fetchInitialMessagesTriggered, fetchInitialMessagesSuccess, fetchInitialMessagesFailed

And the function loadInitialMessages is taking the result from the action and loading to the messaging repository so I think will leave the naming of function as `loadInitialMessages`

JoshuaLai (dismissed):

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/SwiftUI/Chat/ChatViewComponents/MessageInputViewModel.swift
palatter: I have found the time stamp and ID are not always perfectly exact. Sometimes off by a few milliseconds. Something we need to investigate to ensure it's consistent.

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Manager/MessageRepositoryManager.swift
palatter: Maybe addSendingMessage? Not sure if that's any better though 🤷‍♂️

palatter (approved):
Looks good.

PR #402, opened 2022-10-18
[Feature] Implemented chat message history pagination

jimchou-dev:
## Purpose

* Implement chat message history pagination for service and redux layer
* Because UI is not implemented, so added a chat command to fetch previous message and see result in console if the text equals to `fetch` and send using the UI
* Note currently the behaviour maxPagesize is `5` for testing, when page backward will also include recent sent/received message since previous pageNext (should not be an issue when have page size of 50 or 200). Recent/duplicate message will be filtered/ignored by message repository

```
[ ] Yes
[ x ] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...

### DIFF
--- a/AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Manager/MessageRepositoryManager.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Manager/MessageRepositoryManager.swift
@@ -11,6 +11,7 @@ protocol MessageRepositoryManagerProtocol {
 
     // MARK: sending local events
     func addInitialMessages(initialMessages: [ChatMessageInfoModel])
+    func addPreviousMessages(previousMessages: [ChatMessageInfoModel])
     func addNewSendingMessage(message: ChatMessageInfoModel)
     func replaceMessageId(internalId: String, actualId: String)
 
@@ -29,6 +30,26 @@ class MessageRepositoryManager: MessageRepositoryManagerProtocol {
 
     func addInitialMessages(initialMessages: [ChatMessageInfoModel]) {
         messages = initialMessages
+    }
+
+    func addPreviousMessages(previousMessages: [ChatMessageInfoModel]) {
+        // work around for duplicate message
+        // will use more efficient data structure in MessageRepo PR
+        for m in previousMessages {
+            if let index = messages.firstIndex(where: {
+                $0.id == m.id
+            }) {
+                messages[index] = m
+            } else {
+                messages.append(m)
+            }
+        }
+
+        messages.sort { lhs, rhs -> Bool in
+            // createdOn does not have milliseconds
+            return lhs.createdOn == rhs.createdOn ?
+            lhs.id < rhs.id : lhs.createdOn < rhs.createdOn
+        }
     }
 
     func addNewSendingMessage(message: ChatMessageInfoModel) {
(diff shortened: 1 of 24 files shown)


AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Manager/MessageRepositoryManager.swift
```diff
@@ -31,6 +32,25 @@ class MessageRepositoryManager: MessageRepositoryManagerProtocol {
...
+                $0.id == m.id
+            }) {
+                messages[index] = m
+            } else {
+                messages.append(m)
+            }
+        }
+
+        messages.sort { lhs, rhs -> Bool in
+            // createdOn does not have milliseconds
```
palatter, line 48: What is the implication of not having milliseconds here?

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Manager/MessageRepositoryManager.swift
```diff
@@ -31,6 +32,25 @@ class MessageRepositoryManager: MessageRepositoryManagerProtocol {
...
+            return lhs.createdOn == rhs.createdOn ?
+            lhs.id < rhs.id : lhs.createdOn < rhs.createdOn
```
palatter, line 50: Should we always sort on id?

jimchou-dev: Only when two message has exactly the same createdOn without millisecond, in this example each line has the data messageId, createdOn, content.
image
The message `15` and `16` were sent roughly the same second so they have the same createdOn, and when sorting by createdOn only may cause them to be in incorrect order. But the returned messageId help us differentiate which came first after sorting.

And since we can't guarantee the generated epoch time with millisecond will be unique for dummy message placerholderId, we are using UUID for that but should have the approximate createdOn when sorted.

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/SwiftUI/Chat/ChatViewComponents/BottomBarViewModel.swift
```diff
@@ -34,6 +34,12 @@ class BottomBarViewModel: ObservableObject {
     @Published var hasFocus: Bool = true
 
     func sendMessage() {
+        // Added for testing to trigger action
+        if message == "fetch" {
```
jpeng-ms, line 38: I guess this is would be removed in following PRs right?

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/SwiftUI/Chat/ChatViewComponents/MessageListViewModel.swift
```diff
@@ -27,7 +27,7 @@ class MessageListViewModel: ObservableObject {
             // for testing
             print("*Messages count: \(messageRepositoryManager.messages.count)")
             for m in messageRepositoryManager.messages {
-                print("--*Messages: \(m.id) \(m.content)")
+                print("--*Messages: \(m.id) \(m.createdOn.value) \(m.content)")
```
jpeng-ms, line 30: maybe use logger later?

jpeng-ms (approved):
Looks good!

palatter (approved):

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/SwiftUI/Chat/ChatViewComponents/BottomBarViewModel.swift
jimchou-dev: Correct this is because there is no UI to trigger the action yet, will be removed later.

PR #408, opened 2022-10-20
[Feature] Implemented redux for Chat Service events

jimchou-dev:
## Purpose

* Implemented redux for Chat Service events (ChatThreadPropertiesUpdated, ChatThreadDeleted, ChatMessageEdited, ChatMessageDeleted)
* Implemented error handler to handle ChatThreadDeleted to exit the composite

```
[ ] Yes
[ x ] No
```

## Pull Request Type

```
[ ] Bugfix
[ x ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...

### DIFF
--- a/AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/ChatCompositeOptions/ChatCompositeError.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/ChatCompositeOptions/ChatCompositeError.swift
@@ -30,6 +30,9 @@ public struct ChatCompositeErrorCode {
     /// Error when local user fails to resume a chat.
     static let chatResume: String = "chatResume"
 
+    /// Error when local user fails to send message.
+    static let sendMessage: String = "sendMessage"
+
     /// Error when failed to parse threadId from teams meeting URL
     static let parseThreadIdFailed: String = "parseThreadIdFailed"
 }
--- a/AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Manager/MessageRepositoryManager.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Manager/MessageRepositoryManager.swift
@@ -15,7 +15,10 @@ protocol MessageRepositoryManagerProtocol {
     func replaceMessageId(internalId: String, actualId: String)
 
     // MARK: receiving remote events
+    func addTopicUpdatedMessage(chatThreadInfo: ChatThreadInfoModel)
     func addReceivedMessage(message: ChatMessageInfoModel)
+    func updateMessageEdited(message: ChatMessageInfoModel)
+    func updateMessageDeleted(message: ChatMessageInfoModel)
 }
 
 class MessageRepositoryManager: MessageRepositoryManagerProtocol {
@@ -45,6 +48,18 @@ class MessageRepositoryManager: MessageRepositoryManagerProtocol {
         }
     }
 
+    func addTopicUpdatedMessage(chatThreadInfo: ChatThreadInfoModel) {
+        guard let topic = chatThreadInfo.topic else {
+            return
+        }
+        let topicUpdatedSystemMessage = ChatMessageInfoModel(
+            type: .topicUpdated,
+            content: topic,
+            createdOn: chatThreadInfo.receivedOn
+        )
+        messages.append(topicUpdatedSystemMessage)
+    }
+
     func addReceivedMessage(message: ChatMessageInfoModel) {
         if let index = messages.firstIndex(where: {
             $0.id == message.id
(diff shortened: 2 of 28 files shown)


AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Redux/Action/ChatAction.swift
```diff
@@ -27,6 +27,9 @@ enum ChatAction: Equatable {
         case let (.topicRetrieved(lStr), .topicRetrieved(rStr)):
             return lStr == rStr
 
+        case let (.chatTopicUpdated(lInfo), .chatTopicUpdated(rInfo)):
```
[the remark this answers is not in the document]
jimchou-dev: Sounds good, we can do that in the future and do the refactoring for both Calling and Chat together.

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Manager/MessageRepositoryManager.swift
```diff
@@ -74,4 +89,20 @@ class MessageRepositoryManager: MessageRepositoryManagerProtocol {
             messages.append(message)
         }
     }
+
+    func updateMessageEdited(message: ChatMessageInfoModel) {
+        if let index = messages.firstIndex(where: {
```
jpeng-ms, line 94: is it possible to have multiple id match? I guess it won't be possible right?

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Utilities/ChatCompositeInternalError.swift
```diff
@@ -23,8 +24,11 @@ enum ChatCompositeInternalError: String, Error, Equatable {
             return ChatCompositeErrorCode.chatEnd
         case .parseThreadIdFailed:
             return ChatCompositeErrorCode.parseThreadIdFailed
-        case .chatEvicted,
-                .chatDenied:
+        case .sendMessageFailed:
+            return ChatCompositeErrorCode.sendMessage
+        case .chatEvicted:
+            return ChatCompositeErrorCode.chatEvicted
+        case .chatDenied:
```
jpeng-ms, line 31: in which case chat would be denied? invalid token?

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/ChatCompositeOptions/ChatCompositeError.swift
```diff
@@ -30,6 +30,9 @@ public struct ChatCompositeErrorCode {
     /// Error when local user fails to resume a chat.
     static let chatResume: String = "chatResume"
 
+    /// Error when local user fails to send message.
+    static let sendMessage: String = "sendMessage"
```
jpeng-ms, line 34: shouldn't those be named "sendMessageFailed" or "chatResumeFailed" reflect the type of error? Have we ARB-ed those?

jpeng-ms (approved):
Looks good

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Manager/MessageRepositoryManager.swift
jimchou-dev: Yes we're making the assumption that there should not be duplicate Id. There should not be duplicated message id (in epoch time milliseconds from chatSDK) if it's duplicated id then ChatSDK would have the same issue.

The MessageRepository data structure would likely be some kind of map in the future that would be more efficient than a list

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Redux/Middleware/ChatActionHandler.swift
```diff
@@ -65,6 +74,12 @@ class ChatActionHandler: ChatActionHandling {
     }
 
     // MARK: Chat Handler
+    func onChatThreadDeleted(dispatch: @escaping ActionDispatch) -> Task<Void, Never> {
+        Task {
```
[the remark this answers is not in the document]
jimchou-dev: Following similar pattern as CallingComposite (eg. onCameraPermissionIsSet), I'm assuming these may come in as events on separate thread so they can be done async?

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Utilities/ChatCompositeInternalError.swift
jimchou-dev: I'm assuming it would be a participant is not added to a chat thread so when launching chat composite with the token it does not have permission. Maybe denied is not the right word, but this PR error handling part is only to get ChatThreadDeleted event working, there is separate user story to deal with Error handling.

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/ChatCompositeOptions/ChatCompositeError.swift
jimchou-dev: I thought it would be sendMessageFailed also, but since it's of errorCode type so people can probably relate them to error? Also based on the pattern in `CallCompositeErrorCode` a lot of the error code does not have the word `failed` int it also, and those were reviewed by ARB.

Leah-Xia-Microsoft (approved):
LGTM

palatter (approved):
Looks good.

PR #409, opened 2022-10-21
[Task] Updated Clearing Typing Indicator Logic

jpeng-ms:
## Purpose
Previous implementation was using timestamps to determine when the participant should be removed from typing indicator. But after discussion with Android, we came into a conclusion that clearing indicator should be done by a redux action. And it should not be handled in view model.

## Testing
animation is from Fluent UI

https://user-images.githubusercontent.com/109105353/197626771-23953ec5-2977-4e4c-bede-37f0c74ea34f.mov

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[X] Feature
[] Refactoring (no functional changes, no api changes)
[ ] Other... Please describe:
```

## How to Test
Because participant list is still being worked on by josh , so you need to hard code the list as the following to see typing indicator:
in participant reducer:
image
2. send actions like these in Chat Service Event Handler
image

## What to Check
1. make sure participant is removed from indicator after 8 seconds exactly 
2. make sure participant can be added to indicator once action is received
3. make sure indicator can be set to hidden if there is no more participant typing


AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Redux/Middleware/ChatServiceEventHandler.swift
```diff
@@ -94,7 +99,11 @@ class ChatServiceEventHandler: ChatServiceEventHandling {
 
     func handleTypingIndicatorReceived(dispatch: @escaping ActionDispatch,
                                        userEventTimestamp: UserEventTimestampModel) {
-        dispatch(.participantsAction(.typingIndicatorReceived(userEventTimestamp: userEventTimestamp)))
+        let timer = Timer.scheduledTimer(withTimeInterval: Constants.typingIndicatorDisplayInterval,
+                                         repeats: false) {_ in
+            dispatch(.participantsAction(.clearTypingIndicator(id: userEventTimestamp.id)))
```
emlynmac, line 104: What happens in the case where you get a second indicator received and you need to refresh the timer? I'd expect some sort of cancellation logic for the original timer and set up of the new one in this case.

jpeng-ms: good question! that logic is handled at here: https://github.com/Azure/communication-ui-library-ios/pull/409/files#diff-3973ab1e19d9c5da2f229072d5bde7312aae94f69be65b565cf2eaa9a9050db8R32

jpeng-ms: hmmm looks like link doesn't work, but it's handled in reducer: 
image

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Redux/Reducer/ParticipantsReducer.swift
```diff
@@ -32,41 +30,36 @@ extension Reducer where State == ParticipantsState,
...
             return state
-        case .participantsAction(.typingIndicatorReceived(userEventTimestamp: let userEventTimestamp)):
-            let typingExpiringTimestamp = userEventTimestamp.timestamp.value + speakingDurationSeconds
-            typingIndicatorMap[userEventTimestamp.id] = typingExpiringTimestamp
-            if typingIndicatorTimestamp < typingExpiringTimestamp {
-                typingIndicatorTimestamp = typingExpiringTimestamp
-            }
+        case .participantsAction(.typingIndicatorReceived(let id, let timer)):
+            typingIndicatorMap[id]?.invalidate()
+            typingIndicatorMap[id] = timer
```
emlynmac, line 44: It would be nice to have just one timer that varies its length and handles a map of participants and started typing time.

jpeng-ms: will have a discussion with Android and sync out the design. Will create another PR to address this

PR #415, opened 2022-10-26
[Feature] Send read receipt presentation layer

Leah-Xia-Microsoft:
## Purpose

* When user enter a chat, send read receipt for the last message of the initial fetch
* When fetching the second batch of messages, send another read receipt for the last message of the batch
* While user is in the chat, when receiving messages, send a read receipt for each message received.

https://user-images.githubusercontent.com/107075081/198394710-3550a230-4f41-4ff0-8242-2ee79b6b24d0.mov

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get and run the code
*  Set breakpoints in `sendReadReceipt()` function


AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Utilities/StringExtension.swift
```diff
@@ -0,0 +1,16 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import Foundation
+
+extension String {
+    func convertEpochStringToTimestamp() -> Date? {
```
jpeng-ms, line 9: is this necessary? I did notice that in AzureCore, there are couple of functions we can use to convert string to Iso8601Date which essentially is wrapper for Date

jpeng-ms (commented):
Looks good!

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Redux/Reducer/ChatReducer.swift
```diff
@@ -21,6 +22,8 @@ extension Reducer where State == ChatState,
                 break
             }
             topic = newTopic
+        case .participantsAction(.sendReadReceiptSuccess(messageId: let messageId)):
+            lastReadReceiptSentTimestamp = messageId.convertEpochStringToTimestamp()
```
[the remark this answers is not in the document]
Leah-Xia-Microsoft: > so message id is a timestamp??

Yes messageId is in epoch time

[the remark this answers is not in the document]
jpeng-ms: interesting, didn't know that!

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Utilities/StringExtension.swift
Leah-Xia-Microsoft: > is this necessary? I did notice that in AzureCore, there are couple of functions we can use to convert string to Iso8601Date which essentially is wrapper for Date, which should be sufficient for comparison you are doing in view model unless I am wrong

Hi @jpeng-ms, I've found the Iso8601Date init from a string from AzureCore  `let timestamp = Iso8601Date(string: messageId) `. However, they expect a string in some type of format like 2022-10-27, not a epoch number. Is there any other functions from AzureCore that you know of to convert `1666897727748` to a timestamp? Thanks.

jpeng-ms: humm, how about UnixTime? It has an init as this:

```
    public init(timeIntervalSince1970 double: Double) {
        self.value = Date(timeIntervalSince1970: double)
    }
```

jpeng-ms: might need to convert string to double to use it

Leah-Xia-Microsoft: Don't really see a point of using that init since it doesn't replace my function here.
It would just change my current function:
```
func convertEpochStringToTimestamp() -> Date? {
        guard let timestampDouble = Double(self) else {
            return nil
        }
        --> this line: let timestamp = Date(timeIntervalSince1970: timestampDouble / 1000)
        return timestamp
}
```
to:
```
func convertEpochStringToTimestamp() -> Date? {
        guard let timestampDouble = Double(self) else {
            return nil
        }
        --> To: let timestamp = Iso8601Date(timeIntervalSince1970: timestampDouble / 1000)
        return timestamp
}
```
which under the hood is still using my original method.

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/SwiftUI/Chat/ChatViewComponents/MessageListView.swift
```diff
@@ -14,6 +14,9 @@ struct MessageListView: View {
 
         static let defaultMinListRowHeight: CGFloat = 10
     }
+    private let sendReadReceiptInterval: Double = 5.0
+    @State private var sendReadReceiptTimer: Timer?
```
emlynmac, line 18: What happens if the timer is not @state defined?

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/SwiftUI/Chat/ChatViewComponents/MessageListView.swift
```diff
@@ -14,6 +14,9 @@ struct MessageListView: View {
...
+    @State private var lastReadMessageIndex: Int?
```
emlynmac, line 19: Should this be part of the view model?

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/SwiftUI/Chat/ChatViewComponents/MessageListView.swift
```diff
@@ -34,10 +37,28 @@ struct MessageListView: View {
...
+                        }
+                    }
                 }
             }
             .listStyle(.plain)
             .environment(\.defaultMinListRowHeight, Constants.defaultMinListRowHeight)
+            .onAppear {
+                sendReadReceiptTimer = Timer.scheduledTimer(withTimeInterval: sendReadReceiptInterval,
+                                                            repeats: true) { _ in
+                    viewModel.sendReadReceipt(messageIndex: lastReadMessageIndex)
```
emlynmac, line 56: So, the timer fires every 5s and sends a receipt for the last message index. Does it matter if some are skipped? Could we do this without a timer?

Leah-Xia-Microsoft: > Member

1. It doesn't matter if some are skipped. On the receiving side, all we care about is the latest read message for a certain participant. We assume all the messages previous to the latest one is read as well.
2. We had a meeting with the Android side and agreed on this timer approach because we only want to send read receipt for messages that's actually displayed on screen. There might be messages we received but we haven't scroll to them yet. However, we don't want to keep sending read receipt while scrolling. We were deciding between debouncing or a timer. We think timer catches more cases such as receiving messages while chat screen is open.

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/SwiftUI/Chat/ChatViewComponents/MessageListView.swift
Leah-Xia-Microsoft: > What happens if the timer is not @State defined?

Adding @State just because I need to assign value to it later. Should I also move this to ViewModel?

jpeng-ms: I do think these two should be in view model as they have nothing to do with how we layout the view

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/SwiftUI/Chat/ChatViewComponents/MessageListViewModel.swift
```diff
@@ -9,20 +9,21 @@ class MessageListViewModel: ObservableObject {
     private let messageRepositoryManager: MessageRepositoryManagerProtocol
     private let logger: Logger
     private let dispatch: ActionDispatch
-
     private var repositoryUpdatedTimestamp: Date = .distantPast
     private var localUserId: String? // Remove optional?
+    private let sendReadReceiptInterval: Double = 5.0
```
palatter, line 14: Minor: Can we put this with the other private lets?

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/SwiftUI/Chat/ChatViewComponents/MessageListView.swift
```diff
@@ -30,30 +30,34 @@ struct MessageListView: View {
...
+                .listStyle(.plain)
+                .environment(\.defaultMinListRowHeight, Constants.defaultMinListRowHeight)
+                .onAppear {
+                    viewModel.setSendReadReceiptTimer()
+                    scrollToBottom(proxy: proxy, bottomIndex: viewModel.messages.count)
+                }
+                .onChange(of: viewModel.messages.count) { _ in
+                    scrollToBottom(proxy: proxy, bottomIndex: viewModel.messages.count)
+                }
+                .onDisappear {
```
palatter, line 58: Can we move the viewModel.setSendReadReceiptTimer() to the messageList above? As it is not specific to the scrollview/lazyVstack

`
messageList
.onAppear {
    viewModel.setSendReadReceiptTimer()
}
.onDisappear {
    viewModel.sendReadReceiptTimer?.invalidate()
}
`

Leah-Xia-Microsoft: > Can we move the viewModel.setSendReadReceiptTimer() to the messageList above? As it is not specific to the scrollview/lazyVstack
> 
> messageList .onAppear { viewModel.setSendReadReceiptTimer() } .onDisappear { viewModel.sendReadReceiptTimer?.invalidate() }

Good catch. Got misplaced during the merge conflict. Will change it.

palatter (dismissed):
Looks good.

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/SwiftUI/Chat/ChatViewComponents/MessageListViewModel.swift
```diff
@@ -77,4 +78,29 @@ class MessageListViewModel: ObservableObject {
...
+            self.lastReadMessageIndex = index
+            return
+        }
+        if index > lastReadMessageIndex {
+            self.lastReadMessageIndex = index
+        }
+    }
+
+    func setSendReadReceiptTimer() {
+        sendReadReceiptTimer = Timer.scheduledTimer(withTimeInterval: sendReadReceiptInterval,
```
emlynmac, line 93: This repeating timer is not invalidated. Let's manage the lifetime.

Leah-Xia-Microsoft: > This repeating timer is not invalidated. Let's manage the lifetime.

Good catch. Accidentally setup the timer twice during merge.
Removed the extra `viewModel.setSendReadReceiptTimer()`

palatter (dismissed):
Looks good.

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/SwiftUI/Chat/ChatViewComponents/MessageListView.swift
```diff
@@ -21,6 +21,12 @@ struct MessageListView: View {
 
     var body: some View {
         messageList
+            .onAppear {
+                viewModel.setSendReadReceiptTimer()
+            }
+            .onDisappear {
+                viewModel.sendReadReceiptTimer?.invalidate()
```
emlynmac, line 28: While this will do what we need, it would be better to create the externally facing method on the view model for `messageListDisappeared` or something similar and then use that method to implement whatever is needed when the list disappears from view.
Generally, we should keep internal workings isolated; so here that would mean removing the knowledge at the message list view that there's a timer going in the VM.

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/SwiftUI/Chat/ChatViewComponents/MessageListViewModel.swift
emlynmac: What happens if the view model is deallocated without the timer being invalidated?

emlynmac (approved):

PR #420, opened 2022-10-29
[Task] Add UI for sharing call id

vhuseinova-msft:
## Purpose

* Added UI for More Call options
* Added UI for sharing call id
* Added UI tests for sharing call id scenarios
* Updated DrawerContainerViewController `showDrawer` setup

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

## What to Check
* More button is available for Control Bar
* Call id can be shared
* Call id is accessible and can be shared via VoiceOver
* UI and unit tests can be run


AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallComposite.swift
```diff
@@ -32,7 +32,7 @@ public class CallComposite {
     private var avatarViewManager: AvatarViewManagerProtocol?
     private var diagnosticsManager: DiagnosticsManagerProtocol?
 
-    /// Create an instance of CallComposite with options.
+    /// Create an instance of Call Composite with options.
```
JoshuaLai, line 35: nitpick: I think this should be reverted, we are referring to the type of the object so it should be correct to use `CallComposite`

vhuseinova-msft: Updated!

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallCompositeOptions/DiagnosticConfig.swift
```diff
@@ -30,7 +30,9 @@ struct DiagnosticConfig {
 
 }
 
+/// A Call Composite diagnostics information.
 public struct DiagnosticsInfo {
+    /// The last known call id.
     public let lastKnownCallId: String?
 
     init(lastKnownCallId: String? = nil) {
```
JoshuaLai, line 38: You will also need to update this `init` with a block comment so the docs generate it properly

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/Factories/CompositeViewModelFactory.swift
```diff
@@ -51,6 +51,12 @@ protocol CompositeViewModelFactoryProtocol {
     func makeBannerTextViewModel() -> BannerTextViewModel
     func makeLocalParticipantsListCellViewModel(localUserState: LocalUserState) -> ParticipantsListCellViewModel
     func makeParticipantsListCellViewModel(participantInfoModel: ParticipantInfoModel) -> ParticipantsListCellViewModel
+    func makeMoreCallOptionsListViewModel(showSharingViewAction: @escaping () -> Void) -> MoreCallOptionsListViewModel
+    func makeMoreCallOptionsListCellViewModel(icon: CompositeIcon,
```
JoshuaLai, line 55: I wonder does it make sense, or possible instead of having this method take in all these arguments to just have it take in a PrimaryButtonViewModel?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/ViewComponents/SharingActivity/DiagnosticsSharingActivityViewModel.swift
```diff
@@ -0,0 +1,23 @@
...
+
+    init(accessibilityProvider: AccessibilityProviderProtocol,
+         diagnosticsManager: DiagnosticsManagerProtocol) {
+        self.accessibilityProvider = accessibilityProvider
+        self.diagnosticsManager = diagnosticsManager
+    }
+
+    func getDiagnosticsInfo() -> String {
+        let diagnosticsInfo = diagnosticsManager.getDiagnosticsInfo()
+        let callId = diagnosticsInfo.lastKnownCallId ?? "UNKNOWN"
```
JoshuaLai, line 20: If we are going to default to `Unknown` anyways its better to have the `lastKnownCallId` instead of being optional just default to `Unknown` in the init

JoshuaLai:
Everything looks good and tested everything working as expected. 
Just left some suggestions.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallCompositeOptions/DiagnosticConfig.swift
vhuseinova-msft: Do we need to add it even if it's not public?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/ViewComponents/SharingActivity/DiagnosticsSharingActivityViewModel.swift
vhuseinova-msft: I want to show `UNKNOWN` only for UI. When a Contoso developer will request the Diagnostics Info, `lastKnownCallId` will return `nil`

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/Factories/CompositeViewModelFactory.swift
vhuseinova-msft: I don't think that it should use `PrimaryButtonViewModel` as they look quite different. But it may be a good idea to check if we can reuse 1 VM type for drawers cells VM

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallCompositeOptions/DiagnosticConfig.swift
vhuseinova-msft: No, don't need it as it's not public

jpeng-ms:
have we gotten any response regarding my comment in design doc? I feel like being able to access call id on lobby view is a quite important use case for this.

vhuseinova-msft:
> have we gotten any response regarding my comment in design doc ("CallId support for UI library (CSS)")? I feel like being able to access call id on lobby view is a quite important use case for this.

@jpeng-ms call id can be accessed when user is in lobby (More button is added to Control bar instead of InfoHeaderView)

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/Factories/CompositeViewModelFactory.swift
vhuseinova-msft: A new task for reviewing VMs for Drawers cell will be added

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/ViewComponents/SharingActivity/DiagnosticsSharingActivityViewModel.swift
vhuseinova-msft: The current approach is fine, agreed with Pavel and Travis

JoshuaLai (approved):

emlynmac (approved):

PR #422, opened 2022-11-02
[Feature] Wired Background/Foreground Action to Redux

jpeng-ms:
## Purpose

important thing before reviewing this PR:

the background/foreground state used in chat composite has nothing to do with application lifecycle. These states are strictly limited to chat composite, meaning:

- chat is in foreground when its UI shown
- chat is in background when it goes to headless mode (when its UI hidden)

This story is only to wire actions to redux. There are still pending discussions with Android on series of actions/tasks that should be done when chat goes to a particular mode.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## How to Test
run the unit tests and make sure they all passed.

## What to Check
N/A

## Other Information
N/A


AzureCommunicationUI/sdk/common/BaseComposite/Sources/Presentation/Manager/ChatLifeCycleManager.swift
```diff
@@ -0,0 +1,50 @@
...
+        switch state.lifeCycleState.currentStatus {
+        case .background:
+            enterBackground()
+        case .foreground:
+            enterForeground()
+        }
+    }
+
+    func enterBackground() {
+        logger.debug("Composite enters background mode")
```
jpeng-ms, line 42: there are pending discussions remaining on what we should be doing when chat goes to background/foreground state.

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/ChatComposite.swift
```diff
@@ -153,21 +148,23 @@ public class ChatComposite {
         self.dependencyContainer = nil
     }
 
-    private func makeContainerUIHostingController(router: NavigationRouter,
```
jpeng-ms, line 156: changing this since swift lint was complaining that parameter count has reached the limit of 6 after introducing store

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/SwiftUI/Container/ContainerView.swift
```diff
@@ -21,6 +22,8 @@ struct ContainerView: View {
             }
         }
         .environment(\.layoutDirection, isRightToLeft ? .rightToLeft : .leftToRight)
+        .onAppear { store.dispatch(action: .lifecycleAction(.foregroundEntered)) }
+        .onDisappear { store.dispatch(action: .lifecycleAction(.backgroundEntered)) }
```
jpeng-ms, line 26: same logic as Android: https://github.com/Azure/communication-ui-library-android/pull/486

PR #429, opened 2022-11-03
[Feature] Foreground/Background Action Handling

jpeng-ms:
## Purpose
PR on Android side: https://github.com/Azure/communication-ui-library-android/pull/486

basically the parity between both platform is to:
- use container view's lifecycle to decide foreground/background
- clean up view model when app goes to background

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## How to Test
N/A
testing not available until headless mode is supported

## What to Check
N/A
no unit tests available because the trigger point is from view init/deinit.

## Other Information
N/A

### DIFF
--- a/AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Factories/CompositeViewModelFactory.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Factories/CompositeViewModelFactory.swift
@@ -8,6 +8,7 @@ import FluentUI
 
 protocol CompositeViewModelFactoryProtocol {
     // MARK: CompositeViewModels
+    @discardableResult
     func getChatViewModel() -> ChatViewModel
 
     // MARK: ComponentViewModels
@@ -23,6 +24,8 @@ protocol CompositeViewModelFactoryProtocol {
                                   chatState: ChatState) -> MessageListViewModel
     func makeBottomBarViewModel(dispatch: @escaping ActionDispatch) -> BottomBarViewModel
     func makeTypingParticipantsViewModel() -> TypingParticipantsViewModel
+    func handleAppStatusChange(_ status: AppStatus)
+    func destroyChatViewModel()
 }
 
 class CompositeViewModelFactory: CompositeViewModelFactoryProtocol {
@@ -49,6 +52,7 @@ class CompositeViewModelFactory: CompositeViewModelFactoryProtocol {
     }
 
     // MARK: CompositeViewModels
+    @discardableResult
     func getChatViewModel() -> ChatViewModel {
         guard let viewModel = self.chatViewModel else {
             let viewModel = ChatViewModel(compositeViewModelFactory: self,
@@ -97,5 +101,15 @@ class CompositeViewModelFactory: CompositeViewModelFactoryProtocol {
     func makeTypingParticipantsViewModel() -> TypingParticipantsViewModel {
         TypingParticipantsViewModel(logger: logger,
                                     localizationProvider: localizationProvider)
+    }
+
+    // MARK: View Model Life Cycle Handling
+    func destroyChatViewModel() {
+        chatViewModel = nil
+    }
+
+    func handleAppStatusChange(_ status: AppStatus) {
+        store.dispatch(action: .lifecycleAction(status == .background ?
+            .backgroundEntered : .foregroundEntered))
     }
 }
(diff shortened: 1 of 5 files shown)


AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Factories/CompositeViewModelFactory.swift
```diff
@@ -8,6 +8,7 @@ import FluentUI
 
 protocol CompositeViewModelFactoryProtocol {
     // MARK: CompositeViewModels
+    @discardableResult
```
emlynmac, line 11: Why would you want to call a get method, without using the thing you asked for?

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Factories/CompositeViewModelFactory.swift
```diff
@@ -49,6 +52,7 @@ class CompositeViewModelFactory: CompositeViewModelFactoryProtocol {
     }
 
     // MARK: CompositeViewModels
+    @discardableResult
```
emlynmac, line 55: This seems like it should not be here.

emlynmac (commented):
Can you clarify why we need to do this?

My understanding is that the lifecycle middleware will emit actions for foreground / backgrounding of the app. These actions should then cause state updates in the redux store, which will then trigger view model updates.

Why does this need to be short-circuited here?

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Factories/CompositeViewModelFactory.swift
jpeng-ms: this is to make sure the instance of chat view model can be recreated when composite switches from headless mode:
```
 guard let viewModel = self.chatViewModel else {
     let viewModel = ChatViewModel(compositeViewModelFactory: self,
                                  logger: logger,
                                  store: store)
    self.chatViewModel = viewModel
    return viewModel
}
return viewModel
```

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Factories/CompositeViewModelFactory.swift
jpeng-ms: to get rid of warnings like these: 
image

jpeng-ms:
> 

in additional to events from iOS, chat composite's background and foreground state also relies on the view. For example, when Contoso calls public API to switch composite to headless mode, the composite is considered in background mode even though the contoso app is still on foreground

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Factories/CompositeViewModelFactory.swift
jpeng-ms: oh wait, I see what you saying now, updated

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Factories/CompositeViewModelFactory.swift
emlynmac: > > Can you clarify why we need to do this?
> 
> > My understanding is that the lifecycle middleware will emit actions for foreground / backgrounding of the app. These actions should then cause state updates in the redux store, which will then trigger view model updates.
> 
> > Why does this need to be short-circuited here?
> 
> in additional to events from iOS, chat composite's background and foreground state also relies on the view. For example, when Contoso calls public API to switch composite to headless mode, the composite is considered in background mode even though the contoso app is still on foreground

This sounds like two different sets of concerns that should be separated out.

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Factories/CompositeViewModelFactory.swift
emlynmac: A getter should not be `@discardable`.
It goes against the point of the method. If you're calling it somewhere and not using it, then I'd question the why; something else needs to be resolved there.

jpeng-ms:
after a discussion with @emlynmac offline, I am closing this PR since there's a major changes planned for how composite works in terms of its base layer logic, which would makes the logic of this PR no longer applicable.

PR #432, opened 2022-11-03
[Task] handle local user removal in redux layer, PR 1 of 3

jpeng-ms:
## Purpose
This PR handles local user removal case in redux layer. Upcoming PRs would update UI + more unit tests

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## How to Test
run added unit tests and make sure they all passed

## What to Check
N/A
## Other Information

n/a


AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Redux/Middleware/RepositoryMiddlewareHandler.swift
```diff
@@ -143,8 +144,13 @@ class RepositoryMiddlewareHandler: RepositoryMiddlewareHandling {
     }
 
     func participantRemovedMessage(participants: [ParticipantInfoModel],
+                                   localUser: ParticipantInfoModel?,
                                    dispatch: @escaping ActionDispatch) -> Task<Void, Never> {
         Task {
+            if let localUser = localUser, participants.contains(where: { $0.id == localUser.id }) {
+                dispatch(.participantsAction(.localParticipantRemoved))
+                return
+            }
```
jimchou-dev, line 153: Would it be more appropriate to have this in ChatServiceEventHandler instead?

jpeng-ms: yes it would be but it does not have references to local user id, since it's being stored in chat state currently

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Redux/Middleware/RepositoryMiddlewareHandler.swift
```diff
@@ -41,6 +41,7 @@ protocol RepositoryMiddlewareHandling {
                                  dispatch: @escaping ActionDispatch) -> Task<Void, Never>
     @discardableResult
     func participantRemovedMessage(participants: [ParticipantInfoModel],
+                                   localUser: ParticipantInfoModel?,
```
JoshuaLai, line 44: Is there a reason this needs to be optional? Seems odd since when we init the chat client we need all the local user information

jpeng-ms: good question, this is due to how it's stored as optional in chat state right now: 
image
should we change it in chat state since it's required to have a chat client?

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Manager/EventManager.swift
```diff
@@ -0,0 +1,36 @@
...
+
+    private func receive(_ state: AppState) {
+        // MARK: Handle Local User Removal
+        let participantState = state.participantsState
+        if participantState.localParticipantStatus == .removed,
+            let handler = eventsHandler.onLocalUserRemoved {
+            handler()
+        }
+    }
+
```
jpeng-ms, line 35: this manager will be used to trigger the following handles as well: 

```
        public var onRemoteParticipantJoined: (([CommunicationIdentifier]) -> Void)?
        /// Closure to execute when Chat Composite UI is hidden and receive new message
        public var onUnreadMessagesCountChanged: ((Int) -> Void)?
        /// Closure to execute when Chat Composite UI is hidden and receive new message
        public var onNewMessageReceived: ((ChatMessageModel) -> Void)?
```

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Redux/Action/ParticipantsAction.swift
```diff
@@ -18,8 +18,8 @@ enum ParticipantsAction: Equatable {
 
     case participantsAdded(participants: [ParticipantInfoModel])
     case participantsRemoved(participants: [ParticipantInfoModel])
+    case localParticipantRemoved
```
JoshuaLai, line 21: Do we need a brand new action for this? 
Can we leverage the existing participantRemoved action, and the argument inside would just be the local users `ParticipnatInfoModel`?

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/ChatComposite.swift
```diff
@@ -23,6 +23,8 @@ public class ChatComposite {
         public var onUnreadMessagesCountChanged: ((Int) -> Void)?
         /// Closure to execute when Chat Composite UI is hidden and receive new message
         public var onNewMessageReceived: ((ChatMessageModel) -> Void)?
+        /// Closure to execute when local participant removed from the chat thread
+        public var onLocalUserRemoved: (() -> Void)?
```
JoshuaLai, line 27: This api should also include the `CommunicationIdentifier` argument as well. If we have it for remote participants we should have design parity and have it for local as well. I am almost think it makes more sense to just have one api that is `participantRemoved` and returns an identifier and developers can determine for themselves whether its a remote or local. Having developer subscribe to 2 events for the same thing is not a pleasant developer experience IMO

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Redux/Middleware/RepositoryMiddlewareHandler.swift
```diff
@@ -144,8 +145,13 @@ class RepositoryMiddlewareHandler: RepositoryMiddlewareHandling {
     }
 
     func participantRemovedMessage(participants: [ParticipantInfoModel],
+                                   localUser: ParticipantInfoModel,
                                    dispatch: @escaping ActionDispatch) -> Task<Void, Never> {
         Task {
+            if participants.contains(where: { $0.id == localUser.id }) {
+                dispatch(.participantsAction(.localParticipantRemoved))
+                return
```
JoshuaLai, line 153: IMO this feels logic feels out of place. Each one of these tasks should be responsible for one thing, but now we are introducing this short circuit logic here. And this kind of ties in with my comment on why we should try to leverage the existing participant removed action

jpeng-ms:
per discussion with @JoshuaLai we are breaking this story into 2 parts - UI and public API. Since public APIs are not part of public preview so I am closing this PR.

PR #436, opened 2022-11-07
[Feature] Update to Fluent 0.8.8

emlynmac:
## Purpose
Update the Fluent UI dependency to 0.8.8

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
Pull code, build demo app and run; ensure no changes to UI

## What to Check
All UI functionality remains the same


palatter (dismissed):
Looks good. Will test during standup.

jpeng-ms:
thanks, one thing I noticed is that the drawer background colour seems to be inconsistent with cell background:

image
this applies to participant list and end call drawer as well

emlynmac:
@jpeng-ms 
Looks like the internal layout of the cell isn't taking into account the height of the text. We might need to raise an issue with Fluent for the calculation of the height of the cell; it seems like this isn't dynamic here.

emlynmac:
> thanks, would you mind to create mini-project that shows this is an issue on their end? I believe they would ask us to do this anyways for bug report

It's reproducible with the general way we're using it so I've created that demo app and raised the issue:
https://github.com/microsoft/fluentui-apple/issues/1351

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/FluentUI/Wrapper/CompositeParticipantsListCell.swift
```diff
@@ -38,7 +38,7 @@ class CompositeParticipantsListCell: TableViewCell {
                                                          onHoldString: viewModel.getOnHoldString(),
                                                          isMuted: viewModel.isMuted)
         setup(title: viewModel.getCellDisplayName(with: participantViewData),
-              customView: avatarView,
+              customView: avatar,
```
jpeng-ms, line 41: shouldn't this be ```avatar.view``` based on line 24 in the original file?

emlynmac: Not anymore; part of the API changes.

Leah-Xia-Microsoft (dismissed):

jpeng-ms:
hi so I was testing on iPad, it seems the camera view would get cutoff when it goes to Split View. Do you know if this is due to Fluent UI or this is a bug in our app?
Simulator Screen Shot - iPad (9th generation) - 2022-11-10 at 11 03 43

Simulator Screen Shot - iPad (9th generation) - 2022-11-10 at 11 04 38



Simulator Screen Shot - iPad (9th generation) - 2022-11-10 at 11 03 50

jpeng-ms:
also the on hold button doesn't seem to be fixed, I can still reproduce it:

IMG_0108

emlynmac:
> was testing on iPad, it seems the camera view would get cutoff when it goes to Split View. Do you know if this is due to Fluent UI or this is a bug in our app?

Is this something that exists in the regular app, or only with this updated fluent UI version?

emlynmac:
> also the on hold button doesn't seem to be fixed, I can still reproduce it:
> 
> IMG_0108

@jpeng-ms This issue seems to be due to `.fixedSize()` in `LobbyOverlayView.swift:45` pinning the size of the button incorrectly.

jpeng-ms:
tbh I don't think it's an issue on our side though, It looks fine on our test flight build: 

IMG_0110

jpeng-ms (approved):
tested, looks good!

jpeng-ms:
hi so I tested iPad running develop, I can't reproduce the video bug, so I guess there's something wrong with Fluent UI even though I am not sure if we used any components for the video view: 
image

palatter (approved):
Looks good. Tested on iPhone 14 pro max

jpeng-ms:
hey was the Split View bug an issue with simulator? were you able to reproduce it on real devices?

PR #437, opened 2022-11-07
[Spike] ChatComposite api with View and ViewController

jimchou-dev:
## Purpose

* Refactoring public API to have View and ViewController exposed to Contoso
* Naming is not finalized

```
[ ] Yes
[ x ] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...

### DIFF
--- a/AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/ChatUIKitDemoViewController.swift
+++ b/AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/ChatUIKitDemoViewController.swift
@@ -1,0 +1,620 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+import Combine
+import SwiftUI
+import AzureCommunicationUIChat
+import AzureCommunicationChat
+import AzureCommunicationCommon
+
+class ChatUIKitDemoViewController: UIViewController {
+
+    enum LayoutConstants {
+        static let verticalSpacing: CGFloat = 8.0
+        static let stackViewSpacingPortrait: CGFloat = 18.0
+        static let stackViewSpacingLandscape: CGFloat = 12.0
+        static let buttonHorizontalInset: CGFloat = 20.0
+        static let buttonVerticalInset: CGFloat = 10.0
+    }
+
+    private var selectedAcsTokenType: ACSTokenType = .token
+    private var acsTokenUrlTextField: UITextField!
+    private var acsTokenTextField: UITextField!
+
+    private var displayNameTextField: UITextField!
+    private var userIdTextField: UITextField!
+    private var endpointUrlTextField: UITextField!
+    private var selectedMeetingType: ChatType = .groupChat
+    private var chatThreadIdTextField: UITextField!
+    private var teamsMeetingTextField: UITextField!
+    private var settingsButton: UIButton!
+    private var startExperienceButton: UIButton!
+    private var stopButton: UIButton!
+    private var acsTokenTypeSegmentedControl: UISegmentedControl!
+    private var meetingTypeSegmentedControl: UISegmentedControl!
+    private var stackView: UIStackView!
+    private var titleLabel: UILabel!
+    private var titleLabelConstraint: NSLayoutConstraint!
+
+    // The space needed to fill the top part of the stack view,
+    // in order to make the stackview content centered
+    private var spaceToFullInStackView: CGFloat?
+    private var userIsEditing: Bool = false
+    private var isKeyboardShowing: Bool = false
+
+    private var cancellable = Set<AnyCancellable>()
+    private var envConfigSubject: EnvConfigSubject
+    var chatComposite: ChatComposite?
+
+    private lazy var contentView: UIView = {
+        let view = UIView()
+        view.translatesAutoresizingMaskIntoConstraints = false
+        return view
+    }()
+
+    private lazy var scrollView: UIScrollView = {
+        let view = UIScrollView()
+        view.translatesAutoresizingMaskIntoConstraints = false
+        view.showsVerticalScrollIndicator = false
+        view.showsHorizontalScrollIndicator = false
+        return view
+    }()
+
+    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        super.traitCollectionDidChange(previousTraitCollection)
+        updateUIBasedOnUserInterfaceStyle()
+
+        if UIDevice.current.orientation.isPortrait {
+            stackView.spacing = LayoutConstants.stackViewSpacingPortrait
+            titleLabelConstraint.constant = 32
+        } else if UIDevice.current.orientation.isLandscape {
+            stackView.spacing = LayoutConstants.stackViewSpacingLandscape
+            titleLabelConstraint.constant = 16.0
+        }
+    }
+
+    init(envConfigSubject: EnvConfigSubject) {
+        self.envConfigSubject = envConfigSubject
+        super.init(nibName: nil, bundle: nil)
+        self.combineEnvConfigSubject()
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        setupUI()
+        registerNotifications()
+    }
+
+    override func viewWillLayoutSubviews() {
+        super.viewWillLayoutSubviews()
+        guard !userIsEditing else {
+            return
+        }
+        scrollView.setNeedsLayout()
+        scrollView.layoutIfNeeded()
+        let emptySpace = stackView.customSpacing(after: stackView.arrangedSubviews.first!)
+        let spaceToFill = (scrollView.frame.height - (stackView.frame.height - emptySpace)) / 2
+        stackView.setCustomSpacing(spaceToFill + LayoutConstants.verticalSpacing,
+                                   after: stackView.arrangedSubviews.first!)
+    }
+
+    func combineEnvConfigSubject() {
+        envConfigSubject.objectWillChange
+            .throttle(for: 0.5, scheduler: DispatchQueue.main, latest: true).sink(receiveValue: { [weak self] _ in
+                self?.updateFromEnvConfig()
+            }).store(in: &cancellable)
+    }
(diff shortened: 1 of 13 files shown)


AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/ChatComposite.swift
```diff
@@ -29,41 +27,40 @@ public class ChatComposite {
...
     private var lifeCycleManager: LifeCycleManagerProtocol?
     private var compositeManager: CompositeManagerProtocol?
 
     /// Create an instance of ChatComposite with options.
     /// - Parameter options: The ChatCompositeOptions used to configure the experience.
-    public init(withOptions options: ChatCompositeOptions? = nil) {
-        events = Events()
-        themeOptions = options?.themeOptions
-        localizationOptions = options?.localizationOptions
+    public init(withOptions options: ChatCompositeOptions? = nil,
```
emlynmac, line 40: Usually, we want the non-defaulted parameters first

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/ChatComposite.swift
```diff
@@ -78,42 +75,6 @@ public class ChatComposite {
...
-
-    /// Get Chat Composite SwiftUI view.
-    /// - Returns: Chat Composite view
-    public func getCompositeView() throws -> some View {
-        throw ChatCompositeError(code: ChatCompositeErrorCode.showComposite)
-        // stub: to be implemented
-        return Group {}
-    }
-
     private func launch(_ chatConfiguration: ChatConfiguration,
```
emlynmac, line 78: Do we still need this?

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/ChatSwiftUIView.swift
```diff
@@ -0,0 +1,38 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import SwiftUI
+
+public struct ChatSwiftUIView: View {
```
emlynmac, line 8: `ChatView` ?

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/ChatSwiftUIView.swift
```diff
@@ -0,0 +1,38 @@
...
+
+    let router: NavigationRouter
+    let logger: Logger
+    let viewFactory: CompositeViewFactoryProtocol
+    let isRightToLeft: Bool
+
+    public init(with chatComposite: ChatComposite) {
+        self.chatComposite = chatComposite
+
+        self.router = self.chatComposite.dependencyContainer!.resolve()
```
emlynmac, line 20: It would be way nicer to depend upon only those things specifically, maybe through extending the Environment with the requisite objects for router, logger etc.

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/ChatUIKitViewController.swift
```diff
@@ -0,0 +1,40 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+import Combine
+import SwiftUI
+
+public class ChatUIKitViewController: UIViewController {
```
emlynmac, line 10: `ChatViewController` ?

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/ChatSwiftUIView.swift
jimchou-dev: There is already an internal class called ChatView, any suggestion what to rename the internal one to? (and associated viewModel)

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/ChatUIKitViewController.swift
```diff
@@ -0,0 +1,40 @@
...
+
+    private var chatComposite: ChatComposite
+    private var dependencyContainer: DependencyContainer
+    var chatView: UIHostingController<ContainerView>!
+
+    public init(with chatComposite: ChatComposite) {
+
+        self.chatComposite = chatComposite
+        self.dependencyContainer = self.chatComposite.dependencyContainer!
```
JoshuaLai, line 19: I think with the new approach dependency container can be required in the chat composite (if we move it into the init of chat composite) and we should be able to get around this force unwrap here

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/ChatUIKitDemoViewController.swift
```diff
@@ -0,0 +1,620 @@
...
+        startExperienceButton.backgroundColor = .systemGray3
+
+        let link = self.getMeetingLink()
+        Task { @MainActor in
+            if self.chatComposite == nil {
+                await self.startExperience(with: link)
+            }
+            guard let chatComposite = self.chatComposite else {
+                return
+            }
```
JoshuaLai, line 296: Nitpick: I think we can just create the chat composite at this point right? And chain the `.connect()` here as well?

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/ChatComposite.swift
```diff
@@ -29,41 +27,40 @@ public class ChatComposite {
...
-                credential: remoteOptions.credential,
-                displayName: remoteOptions.displayName)
-            launch(chatConfiguration,
-                   localOptions: localOptions)
-        } catch let error {
-            print("Failed to launch, reason: \(error.localizedDescription)")
-        }
+    /// Start connection to the chat composite to Azure Communication Service.
+    public func connect() {
+        compositeManager?.start()
```
JoshuaLai, line 63: Is it possible for this optional to ever fail?

JoshuaLai: I ask cause when I read this method, it tells me that compositeManager can potentially be nil, in which case start may never be called.

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/ChatComposite.swift
jpeng-ms: Sorry what's chat composite options? If chat options are about theming, style, etc. can we name it as chat settings to differentiate it? otherwise, it seems to me that remote and local can be properties in ChatCompositeOptions since they are all "options" and it looks very redundant this way

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/ChatSwiftUIView.swift
jpeng-ms: ```chatBaseView``` ?

jpeng-ms (approved):

jpeng-ms:
sorry approve it by mistake

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/ChatComposite.swift
jimchou-dev: ChatCompositeOptions its for localization/theming.

When we were developing CallingComposite there were mixture of `settings`, `configs`, `options` and we think it would be confusing for Contoso to distinguish what is the difference between them. So we decide to be consistent and only use the term `options`

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/ChatUIKitDemoViewController.swift
jimchou-dev: Yes but having the parameters in startExperience() to create the chatComposite. And also the chatComposite is optional so there is a separate button that can set it to nil for clean up the composite

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Navigation/NavigationRouter.swift
```diff
@@ -42,10 +42,10 @@ class NavigationRouter: ObservableObject {
...
             dismiss()
-            guard let didNavigateBack = eventsHandler.onNavigateBack else {
-                return
-            }
-            didNavigateBack()
+//            guard let didNavigateBack = eventsHandler.onNavigateBack else {
+//                return
+//            }
+//            didNavigateBack()
         case .exit:
```
jpeng-ms, line 49: do we still need exit here? Is this for call w/ chat?

jimchou-dev: Right now NavigationRouter is technically not needed as we will always be showing `chatView` only, but in the future when we do have a full composite we may need to use it for dismissing the composite.

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/ChatSwiftUIView.swift
emlynmac: I sent you some links offline.

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Navigation/NavigationRouter.swift
jpeng-ms: if this is the case, I guess maybe we should update the error manager to remove the logic on how we dispatch exit when fatal error is received?

PR #441, opened 2022-11-10
[Feature] Implemented service/redux layer for chat participants

jimchou-dev:
## Purpose

* Implemented service layer and redux layer for chat participant and associated unit tests
* TopBarView now will update participants initial participant count and be updated when participants added/removed from chat thread
* Refactored getParticipantList with async while-loop (thanks to @emlynmac )
  * will have separate PR to refactor other ChatSDKWrapper to be using async method 

```
[ ] Yes
[ x ] No
```

## Pull Request Type

```
[ ] Bugfix
[ x ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...


AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Redux/Middleware/ChatActionHandler.swift
```diff
@@ -143,7 +143,7 @@ class ChatActionHandler: ChatActionHandling {
         Task {
             do {
                 let listOfParticipants = try await chatService.getListOfParticipants()
-                dispatch(.participantsAction(.participantsAdded(participants: listOfParticipants)))
```
JoshuaLai, line 146: Can we sync up with Android on this, right now they are leveraging the participant added action, which IMO makes sense cause they are effectively the same thing, if we are going to create a new action lets make sure we align.  cc @iaulakh

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Service/Chat/Extension/ChatSDKAsyncExtension.swift
```diff
@@ -0,0 +1,39 @@
...
+
+extension PagedCollection {
+    func nextPage() async throws -> [SingleElement] {
+        try await withCheckedThrowingContinuation { continuation in
+            nextPage { result in
+                switch result {
+                case .failure(let error):
+                    continuation.resume(throwing: error)
+                case .success(let pageItems):
+                    continuation.resume(returning: pageItems)
```
JoshuaLai, line 34: Same nitpick comment here

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Service/Chat/ChatSDKWrapper.swift
```diff
@@ -92,23 +71,22 @@ class ChatSDKWrapper: NSObject, ChatSDKWrapperProtocol {
...
         do {
-            return try await withCheckedThrowingContinuation { continuation in
-                chatThreadClient?.listParticipants(completionHandler: { result, _ in
-                    switch result {
-                    case .success(let pagedCollection):
-                        guard let items = pagedCollection.pageItems else {
-                            continuation.resume(returning: [])
-                            return
-                        }
+            let participantsPageSize: Int32 = 200
```
JoshuaLai, line 74: nipick: Do we need to declare this as `int32` could Swift not infer the type?

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Redux/Reducer/ParticipantsReducer.swift
```diff
@@ -20,13 +20,20 @@ extension Reducer where State == ParticipantsState,
         var typingIndicatorTimestamp = participantsState.typingIndicatorUpdatedTimestamp
 
         switch action {
+        case .participantsAction(.fetchListOfParticipantsSuccess(let participants)):
+            var newParticipants: [String: ParticipantInfoModel] = [:]
+            for participant in participants {
+                newParticipants[participant.id] = participant
+            }
+            currentParticipants = newParticipants
```
JoshuaLai, line 28: If we are going with this approach, is it safe to just replace everything? Or should we append to what is already there? I am thinking in case any race conditions with the add action?

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Redux/Middleware/ChatActionHandler.swift
jimchou-dev: Right now when we were using participantAdded action, Patric noticed it triggers and adds a message `X,Y,Z has joined the chat` message to be added to the message repository, and it's kind of odd to have that as those participant are already in the chat.
Will sync with Inderpal

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Redux/Reducer/ParticipantsReducer.swift
```diff
@@ -20,13 +20,20 @@ extension Reducer where State == ParticipantsState,
...
+            typingIndicatorMap = [:]
+            participantsUpdatedTimestamp = currentTimestamp
+            typingIndicatorTimestamp = currentTimestamp
         case .participantsAction(.participantsAdded(let participants)):
```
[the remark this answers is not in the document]
jimchou-dev: Participant action is only adding list of new participants, when initializing it should replace all the participant and reset the state and it should only be called once (or when switching chat thread with single thread support for now using `connect`)

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Redux/Reducer/ParticipantsReducer.swift
jimchou-dev: Initializing participant action should only be called once at the start, or when thread context switches by `connect()` before we implement multiple chat thread support for composite/adapter

Leah-Xia-Microsoft (dismissed):

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/SwiftUI/Chat/ChatViewComponents/TopBarViewModel.swift
```diff
@@ -54,7 +54,10 @@ class TopBarViewModel: ObservableObject {
     }
 
     func update(participantsState: ParticipantsState) {
-        updateNumberOfParticipantsLabel(
-            numberOfParticpants: participantsState.numberOfParticipants)
+        if participantsUpdatedTimestamp != participantsState.participantsUpdatedTimestamp {
```
jpeng-ms, line 57: so participantsUpdatedTimestamp and participantsUpdatedTimestamp are defined in view model and reducer, it seems there would always be few ms difference and essentially making this if check always true?

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Redux/Reducer/ParticipantsReducer.swift
```diff
@@ -32,9 +38,7 @@ extension Reducer where State == ParticipantsState,
                 typingParticipants = typingParticipants.filter { $0.id != participant.id }
                 currentParticipants.removeValue(forKey: participant.id)
             }
-            let state = ParticipantsState(participants: currentParticipants,
-                                          participantsUpdatedTimestamp: participantsUpdatedTimestamp)
-            return state
+            participantsUpdatedTimestamp = Date()
```
jpeng-ms, line 41: can we get rid of the timestamp and use data model to see if view needs to be updated?

for example, if viewModel.participants != state.participants we update the view rather than using timestamp?

Leah-Xia-Microsoft (approved):

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/SwiftUI/Chat/ChatViewComponents/TopBarViewModel.swift
jimchou-dev: The viewModel one is a cache to see if there is a difference, and if the timestamp in participantState (from reducer) is updated then will update the cached timestamp and perform associated updates. So this should not always be true unless the timestamp has been changed in participantState.participantsUpdatedTimestamp

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Redux/Reducer/ParticipantsReducer.swift
jimchou-dev: When comparing only the timestamp is much more efficient than comparing list that it would need to check each item in the list if are equal. This pattern has also been repeatedly used in CallingComposite.

jpeng-ms: right, but it seems like this timestamp was only used for updating numberOfParticipants?

jimchou-dev: Yes for now, as participant is not in scope for Private Preview. However for GA the timestamp will be needed for displaying full participants list and knowing when participants are updated and displayName/avatar may be changing.

jpeng-ms (approved):

PR #450, opened 2022-11-15
[Task] Handling Local User Removed

jpeng-ms:
## Purpose

image

- [ ] Yes
- [X] No

## Pull Request Type

- [ ] Bugfix
- [X] Feature

## Pull request checklist

This PR has considered:
- [ ] Color Theming
- [X] iOS supported versions (all major versions greater than or equal current target deployment version)
- [X] View Data Injection
- [X] Demo App UI/UX update

## How to Test
since there's no way to remove a participant currently from neither web/mobile, you need to hard code the following in ```ChatServiceEventHandler``` to simulate an event from SDK:

```
        DispatchQueue.main.asyncAfter(deadline: .now() + 5, execute: {
              dispatch(.chatAction(.chatMessageLocalUserRemoved))
        })
```

## What to Check

## Other Information

https://user-images.githubusercontent.com/109105353/201750680-7d5338b5-fe79-4398-a04a-e20c57fb2101.mov


palatter (commented):
What happens if you try to join a chat that you have already been removed from?

jpeng-ms:
> What happens if you try to join a chat that you have already been removed from?

good question, I will expect an error of chat joined failed thrown from SDK:
https://github.com/Azure/communication-ui-library-ios/pull/452/files#diff-4ba420b61e3f5ecb08f13910e021733c7b8e231cc10b2e481c541a0a2d5aa2f7R64

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Redux/State/ParticipantsState.swift
```diff
@@ -18,9 +19,16 @@ struct ParticipantsState {
...
+         localParticipantStatus: LocalParticipantStatus = .joined) {
         self.participants = participants
+        self.localParticipantStatus = localParticipantStatus
         self.typingParticipants = typingParticipants
         self.participantsUpdatedTimestamp = participantsUpdatedTimestamp
     }
 }
+
+enum LocalParticipantStatus {
+    case joined
```
jimchou-dev, line 32: Can we change the enum to disconnected/connected/removed? As the new refactored API will have connect() and disconnect() that can be called anytime with or without the UI

jpeng-ms: good point! thanks

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Model/ChatMessageInfoModel.swift
```diff
@@ -25,6 +24,7 @@ struct ChatMessageInfoModel: BaseInfoModel, Identifiable, Equatable, Hashable {
     var createdOn: Iso8601Date
     var editedOn: Iso8601Date?
     var deletedOn: Iso8601Date?
+    var isLocalUser: Bool
```
JoshuaLai, line 27: Can we not make this a computed property? 
```swift
var isLocalUser: Bool {
// if id is in participants list == true we know local user is effected in some way
return false
}
```
I'm not entirely clear if this is how id works / should be used

JoshuaLai: If this doesn't make sense then the way you have it laid out here is fine cause it aligns with Android

jpeng-ms: so making this a computed value will mean it would be a get only property. If this is the case, then isLocalUser would always be false, then how do we display system message then? We do want this property to be true when we want to show system message for local user removed right?

JoshuaLai: Why would it always be false? You would have logic in there to compute at runtime whether it should true or false right

jpeng-ms: I see, then we need to bring local user id to this model. Would it be a good idea? I am open to this

jimchou-dev: Yeah it would have a `localUserId: String?` also and based on the messageType to do logic for computed property isLocalUser

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Manager/MessageRepositoryManager.swift
```diff
@@ -92,6 +93,15 @@ class MessageRepositoryManager: MessageRepositoryManagerProtocol {
         messages.append(topicUpdatedSystemMessage)
     }
 
+    func addLocalUserRemovedMessage() {
+        let localUserRemovedSystemMessage = ChatMessageInfoModel(
+            type: .participantsRemoved,
+            createdOn: Iso8601Date(),
+            isLocalUser: true
+        )
+        messages.append(localUserRemovedSystemMessage)
```
jimchou-dev, line 102: Would this also apply for when user is removed from chat thread due to chat thread being deleted?

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/SwiftUI/Chat/ChatViewComponents/BottomBarView.swift
```diff
@@ -55,4 +61,9 @@ struct BottomBarView: View {
         IconButton(viewModel: viewModel.sendButtonViewModel)
             .flipsForRightToLeftLayoutDirection(true)
     }
+
+    var localParticipantInfoBanner: some View {
```
jimchou-dev, line 65: Would this be a generic banner? or only for participant removed? If so probably use a more descriptive name here

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/SwiftUI/Chat/ChatViewComponents/BottomBarViewModel.swift
```diff
@@ -54,4 +56,8 @@ class BottomBarViewModel: ObservableObject {
             lastTypingIndicatorSendTimestamp = Date()
         }
     }
+
+    func update(chatState: ChatState) {
+        isLocalUserRemoved = chatState.isLocalUserRemovedFromChat
+    }
```
jimchou-dev, line 62: You may want to check if the value is different before setting it, otherwise when it's `false` and set the published variable to `false` again will trigger a re-render, not good for performance.

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Service/Chat/Extension/ChatParticipantExtension.swift
```diff
@@ -8,19 +8,21 @@ import AzureCore
 import AzureCommunicationChat
 
 extension ChatParticipant {
-    func toParticipantInfoModel() -> ParticipantInfoModel {
+    func toParticipantInfoModel(_ localParticipantID: String = "") -> ParticipantInfoModel {
```
jimchou-dev, line 11: nitpick: personally prefer `localParticipantId` (for consistency as other places in the code use lower case `Id` already), similar for other occurance of localParticipantID

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Model/ChatMessageInfoModel.swift
jpeng-ms: actually this won't work because when we adding localUserRemovedSystemMessage to message repo, the list would be empty since we didn't have data from middleware in the first place. So this computed variable would be false. 

see MessageRepositoryManager.addLocalUserRemovedMessage()

jpeng-ms: the list would only be populated when we receive addParticipantAdded or addParticipantRemoved. 

Actually this gave me an idea, can we use this list as the logic to render system message? 

image

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Manager/MessageRepositoryManager.swift
jpeng-ms: no it won't. When chat thread is deleted, the error reducer listens to the action, then reducer an error state of fatal error and pass to error manager to call error handlers from Contoso

jpeng-ms: so it's up to SDK, if they end participants removed action + chat thread deleted action to us, then we would have both behaviours

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/SwiftUI/Chat/ChatViewComponents/BottomBarView.swift
jpeng-ms: only for local participant removed, we don't have any other use case that needs this label

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Service/Chat/Extension/ChatParticipantExtension.swift
jpeng-ms: thanks, updated!

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/SwiftUI/Chat/ChatViewComponents/BottomBarViewModel.swift
jpeng-ms: thanks, updated!

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/SwiftUI/Chat/ChatViewComponents/BottomBarView.swift
jimchou-dev: Could we rename this to something more descriptive?

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Manager/MessageRepositoryManager.swift
jimchou-dev: > no it won't. When chat thread is deleted, the error reducer listens to the action, then reducer an error state of fatal error and pass to error manager to call error handlers from Contoso
This behaviour was implemented when the design was still a full screen chat experience, and now with the new API fatal error would not exit the composite.

Similar to the participantRemoved (for local user's identifier) and chatThreadDeleted trouter event, these both will be the last event will be received from chatSDK, can you confirm with Jorge/Alex on how the app should behave in the case when a chat thread is deleted?

jpeng-ms: sorry I was saying that it would call error handlers from Contoso, I think this is the expected behaviour which aligns with the proposal for local user removed. But I will check it for you

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/SwiftUI/Chat/ChatViewComponents/BottomBarView.swift
jpeng-ms: I thought this is descriptive enough? Do you have any suggestions?

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Model/ChatMessageInfoModel.swift
jpeng-ms: on second thoughts, this would probably cause confusion and not necessary. I am keeping the bool for now

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Manager/MessageRepositoryManager.swift
jpeng-ms: pinged PM and UX

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/SwiftUI/Chat/ChatViewComponents/BottomBarView.swift
JoshuaLai: Might be more clear if we just add `removed`
```suggestion
    var localParticipantRemovedBanner: some View {
```

jpeng-ms: updated thanks

Leah-Xia-Microsoft:
The color and sizing or the icon doesn't seem quite right

jpeng-ms:
> The color and sizing or the icon doesn't seem quite right

thanks, it added after I rebased my branch. Please open an issue for this.

jimchou-dev (approved):
LGTM

JoshuaLai (approved):

PR #455, opened 2022-11-17
[Feature] Implemented service layer and redux layer for editing/deleting message

jimchou-dev:
## Purpose

* Implemented service layer and redux layer for editing/deleting message
* Chat action include previous content incase of failure, may change depend on specs when implementing presentation layer during GA 

- [ ] Yes
- [ x ] No

## Pull Request Type

- [ ] Bugfix
- [ x ] Feature

## Pull request checklist

This PR has considered:
- [ ] Color Theming
- [ ] View Data Injection
- [ ] Demo App UI/UX update

## How to Test

* Open...
* Click on...

## What to Check


vhuseinova-msft (approved):

jpeng-ms (approved):

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Manager/MessageRepositoryManager.swift
```diff
@@ -62,6 +66,27 @@ class MessageRepositoryManager: MessageRepositoryManagerProtocol {
         messages.append(message)
     }
 
+    func editMessage(messageId: String, content: String) {
+        if let index = messages.firstIndex(where: {
```
kevinyulu, line 70: minor: the pattern of getting the first index kinda gets duplicated in the functions and would be nice to create a private function to find index for the message.

kevinyulu (approved):

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Manager/MessageRepositoryManager.swift
jimchou-dev: It is a built in function already, when extracting we could still be passing in a different `where` condition, so extracting it to another function or extension the code would not be more clean than it is right now. Will leave it as is but thanks for the suggestion =)

PR #458, opened 2022-11-18
[Refactor] Refactor and renaming of public api for chat

jimchou-dev:
## Purpose

* Refactor public api for chat
* Rename composite to adapter for chat
* Refactor demo app to have separate UIKit and SwiftUI views for chat

- [ ] Yes
- [ x ] No

## Pull Request Type

- [ ] Bugfix
- [ ] Feature
- [ x ] Refactoring (no functional changes, no api changes)

## Pull request checklist

This PR has considered:
- [ ] Color Theming
- [ ] View Data Injection
- [ ] Demo App UI/UX update

## How to Test

* Open...
* Click on...

## What to Check


AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/ChatAdapter.swift
```diff
@@ -0,0 +1,150 @@
...
+    private var errorManager: ErrorManagerProtocol?
+    private var lifeCycleManager: LifeCycleManagerProtocol?
+    private var compositeManager: CompositeManagerProtocol?
+
+    /// Create an instance of ChatComposite with options.
+    public init(communicationIdentifier: CommunicationIdentifier,
+                credential: CommunicationTokenCredential,
+                endpoint: String,
+                displayName: String? = nil) {
+        self.chatConfiguration = ChatConfiguration(
```
JoshuaLai, line 41: I wonder if we just make `ChatConfiguration` public and instead of having all these arguments just have developers pass in this object

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/ChatAdapter.swift
```diff
@@ -0,0 +1,150 @@
...
+        self.dependencyContainer = depCont
+        self.logger = depCont.resolve() as Logger
+        depCont.registerDependencies(self.chatConfiguration,
+                                     chatCompositeEventsHandler: events,
+                                     connectEventHandler: nil)
+        self.errorManager = depCont.resolve() as ErrorManagerProtocol
+        self.lifeCycleManager = depCont.resolve() as LifeCycleManagerProtocol
+        self.compositeManager = depCont.resolve() as CompositeManagerProtocol
+
+        compositeManager?.start()
```
JoshuaLai, line 86: Nitpick: I think this can just be moved into helper that both connect methods can call

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/ChatAdapter.swift
```diff
@@ -0,0 +1,150 @@
...
+        self.events = Events()
+    }
+
+    deinit {
+        logger?.debug("Composite deallocated")
+    }
+
+    /// Start connection to the chat composite to Azure Communication Service.
+    public func connect(threadId: String,
+                        completionHandler: ((Result<Void, ChatCompositeError>) -> Void)? = nil) {
```
JoshuaLai, line 55: Is this completion just being called atm?

jimchou-dev: Yes the SwiftUI demo app is using the completionHandler to print out in console

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/ChatAdapter.swift
jimchou-dev: Made the async method is calling the sync method

JoshuaLai:
I think this is a fine first draft. But we will need to create tasks to fill in the missing pieces, code clean up and perform memory tests as well.

JoshuaLai (dismissed):

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/ChatSwiftUIDemoView.swift
```diff
@@ -282,10 +223,6 @@ extension CredentialsView {
     }
 
     private func showError(error: ChatCompositeError) {
-        DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
-            self.alertTitle = "Error - \(error.code)"
-            self.alertMessage = error.error?.localizedDescription ?? error.localizedDescription
-            self.isAlertDisplayed = true
-        }
+        print("Error - \(error.code): \(error.error?.localizedDescription ?? error.localizedDescription)")
```
emlynmac, line 226: I guess this needs conversion to a `Logger.error()` call?

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/ChatUIKitDemoViewController.swift
```diff
@@ -0,0 +1,616 @@
...
+//
+
+import UIKit
+import Combine
+import SwiftUI
+import AzureCommunicationUIChat
+import AzureCommunicationChat
+import AzureCommunicationCommon
+
+class ChatUIKitDemoViewController: UIViewController {
```
emlynmac, line 13: `ChatViewController` should be sufficient here?

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/ChatAdapter.swift
emlynmac: The completion handler should not be allowed to be nil. 
Given that connect requires a callback to complete or fail, we must force consumers of this API to implement this method.

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/ChatAdapter.swift
```diff
@@ -0,0 +1,138 @@
...
+        self.errorManager = depCont.resolve() as ErrorManagerProtocol
+        self.lifeCycleManager = depCont.resolve() as LifeCycleManagerProtocol
+        self.compositeManager = depCont.resolve() as CompositeManagerProtocol
+
+        compositeManager?.start()
+    }
+
+    /// Start connection to the chat composite to Azure Communication Service.
+    public func connect(threadId: String) async throws {
+        connect(threadId: threadId, completionHandler: nil)
```
emlynmac, line 74: This needs to be implemented as a wrapper with a continuation.

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/ChatCompositeOptions/ChatConfiguration.swift
```diff
@@ -7,50 +7,23 @@ import AzureCommunicationCommon
 import Foundation
 
 struct ChatConfiguration {
-    let chatThreadId: String
+    var chatThreadId: String
```
emlynmac, line 10: Why does the thread ID need to be a var here?
Perhaps it does not belong in the ChatConfiguration, if it cannot be determined at the same point in the connection set-up?

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/ChatCompositeView.swift
```diff
@@ -0,0 +1,38 @@
...
+
+    let router: NavigationRouter
+    let logger: Logger
+    let viewFactory: CompositeViewFactoryProtocol
+    let isRightToLeft: Bool
+
+    public init(with chatAdapter: ChatAdapter) {
+        self.chatAdapter = chatAdapter
+
+        self.router = self.chatAdapter.dependencyContainer!.resolve()
```
emlynmac, line 20: Forced-unwrapped values must not be used; our linter should be catching this.

emlynmac (dismissed):
Looks like a good start!

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/ChatSwiftUIDemoView.swift
jimchou-dev: The demo app doesn't have logger, we can refactor the demo app to have its logger in the future but will skip that in this PR.

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/ChatCompositeOptions/ChatConfiguration.swift
jimchou-dev: Josh made a similar comment on ChatConfiguration, I'll create a task (ADO 3061412) to refactor this out, I'll create a task for it to refactor it out.

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/ChatAdapter.swift
jimchou-dev: I think from previous discussion with calling there were different terms for options/config/settings etc... and we decide to only be using Options, but we got rid of RemoteOptions, so doesnt that mean we would be re-introducing it but without the threadId/locator?

Also the reason we have chatConfiguration internally is it's collectively passing information from redux layer to service layer in a simple call, but since it does not contain threadId anymore. I'll create a task (ADO 3061412) to refactor this out.

JoshuaLai (approved):

PR #471, opened 2022-11-23
[Bugfix] Hide setup control bar while joining call when video is on

Leah-Xia-Microsoft:
## Purpose

* When video is on, and join call button is tapped waiting to join a call, we should hide the setup control bar
* When video is off, we don't need to hide that bar

- [ ] Yes
- [x] No

## Pull Request Type

- [x] Bugfix
- [ ] Feature

## Pull request checklist

This PR has considered:
- [ ] Color Theming
- [ ] View Data Injection
- [x] Demo App UI/UX update

## How to Test

* Launch the app and tap on Start Experience
* Turn on video
* Tap on Join call button
* Check that `Video on`, `Mic off` and `Speaker` buttons are all hidden

## What to Check

 | 

When video is off, it behaves as before


jimchou-dev (approved):
LGTM

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/Setup/SetupView.swift
```diff
@@ -37,7 +37,9 @@ struct SetupView: View {
                                 PreviewAreaView(viewModel: viewModel.previewAreaViewModel,
                                                 viewManager: viewManager,
                                                 avatarManager: avatarManager)
-                                SetupControlBarView(viewModel: viewModel.setupControlBarViewModel)
+                                if viewModel.shouldShowSetupControlBarView() {
```
vhuseinova-msft, line 40: I'm wondering if it should be consistent behaviour for both cases when camera on and off. So we should hide the controll bar when joining the call.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/Setup/SetupView.swift
```diff
@@ -37,7 +37,9 @@ struct SetupView: View {
...
+                                    SetupControlBarView(viewModel: viewModel.setupControlBarViewModel)
```
vhuseinova-msft, line 41: Also, it may be a good idea to disable buttons instead of hiding the whole `SetupControlBarView`

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/Setup/SetupView.swift
Leah-Xia-Microsoft: > I'm wondering if it should be consistent behaviour for both cases when camera on and off. So we should hide the controll bar when joining the call with camera on or off.

This is our UX's preferred behavior to hide it only when video is on.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/Setup/SetupView.swift
Leah-Xia-Microsoft: > Also, it may be a good idea to disable buttons instead of hiding the whole `SetupControlBarView`

Those buttons are disabled. Disabled buttons are gray and it looks bad when camera is on. Confirmed with our UX and he wants to hide the bar only when camera is on.

vhuseinova-msft (approved):

PR #474, opened 2022-11-24
[Feature] Chat parse Teams html RichText message

jimchou-dev:
## Purpose

* Able to parse and display Teams html RichText message
  * When receiving TrouterEvent message will be of type `.custom("RichText/Html)`
  * When fetching initial/previous list of messages will be of type `.html`
* Able to display emoji and HTML character entity characters (eg. ><!@#$%^...)
* Does not display Teams image or Teams animated emoji, will be stripped out
* To test please use the same token to join a Teams call in a separate CallingComposite and be admitted to the meeting before starting ChatComposite in a different app

- [ ] Yes
- [x] No

## Pull Request Type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

This PR has considered:
- [ ] Color Theming
- [ ] View Data Injection
- [ ] Demo App UI/UX update

## How to Test

* Open...
* Click on...

## What to Check


AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/SwiftUI/Chat/ChatViewComponents/Message/TextMessageView.swift
```diff
@@ -56,7 +56,7 @@ struct TextMessageView: View {
                 timeStamp
                 edited
             }
-            Text(viewModel.message.content ?? "No Content") // Handle nil?
+            Text(viewModel.getContentString())
```
vhuseinova-msft, line 59: I'm wondering if text color or font should be changed when `getContentString` returns "Text not available". So it would be easier for user to distinguish that something is wrong in this case

vhuseinova-msft (dismissed):

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/SwiftUI/Chat/ChatViewComponents/Message/TextMessageView.swift
jimchou-dev: The font/color will be in the html attributes wrapped between the angle brackets `<p style="color: red"> some text</p>` will be stripped out so the text will still be displayed. But for images, span html element that will just be stripped out and become blank message (this is same behaviour for Web's Chat composite as well right now)

It should only display "Text not available" when the content is nil for some reason received from ChatSDK

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/SwiftUI/Chat/ChatViewComponents/Message/TextMessageViewModel.swift
```diff
@@ -55,4 +55,8 @@ class TextMessageViewModel: MessageViewModel {
             return nil
         }
     }
+
+    func getContentString() -> String {
+        return message.content ?? "Text not available"
```
JoshuaLai, line 60: Looks like the content changed from "No Content" to "Text not available". Is this a change that was expected?

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/SwiftUI/Chat/ChatViewComponents/MessageListViewModel.swift
```diff
@@ -167,7 +167,7 @@ class MessageListViewModel: ObservableObject {
         let isConsecutive = message.senderId == lastMessage.senderId
 
         switch type {
-        case .text:
+        case .text, .html, .custom("RichText/Html"):
```
JoshuaLai, line 170: Is there a way to make the string argument inside `custom` case not case sensitive?
What I am thinking is maybe
```swift
case .custom(let content):
   if content.lowercased == "something"
```
Or something along these lines? I am not sure if there is a way to do a lowercased check inside the value?

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/SwiftUI/Chat/ChatViewComponents/Message/TextMessageViewModel.swift
jimchou-dev: I think that might have been from a back merge? But it doesn't really matter, as I spoke wit Alex and he said we should hide the chat bubble if content is nil, separate refactoring Task #3071479 was created.

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/SwiftUI/Chat/ChatViewComponents/MessageListViewModel.swift
jimchou-dev: Sounds good, moved logic to ChatMessageType extension

vhuseinova-msft (dismissed):

JoshuaLai (approved):

PR #478, opened 2022-11-25
[BuildSystem] Add pipeline configuration to the repository

emlynmac:
## Purpose
Migrating the CI process to be managed through the repository.
This is the first step; enabling our CI pipeline.

- [ ] Yes
- [x] No

## Pull Request Type

- [ ] Bugfix
- [ ] Feature
- [x] Other... Please describe: Add CI pipeline to the repository

## Pull request checklist

This PR has considered:
- [ ] Color Theming
- [ ] View Data Injection
- [ ] Demo App UI/UX update

## How to Test
No code changes, check new build job is fired


azure-pipelines.yml
```diff
@@ -0,0 +1,154 @@
...
+# Performs basic testing for PRs and pushes to specific branches
+
+# Run on pushes to these branches
+trigger:
+  branches:
+    include:
+      - main
+      - develop
+      - release/*
+      - feature/*
```
JoshuaLai, line 11: If we are running it on every feature branch should we also run them on `refactor/*` or `bug/*` `defect/*` branches? 
Part of me thinks no cause all that will be triggered again in the develop branch anyways? Is feature in here explicitly for the chat feature branch?

emlynmac: We can if we want. It's not difficult to change so maybe we start with these?

jimchou-dev (commented):
Does it make sense to have the file in the root directory of the repo? Maybe move it to some folder under /scripts for organization?
Also probably can make the filename more specific to the job it's doing as I can see we will have other azure pipelines in the future

JoshuaLai:
> Does it make sense to have the file in the root directory of the repo? Maybe move it to some folder under /scripts for organization? Also probably can make the filename more specific to the job it's doing as I can see we will have other azure pipelines in the future

I agree with this, but we may also want to organize the scripts folder a bit more. I think right now we just have the script to do the version updates, maybe inside scripts we can have a new folder for pipelines, and drop this in there?

emlynmac:
> Does it make sense to have the file in the root directory of the repo? Maybe move it to some folder under /scripts for organization?
> Also probably can make the filename more specific to the job it's doing as I can see we will have other azure pipelines in the future

I expect we'll have only one automated CI pipeline file. 

`scripts` does seem like a natural home for this, but the more standard practice in azure is to have an `eng` folder at the top level and then contain things there. 

eg.
 https://github.com/Azure/communication-ui-library-android/tree/develop/eng
 https://github.com/Azure/azure-sdk-for-ios/tree/main/eng
 
 I'll move the file there for now, and create a task for migrating the other files

vhuseinova-msft (dismissed):

vhuseinova-msft (approved):

jimchou-dev (approved):

PR #479, opened 2022-11-25
Bugfix/spi workaround calling sdk wrapper mock archive issue

kevinyulu:
## Purpose

* The recent introduction of using custom callingSDKWrapper in UITest caused issue in archiving the demo app. This PR includes a workaround for resolving the issue in archiving.

Issue:
Screen Shot 2022-11-23 at 2 17 39 PM

Cause: 
Swift Package Manager currently does not support the use of @testable import in archive, resulting in compiler not able to find reference of the module. The workaround is to expose the custom init function from CallComposite and its underling CallingSDKWrapperProtocol types by using @_spi attribute so that they can be imported by using SPI import. 

- [ ] Yes
- [x] No

## Pull Request Type

- [ ] Bugfix
- [ ] Feature

## Pull request checklist

This PR has considered:
- [ ] Color Theming
- [ ] View Data Injection
- [ ] Demo App UI/UX update

## How to Test

* Open...
* Click on...

## What to Check


AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Utilities/CallCompositeInternalError.swift
```diff
@@ -59,8 +59,9 @@ enum CallCompositeInternalError: Error, Equatable {
     }
 }
 
-extension CallCompositeInternalError {
-    static func == (lhs: CallCompositeInternalError, rhs: CallCompositeInternalError) -> Bool {
+@_spi(CallCompositeUITest) public  extension CallCompositeInternalError {
+    @_spi(CallCompositeUITest) static func == (lhs: CallCompositeInternalError,
```
vhuseinova-msft, line 63: I'm wondering if `public` is missed here

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/DI/DependencyContainer.swift
```diff
@@ -32,11 +32,13 @@ final class DependencyContainer {
 
     func registerDependencies(_ callConfiguration: CallConfiguration,
                               localOptions: LocalOptions?,
-                              callCompositeEventsHandler: CallComposite.Events) {
+                              callCompositeEventsHandler: CallComposite.Events,
+                              withCallingSDKWrapper wrapper: CallingSDKWrapperProtocol? = nil) {
```
vhuseinova-msft, line 36: `with` can be removed as other parameters doesn't include it

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Service/Calling/AzureCommunicationCalling/CallingSDKEventsHandler.swift
```diff
@@ -147,14 +147,14 @@ class CallingSDKEventsHandler: NSObject, CallingSDKEventsHandling {
     }
 }
 
-extension CallingSDKEventsHandler: CallDelegate,
+@_spi(CallCompositeUITest) extension CallingSDKEventsHandler: CallDelegate,
```
vhuseinova-msft, line 150: is`public` missed here?

kevinyulu: this is by swift I think. swift does not allow us to specify access level on extension that declare protocol conformances. 
`'public' modifier cannot be used with extensions that declare protocol conformances`

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Utilities/CallCompositeInternalError.swift
kevinyulu: yeah good call. Actually swift does it for us in this case. adding `public` would have this warning so I removed the public 
`'public' modifier is redundant for operator function declared in a public extension`

emlynmac:
I think given how much needs to be made public, we should probably just ensure that the changes for test can only be used with a debug build.
This would mean wrapping imports with #if DEBUG and corresponding functionality for the tests.

PR #489, opened 2022-11-29
[Task] Add Alert to Demo App

jpeng-ms:
## Purpose
image

- [ ] Yes
- [x] No

## Pull Request Type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

This PR has considered:
- [ ] Color Theming
- [x] iOS supported versions (all major versions greater than or equal current target deployment version)
- [ ] View Data Injection
- [ ] Demo App UI/UX update

## How to Test

* Open...
* Click on...

## What to Check


AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/ChatDemoView.swift
```diff
@@ -29,6 +31,15 @@ struct ChatDemoView: View {
             .navigationTitle("UI Library - Chat Sample")
             .navigationBarTitleDisplayMode(.inline)
         }
+        .alert(isPresented: $isErrorDisplayed) {
```
emlynmac, line 34: How about implementing this through a `ViewModifier` for better reuse?
https://developer.apple.com/documentation/swiftui/viewmodifier

jpeng-ms: good idea, but this is a demo app, so I was wondering if we should make it as straightforward as possible?

vhuseinova-msft (dismissed):

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/ViewModifier/AlertView.swift
```diff
@@ -0,0 +1,21 @@
...
+    var errorMessage: String
+    var onDismiss: (() -> Void)?
+
+    func body(content: Content) -> some View {
+        content.alert(isPresented: $isPresented) {
+                Alert(title: Text("Error"),
+                      message: Text(errorMessage),
+                      dismissButton: .default(Text("Dismiss"), action: onDismiss))
+        }
+    }
```
jimchou-dev, line 20: Can we change this to something more generic as an alert may not always be an error

jpeng-ms: Are you thinking about having this for chat events? If so, I don’t think it will make too much sense?

jpeng-ms: Maybe let me rename this to ErrorView instead

vhuseinova-msft (dismissed):

jimchou-dev (approved):
LGTM

Leah-Xia-Microsoft (approved):

PR #494, opened 2022-11-30
[Refactor] Clarify dependency construction for chat library

emlynmac:
## Purpose
`DependencyContainer.swift` contained at type-erased dictionary of objects, primarily used by the main chat class.
The pattern of being able to `register()` and `resolve()` objects at run-time is convenient for applications where they will be changed at run-time, but this is not the case with our library.

`DependencyContainer.swift` provided a means to do a lot of object construction, which is then handed back to the calling class.

This PR provides a clear set of dependencies and maintains type information. 
As a result, we have
* Reduced complexity in the code
* Reduced the number of test cases as checks are enforced at compile-time
* Clarified object usage in the code at call sites, enabling further refactoring of dependencies

This PR covers only the Chat library; but there are deeper improvements to be made on the calling library side.

- [ ] Yes
- [x] No

## Pull Request Type

- [ ] Bugfix
- [ ] Feature

## Pull request checklist

This PR has considered:
- [ ] Color Theming
- [ ] View Data Injection
- [ ] Demo App UI/UX update

## How to Test
Perform the usual tests; note no breakages
## What to Check
No change to functionality.


palatter:
With the differing structure of chat, is it possible to get rid of the container object here as well?

palatter:
> With the differing structure of chat, is it possible to get rid of the container object here as well?

Thanks, looks good!

palatter (approved):
Looks good.

PR #501, opened 2022-12-01
[Refactor] Clarify dependency construction for the CallComposite

emlynmac:
## Purpose
`DependencyContainer.swift` contained at type-erased dictionary of objects, primarily used by the CallComposite class.
The pattern of being able to `register()` and `resolve()` objects at run-time is convenient for applications where they will be changed at run-time, but this is not the case with our library.

`DependencyContainer.swift` provided a means to do a lot of object construction, which is then handed back to the calling class.

This PR constructs all dependencies for the CallComposite internally and owns those objects.

As a result, we have
* Reduced complexity in the code
* Reduced the number of test cases 
* Maintained type-checking at compile time
* Clarified object usage in the code

The chat code has a similar PR:
https://github.com/Azure/communication-ui-library-ios/pull/494

- [ ] Yes
- [x] No

## Pull Request Type

- [ ] Bugfix
- [ ] Feature


palatter (approved):
This looks good! Thanks for simplifying this!

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Tests/Presentation/Setup/CallCompositeTests.swift
```diff
@@ -0,0 +1,37 @@
...
+class CallCompositeTests: XCTestCase {
+    func test_verify_construction() {
+        let callComposite = CallComposite()
+
+        XCTAssertNotNil(callComposite)
+    }
+
+    func test_verify_launch() throws {
+
+        let sampleToken = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjMyNTAzNjgwMDAwfQ.9i7FNNHHJT8cOzo-yrAUJyBSfJ-tPPk2emcHavOEpWc"
```
JoshuaLai, line 21: I'm not sure its a good idea to check in a token into the repo. Maybe we can inject in the AppConfig

emlynmac: I just moved it. It's in the repo already.

emlynmac: There's nothing in the token; have a look at https://jwt.io to evaluate.

JoshuaLai (approved):

PR #517, opened 2022-12-09
[BuildSystem] Enable code coverage / build quality check

emlynmac:
## Purpose
Re-enable the code coverage check, with 50% code coverage requirement.

We can increase this with more test coverage.

### DIFF
--- a/eng/pipelines/jobs/CI-build-job.yml
+++ b/eng/pipelines/jobs/CI-build-job.yml
@@ -45,6 +45,6 @@ jobs:
         displayName: "Check build quality"
         inputs:
           checkCoverage: true
-          coverageThreshold: 50
+          coverageThreshold: 58
           coverageFailOption: fixed
           coverageType: lines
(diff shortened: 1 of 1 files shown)


eng/pipelines/jobs/CI-build-job.yml
```diff
@@ -43,8 +43,8 @@ jobs:
 
       - task: BuildQualityChecks@8
         displayName: "Check build quality"
-        enabled: false
         inputs:
           checkCoverage: true
+          coverageThreshold: 58
```
jpeng-ms, line 48: does it mean the build will fail if coverage is below 58? where is this number coming from?

jpeng-ms: I see, so if we don't define `coverageThreshold` its default value would be 60?

emlynmac: Correct.

jpeng-ms (approved):

vhuseinova-msft (approved):

PR #544, opened 2022-12-15
[Bugfix] Fixed Scrolling issue

palatter:
## Purpose

I'm not sure exactly how it was working before and why the merge caused the issue to become apparent. After investigating, I found two issues. 

One, the scroll to the bottom was relying on the object itself as the id, I changed this back to the index for the id, as this seems to be more reliable. 

The second issue is that chatState was being recreated when receiving a message, which caused the sendMessageTimeStamp to use the current time as a default value. So the message list thought a new message had been sent and would scroll to the bottom. Changing this to default to a nil value prevents this from happening.

- [ ] Yes
- [ ] No

## Pull Request Type

- [ ] Bugfix
- [ ] Feature

## Pull request checklist

This PR has considered:
- [ ] Color Theming
- [ ] View Data Injection
- [ ] Demo App UI/UX update

## How to Test

* Open...
* Click on...

## What to Check


Leah-Xia-Microsoft (approved):
LGTM

jpeng-ms (approved):
Tested and works great! Thanks for the quick fix!

PR #549, opened 2022-12-19
[Feature] Disconnect public api stops the real time notifications

JoshuaLai:
## Purpose
- Cleans up the trouter events with the disconnect method. 
- Check the power usage after trouter has been disconnected 
image

- [ ] Yes
- [x] No

## Pull Request Type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

This PR has considered:
- [ ] Color Theming
- [ ] View Data Injection
- [x] Demo App UI/UX update

## How to Test

Start / Join Chat
Close the chat window 
Tap on Stop button to disconnect 
Monitor the trouter logs and energy levels


AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/ChatAdapter.swift
```diff
@@ -82,9 +82,7 @@ public class ChatAdapter {
 
     /// Stop connection to chat composite to Azure Communication Service
     public func disconnect() {
-//    public func disconnect(threadId: String? = nil
-//                           completionHandler: ((Result<Void, ChatCompositeError>) -> Void)? = nil) {
-
+        compositeManager?.stop(completionHandler: {})
```
emlynmac, line 85: If the underlying method is async / callback oriented, we need to propagate that up.

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Service/Chat/ChatService.swift
```diff
@@ -38,6 +39,9 @@ class ChatService: NSObject, ChatServiceProtocol {
         try await chatSDKWrapper.initializeChat()
     }
 
+    func disconnectChatService() {
```
emlynmac, line 42: Why are we renaming this to disconnect, rather than keeping the true functional meaning here?

emlynmac (commented):
I think we need to clean up some naming and potentially review callback / async method usage.

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/ChatAdapter.swift
JoshuaLai: This is a good point. They have both implementations but I don't see why we would just leverage the async / callback one in the sdk wrapper. I will make that change now

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Manager/CompositeManager.swift
```diff
@@ -3,31 +3,46 @@
...
     }
 
     func start() {
         store.dispatch(action: .chatAction(.initializeChatTriggered))
     }
 
-    func stop(completionHandler: @escaping (() -> Void)) {
-        store.dispatch(action: .compositeExitAction)
-        compositeCompletionHandler = completionHandler
+    func stop(completionHandler: @escaping ((Result<Void, ChatCompositeError>) -> Void)) {
```
jpeng-ms, line 44: maybe this can be optional so we don't need to do:

```
self.chatAdapter?.disconnect(completionHandler: { _ in })
```

JoshuaLai: I don't agree here. Ideally we would want developers after they have disconnected to know it is safe to clean up the adapter and that is best practice (something we should address in our documentation). Making it optional now make make developers just ignore it, and we can't make it required later on. I would rather make it required now, and in the future if we feel it needs to be optional, introduce the optionality without it being a breaking change.

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/ChatDemoView.swift
```diff
@@ -161,8 +169,15 @@ struct ChatDemoView: View {
                 .accessibility(identifier: AccessibilityId.startExperienceAccessibilityID.rawValue)
 
                 Button("Stop") {
-                    self.chatAdapter = nil
-                    self.isShowingChatView = false
+                    self.chatAdapter?.disconnect(completionHandler: { result in
+                        switch result {
+                        case .success:
```
emlynmac, line 174: usually, if you switch on `Result<>`, we also unwrap the error case rather than a `default`
Generally, we want to avoid default cases.

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/ChatDemoView.swift
```diff
@@ -161,8 +169,15 @@ struct ChatDemoView: View {
...
+                            self.chatAdapter = nil
+                            self.isShowingChatView = false
```
emlynmac, line 176: Does setting this clear the UI? Might be clearer to have this in a method call?

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/ChatDemoViewController.swift
```diff
@@ -199,7 +208,7 @@ class ChatDemoViewController: UIViewController {
             guard let self = self else {
                 return
             }
-            self.chatAdapter?.disconnect()
+            self.chatAdapter?.disconnect(completionHandler: { _ in })
             self.chatAdapter = nil
```
emlynmac, line 212: Are we ok with nilling the chatAdapter before the completion handler has come back? Usually, you'd want to do any subsequent logic on the callback. This is an example of where an asynchronous context and `await` would make the code linear

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/ChatDemoViewController.swift
```diff
@@ -340,7 +349,15 @@ class ChatDemoViewController: UIViewController {
 
     @objc func onStopBtnPressed() {
         Task { @MainActor in
-            self.chatAdapter = nil
+            self.chatAdapter?.disconnect(completionHandler: { result in
```
emlynmac, line 352: Here, you are in an asynchronous context, which means you could also implement an async version of the disconnect method and then call that with `await` for clearer code.

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Manager/CompositeManager.swift
```diff
@@ -3,31 +3,46 @@
...
+            .sink { [weak self] state in
+                self?.receive(state)
+            }.store(in: &cancellables)
+    }
+
+    private func receive(_ state: AppState) {
+        let chatState = state.chatState
+        if !chatState.isRealTimeNotificationConnected {
+            for handler in compositeCompletionHandler {
+                handler(.success(Void()))
```
emlynmac, line 35: You need to remove the handler from the array here once you've called it

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Manager/CompositeManager.swift
```diff
@@ -3,31 +3,46 @@
...
+    func stop(completionHandler: @escaping ((Result<Void, ChatCompositeError>) -> Void))
 }
 
 class CompositeManager: CompositeManagerProtocol {
     private let logger: Logger
     private let store: Store<AppState>
 
-    private var compositeCompletionHandler: (() -> Void)?
+    var cancellables = Set<AnyCancellable>()
+    private var compositeCompletionHandler: [((Result<Void, ChatCompositeError>) -> Void)] = []
```
emlynmac, line 19: Maybe pluralize the name as it's an array?

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Redux/Middleware/ChatMiddleware.swift
```diff
@@ -55,14 +55,18 @@ private func handleChatAction(_ action: ChatAction,
...
+        break
     case .sendTypingIndicatorTriggered:
        actionHandler.sendTypingIndicator(state: getState(),
                                          dispatch: dispatch)
     case .realTimeNotificationConnected:
         // stub: stop-pulling msg to be implemented for GA
         break
     case .realTimeNotificationDisconnected:
-        // stub: pulling msg to be implemented for GA
+        // stub: stop-pulling msg to be implemented for GA
```
emlynmac, line 69: should pulling be polling here?

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/ChatAdapter.swift
```diff
@@ -81,10 +81,8 @@ public class ChatAdapter {
     }
 
     /// Stop connection to chat composite to Azure Communication Service
-    public func disconnect() {
-//    public func disconnect(threadId: String? = nil
-//                           completionHandler: ((Result<Void, ChatCompositeError>) -> Void)? = nil) {
-
+    public func disconnect(completionHandler: @escaping ((Result<Void, ChatCompositeError>) -> Void)) {
```
jpeng-ms, line 84: shouldn't the ```connect``` function be taken a escaping closure as well? currently it looks like this:
```
public func connect(threadId: String,
                        completionHandler: ((Result<Void, ChatCompositeError>) -> Void)?) 
```

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Manager/CompositeManager.swift
JoshuaLai: Ah yes good catch.

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Manager/CompositeManager.swift
JoshuaLai: I'm actually debating if this should be called `completionQueue` or something.

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/ChatDemoViewController.swift
jpeng-ms: agree, maybe we can also be consistent given ```connect``` has both sync and async implementations

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/ChatDemoViewController.swift
JoshuaLai: No we are not. This is technically incorrect. I should fix this up, just a bit of a hard scenario to test.

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/ChatDemoView.swift
JoshuaLai: I'm not going to lie I didn't put much effort into cleaning up the demo app. But yes we can put this into a new method

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/ChatDemoViewController.swift
```diff
@@ -141,7 +141,16 @@ class ChatDemoViewController: UIViewController {
     @objc func onBackBtnPressed() {
         Task { @MainActor in
             print("Dismissing chat")
-            self.dismiss(animated: true, completion: nil)
+            self.dismiss(animated: true, completion: { [weak self] in
+                self?.chatAdapter?.disconnect(completionHandler: { result in
+                    switch result {
+                    case .success:
+                        self?.chatAdapter = nil
+                    default:
```
emlynmac, line 149: .failure(error) to replace the default?

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/ChatDemoViewController.swift
```diff
@@ -141,7 +141,16 @@ class ChatDemoViewController: UIViewController {
     @objc func onBackBtnPressed() {
         Task { @MainActor in
             print("Dismissing chat")
-            self.dismiss(animated: true, completion: nil)
+            self.dismiss(animated: true, completion: { [weak self] in
+                self?.chatAdapter?.disconnect(completionHandler: { result in
```
emlynmac, line 145: There seem to be a few places where this is called, maybe this should be done in a separate method for consistency?

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Manager/CompositeManager.swift
```diff
@@ -3,31 +3,64 @@
...
+    func stop(completionHandler: @escaping ((Result<Void, ChatCompositeError>) -> Void))
 }
 
 class CompositeManager: CompositeManagerProtocol {
     private let logger: Logger
     private let store: Store<AppState>
 
-    private var compositeCompletionHandler: (() -> Void)?
+    var cancellables = Set<AnyCancellable>()
+    private var compositeCompletionHandlerQueue: [((Result<Void, ChatCompositeError>) -> Void)] = []
```
emlynmac, line 19: I would call it `compositeCompletionHandlers`. This isn't a queue, there's no ordering or processing. 
Basically any completion is going to signal all of the callbacks and remove them all from the array, this is just to handle the case of multiple callees, correct?

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/ChatDemoViewController.swift
JoshuaLai: Agreed. I will tackle the cleaning up of the demo app in a follow up task

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Manager/CompositeManager.swift
JoshuaLai: Done

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Redux/Middleware/ChatMiddleware.swift
JoshuaLai: Updated after conversation with trouter what the actual behaviour is

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/ChatDemoViewController.swift
JoshuaLai: Agreed. I will tackle the cleaning up of the demo app in a follow up task

emlynmac (dismissed):

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/ChatAdapter.swift
```diff
@@ -73,7 +84,10 @@ public class ChatAdapter {
         compositeManager?.start()
     }
 
-    /// Start connection to the chat composite to Azure Communication Service.
+    /// Start connection with chat client and registers for chat events
+    /// This function should be called before adding the Chat Composite to a view
+    /// - Parameters:
+    ///    - threadId: The unique identifier of a chat thread
```
emlynmac, line 90: Looks like the threadId is gone from the method signature?

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/ChatAdapter.swift
```diff
@@ -62,7 +70,10 @@ public class ChatAdapter {
         logger.debug("Composite deallocated")
     }
 
-    /// Start connection to the chat composite to Azure Communication Service.
+    /// Start connection with chat client and registers for chat events
+    /// This function should be called before adding the ChatComposite to a view
+    /// - Parameters:
+    ///    - completionHandler: The closure that would be executed when connection is established
```
emlynmac, line 76: '... that will be called back when...'

emlynmac (approved):

kevinyulu (approved):

palatter (approved):
Looks good. Only comment is to sync with Android on the additional state item.

vhuseinova-msft (approved):

PR #556, opened 2022-12-21
[Bugfix] Resolve the chat view still stays on screen even after chat has been stopped on iPad.

kevinyulu:
## Purpose

* This PR is to resolve the chat view still stays even after chat has been stopped, in pad.
* The side effect of this bug is that user can still send message even after the chat has been disconnected.
The fix was to make chat view visibility state update not happen at the same time with chatAdapter

- [ ] Yes
- [ ] No

## Pull Request Type

- [x] Bugfix
- [ ] Feature

## Pull request checklist

This PR has considered:
- [ ] Color Theming
- [ ] View Data Injection
- [ ] Demo App UI/UX update

## How to Test

* Open...
* Click on...

## What to Check


AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/ChatDemoView.swift
```diff
@@ -162,7 +162,9 @@ struct ChatDemoView: View {
 
                 Button("Stop") {
                     self.chatAdapter = nil
-                    self.isShowingChatView = false
+                    DispatchQueue.main.asyncAfter(deadline: .now() + 0.01) {
```
emlynmac, line 165: Let's use the new Task { @MainActor [weak self] in } syntax.

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/ChatDemoView.swift
```diff
@@ -162,7 +166,10 @@ struct ChatDemoView: View {
 
                 Button("Stop") {
                     self.chatAdapter = nil
-                    self.isShowingChatView = false
+                    Task { @MainActor  in
+                        try await Task.sleep(nanoseconds: Constant.oneSecond / 100)
```
emlynmac, line 170: It would be better to set a constant to one hundred milliseconds, rather than computing it every time.
Why does this value work?
Does it work without sleeping the task?

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/ChatDemoView.swift
```diff
@@ -9,6 +9,10 @@ import SwiftUI
 
 struct ChatDemoView: View {
 
+    private enum Constant {
+        static let oneSecond: UInt64 = 1 * 1_000_000_000
```
emlynmac, line 13: No need to make the compiler work here.

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/ChatDemoView.swift
kevinyulu: I'll update to a constant rather than computing every time. so far the code need to call sleep so that app will not update two states at the same time. without the sleep, ui would have unexpected behaviour if the state is updated while the ui is rendering. from what I see, updating two states will lead to this scenario

palatter (approved):

JoshuaLai (approved):

PR #568, opened 2023-01-03
[Refactor] Cleaning up the disconnect method

JoshuaLai:
## Purpose
Disconnect method could be cleaned up a bit in the demo app. 

- [ ] Yes
- [x] No

## Pull Request Type

- [ ] Bugfix
- [ ] Feature


AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/ChatDemoViewController.swift
```diff
@@ -150,13 +150,11 @@ class ChatDemoViewController: UIViewController {
...
             print("Dismissing chat")
             self.dismiss(animated: true, completion: { [weak self] in
-                self?.chatAdapter?.disconnect(completionHandler: { result in
-                    switch result {
-                    case .success:
-                        self?.chatAdapter = nil
-                    default:
-                        break
+                self?.chatAdapter?.disconnect(completionHandler: { [weak self] result in
+                    guard let self = self else {
```
emlynmac, line 154: this could be more succinctly `let self else {`

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/ChatDemoView.swift
```diff
@@ -255,6 +241,19 @@ extension ChatDemoView {
         }
     }
 
+    private func onDisconnectFromChat(with result: Result<Void, ChatCompositeError>) {
+        switch result {
+        case .success:
+            self.chatAdapter = nil
+            Task { @MainActor  in
+                try await Task.sleep(nanoseconds: Constant.oneMiliSecond)
```
emlynmac, line 249: `millisecond`

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/ChatDemoView.swift
```diff
@@ -172,16 +167,7 @@ struct ChatDemoView: View {
...
-                        case .success:
-                            self.chatAdapter = nil
-                            Task { @MainActor  in
-                                try await Task.sleep(nanoseconds: Constant.oneMiliSecond)
-                                self.isShowingChatView = false
-                            }
-                        case .failure(let error):
-                            print("disconnect error \(error)")
-                        }
+                        self.onDisconnectFromChat(with: result)
```
emlynmac, line 170: Should this be a weak capture of self?

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/ChatDemoView.swift
JoshuaLai: There is no `Task.sleep(millisecond:` and `Task.sleep(from:` is a method that only exist on 16 or later. 
Probably poor naming but 
`static let oneMiliSecond: UInt64 = 10_000_000` is nanosecond converted to milliseconds so I think this is correct

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/ChatDemoViewController.swift
JoshuaLai: Not really sure what you mean by this comment? Do you mean to use a `if let` in this case instead of `guard let`?

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/ChatDemoView.swift
emlynmac: The point is the typo :)

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/ChatDemoViewController.swift
emlynmac: You don't need `guard let self  = self else {`
You can write less code: `guard let self else {`

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/ChatDemoView.swift
emlynmac: It should be consistent with the same usage below.

emlynmac (approved):

jimchou-dev (approved):
LGTM

PR #584, opened 2023-01-05
[Bugfix] Chat duplicate system message

jimchou-dev:
## Purpose

* Fix duplicate system message
  * Root cause: receiving trouter event for system message will not contain messageId (we generate random messageId), and not able to find the system message by matching messageId
  * Workaround: Ignore system message in the fetched previous message list if it happens after initial message fetch timestamp
    * Regular message will always be added/replaced
    * System message after initial fetch will depend on Trouter event messages

- [ ] Yes
- [x] No

## Pull Request Type

- [x] Bugfix
- [ ] Feature

## Pull request checklist

This PR has considered:
- [ ] Color Theming
- [ ] View Data Injection
- [ ] Demo App UI/UX update

## How to Test

* Open...
* Click on...

## What to Check

### DIFF
--- a/AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Manager/MessageRepositoryManager.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Manager/MessageRepositoryManager.swift
@@ -35,6 +35,7 @@ protocol MessageRepositoryManagerProtocol {
 
 class MessageRepositoryManager: MessageRepositoryManagerProtocol {
     var messages: [ChatMessageInfoModel] = []
+    var initialFetchTimestamp = Iso8601Date()
 
     private let eventsHandler: ChatAdapter.Events
 
@@ -53,6 +54,7 @@ class MessageRepositoryManager: MessageRepositoryManagerProtocol {
         if let index = messages.lastIndex(where: {$0.isLocalUser}) {
             messages[index].update(sendStatus: .seen)
         }
+        initialFetchTimestamp = Iso8601Date()
     }
 
     func addPreviousMessages(previousMessages: [ChatMessageInfoModel]) {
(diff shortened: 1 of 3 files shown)


palatter:
I think we can put this in with the bug bash tomorrow morning. Is there a potential full solution at one point? Would a change in the Chat SDK be required?

vhuseinova-msft (approved):

jimchou-dev:
> I think we can put this in with the bug bash tomorrow morning. Is there a potential full solution at one point? Would a change in the Chat SDK be required?

So should this be merged in order to be tested in the bug bash tomorrow? I can revert the change if there's issue found during the bug bash.

Tbh without ChatSDK change, this is likely the 'full solution'.
We did consider some hacky way of setting the message ID in order to compare the fetched list but may not be working as intended if a user is added and removed multiple times (the previous system that should be there would be replaced), and it would be not good for performance.

AzureCommunicationUI/sdk/AzureCommunicationUIChat/Sources/Presentation/Manager/MessageRepositoryManager.swift
```diff
@@ -62,7 +64,11 @@ class MessageRepositoryManager: MessageRepositoryManagerProtocol {
                 $0.id == m.id
             }) {
                 messages[index] = m
-            } else {
+            } else if m.createdOn < initialFetchTimestamp {
```
palatter, line 67: We no longer have an else here, so it could fall through and do nothing. Is this ok?

jimchou-dev: It is intended to add message if matching these conditions
- message id already in list
- message that happen before initial fetch
- message that happens after initial fetch AND are not system message (let regular message through)

Otherwise the logic would look like this with an empty if-else case:
```
if let index = messages.firstIndex(where: {
                $0.id == m.id
            }) {
                messages[index] = m
            } else if m.createdOn > initialFetchTimestamp &&  !m.type.isSystemMessage {
                // do nothing to drop system message
            } else {
                messages.append(m)
            }
```

palatter (approved):

PR #591, opened 2023-01-06
[Refactor] Final UI tests updates

vhuseinova-msft:
## Purpose

* Improved code style
* Updated setup test

- [ ] Yes
- [x] No

## Pull Request Type

- [ ] Bugfix
- [ ] Feature

### DIFF
--- a/AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/FloatingUITestWindow.swift
+++ b/AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/FloatingUITestWindow.swift
@@ -86,100 +86,70 @@ class FloatingUITestWindow: UIWindow {
     @objc func holdButtonTapped(sender: UIButton) {
         debugPrint("UI Test:: holdButtonTapped")
         Task {
-            do {
-                try await callingSDKWrapperMock?.holdCall()
-            } catch {
-            }
+            try? await callingSDKWrapperMock?.holdCall()
         }
     }
 
     @objc func resumeButtonTapped(sender: UIButton) {
         debugPrint("UI Test:: resumeButtonTapped")
         Task {
-            do {
-                try await callingSDKWrapperMock?.resumeCall()
-            } catch {
-            }
+            try? await callingSDKWrapperMock?.resumeCall()
         }
     }
 
     @objc func transcriptionOnButtonTapped(sender: UIButton) {
         debugPrint("UI Test:: transcriptionOnButtonTapped")
         Task {
-            do {
-                try await callingSDKWrapperMock?.transcriptionOn()
-            } catch {
-            }
+            try? await callingSDKWrapperMock?.transcriptionOn()
         }
     }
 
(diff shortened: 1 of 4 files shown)


vhuseinova-msft:
`try catch` changed to `try?`:
1) for now there is no additional logic that should be handled in `FloatingUITestWindow` by catching errors and in case of errors, UI tests will fail as expected
2) it looks like addtional printing is not a usuful tool here as it's not added to Debug output and/or to AppCenter logs.

emlynmac (dismissed):

jimchou-dev (dismissed):
LGTM

jimchou-dev (approved):

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/FloatingUITestWindow.swift
```diff
@@ -86,100 +86,70 @@ class FloatingUITestWindow: UIWindow {
     @objc func holdButtonTapped(sender: UIButton) {
         debugPrint("UI Test:: holdButtonTapped")
         Task {
-            do {
-                try await callingSDKWrapperMock?.holdCall()
-            } catch {
-            }
+            try? await callingSDKWrapperMock?.holdCall()
```
JoshuaLai, line 89: Do we not want to catch and log the error if something goes wrong?

eng/pipelines/jobs/templates/set-appcenter-devices.yml
```diff
@@ -107,7 +107,7 @@ steps:
               APPCENTER_DEVICES=iphonexrset
               ;;
             04)
-              APPCENTER_DEVICES=iphonexrmaxset
+              APPCENTER_DEVICES=iphonexsmaxset
```
JoshuaLai, line 110: Was this change intentional? I remember that this was based on a set schedule so did the schedule change?

JoshuaLai (approved):
Just have a few questions otherwise the changes look fine

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/FloatingUITestWindow.swift
vhuseinova-msft: not now. There aren't any errors for `callingSDKWrapperMock ` functions that used in `FloatingUITestWindow` for now. I don't want to add `fatalError` inside catch, as a default behaviour, because in this case error might not be fatal (as it's for UI tests). Also, it looks like addtional printing is not a usuful tool here as it's not added to Debug output and/or to AppCenter logs.
In case if some functions will be updated with throwing error, it would mean that it was done to be handled in UI tests. So the author can add additional logic for error in  that specific case.

eng/pipelines/jobs/templates/set-appcenter-devices.yml
vhuseinova-msft: we don't have iphonexrmaxset set and because of this 1 of the today's test failed

PR #606, opened 2023-01-26
[Feature][Calling] CSS call history

pavelprystinka:
## Purpose

* CSS call history support

- [ ] Yes
- [ ] No

## Pull Request Type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

This PR has considered:
- [ ] Color Theming
- [ ] View Data Injection
- [ ] Demo App UI/UX update

## How to Test

* Open...
* Click on...

## What to Check

### DIFF
--- a/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallComposite.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallComposite.swift
@@ -161,9 +161,9 @@ public class CallComposite {
             avatarViewManager: avatarViewManager
         )
 
-        self.callHistoryRepository = CallHistoryRepository()
+        self.callHistoryRepository = createCallHistoryRepository()
         let debugInfoManager = createDebugInfoManager()
-        self.callHistoryService = CallHistoryService(store: store)
+        self.callHistoryService = CallHistoryService(store: store, callHistoryRepository: self.callHistoryRepository!)
 
         return CompositeViewFactory(
             logger: logger,
@@ -182,7 +182,11 @@ public class CallComposite {
     }
 
     private func createDebugInfoManager() -> DebugInfoManager {
-        return DebugInfoManager(callHistoryRepository: self.callHistoryRepository ?? CallHistoryRepository())
+        return DebugInfoManager(callHistoryRepository: self.callHistoryRepository ?? createCallHistoryRepository())
+    }
+
+    private func createCallHistoryRepository() -> CallHistoryRepositoryProtocol {
+        return CallHistoryRepository(dbHelper: DBHelper())
     }
 
     private func cleanUpManagers() {
--- a/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallComposite.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallComposite.swift
@@ -40,14 +40,13 @@ public class CallComposite {
     private var avatarViewManager: AvatarViewManagerProtocol?
     private var customCallingSdkWrapper: CallingSDKWrapperProtocol?
     private var debugInfoManager: DebugInfoManagerProtocol?
+    private var callHistoryService: CallHistoryServiceProtocol?
+    private var callHistoryRepository: CallHistoryRepositoryProtocol?
 
     /// Get debug information for the Call Composite.
     public var debugInfo: DebugInfo {
-        guard let debugInfoManager = debugInfoManager else {
-            return DebugInfo()
-        }
-
-        return debugInfoManager.getDebugInfo()
+        let localDebugInfoManager = debugInfoManager ?? createDebugInfoManager()
+        return localDebugInfoManager.getDebugInfo()
     }
 
     /// Create an instance of CallComposite with options.
@@ -162,8 +161,9 @@ public class CallComposite {
             avatarViewManager: avatarViewManager
         )
 
-        let debugInfoManager = DebugInfoManager(store: store)
-        self.debugInfoManager = debugInfoManager
+        self.callHistoryRepository = CallHistoryRepository()
+        let debugInfoManager = createDebugInfoManager()
+        self.callHistoryService = CallHistoryService(store: store)
 
         return CompositeViewFactory(
             logger: logger,
@@ -179,6 +179,10 @@ public class CallComposite {
                 localOptions: localOptions
             )
         )
+    }
+
+    private func createDebugInfoManager() -> DebugInfoManager {
+        return DebugInfoManager(callHistoryRepository: self.callHistoryRepository ?? CallHistoryRepository())
     }
 
     private func cleanUpManagers() {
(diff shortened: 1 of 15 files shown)


AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallComposite.swift
```diff
@@ -162,8 +161,9 @@ public class CallComposite {
             avatarViewManager: avatarViewManager
         )
 
-        let debugInfoManager = DebugInfoManager(store: store)
-        self.debugInfoManager = debugInfoManager
+        self.callHistoryRepository = createCallHistoryRepository()
+        let debugInfoManager = createDebugInfoManager()
+        self.callHistoryService = CallHistoryService(store: store, callHistoryRepository: self.callHistoryRepository!)
```
vhuseinova-msft, line 166: usually it's not a good idea to use a force unwrap (`self.callHistoryRepository!`) in the code as in case if the value = `nil`, the app will crash.

The code can be a couple of possible improvements:
1) `callHistoryRepository` can be changed to not optional lazy type . A couple of examples can be found [here](https://www.avanderlee.com/swift/lazy-var-property/). [Apple documentation](https://docs.swift.org/swift-book/LanguageGuide/Properties.html#ID257). It may be a prefferable option.
2) `callHistoryRepository` can be changed to not optional type and be initialized in `init` funcs
3) create a new local constant for variable returned by `createCallHistoryRepository()` and assidn it to `self.callHistoryRepository`

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Data/CallHistoryRepository.swift
```diff
@@ -0,0 +1,84 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import Foundation
+import SQLite3
```
vhuseinova-msft, line 7: CoreData is framework that helps to work with DB for iOS/macOS. Could you clarify why SQLite3 is used directly in this case?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Data/CallHistoryRepository.swift
```diff
@@ -0,0 +1,84 @@
...
+    init(dbHelper: DBHelper) {
+        self.dbHelper = dbHelper
+    }
+
+    func insert(callDate: Date, callId: String) {
+        if let db = dbHelper.openDatabase() {
+            let fmtr = DateFormatter()
+            fmtr.dateFormat = dateFormat
+
+            let sql = "INSERT INTO CallHistory (Date,CallId) VALUES('"
```
vhuseinova-msft, line 29: it may be a good idea to use CoreData here instead of opening/inserting/closing DB by ourselves

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Data/CallHistoryRepository.swift
```diff
@@ -0,0 +1,84 @@
...
+            sqlite3_exec(db, sql, nil, nil, nil)
+
+            self.dbHelper.dbClose(db: db)
+        }
+    }
+
+    func getAll() -> [CallHistoryRecordData] {
+        var callHistory: [CallHistoryRecordData] = []
+        if let db = dbHelper.openDatabase() {
+            let fmtr = DateFormatter()
```
vhuseinova-msft, line 41: it would be better to have 1 `DateFormatter` instance for `CallHistoryRepository` instead of creating one every time because creating `DateFormatter` is an expenive operation.

After this change, the recommendation is to setup `dateFormat` every time as `dateFormat` can be incorrect for some cases (if there a couple of date formats are used)

vhuseinova-msft: Additional info: https://sarunw.com/posts/how-expensive-is-dateformatter/

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Data/DBHelper.swift
```diff
@@ -0,0 +1,63 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import Foundation
+import SQLite3
+
+class DBHelper {
+    private var dbIsOpen: Bool = false
```
vhuseinova-msft, line 10: It would be better to use CoreData here too

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/Manager/DebugInfoManager.swift
```diff
@@ -11,30 +11,30 @@ protocol DebugInfoManagerProtocol {
...
 
     func getDebugInfo() -> DebugInfo {
-        return DebugInfo(lastCallId: callId)
+        return DebugInfo(callHistoryRecords: getCallHistory())
+    }
+
+    private func getCallHistory() -> [CallHistoryRecord] {
+
+        let callHistoryRecords = callHistoryRepository.getAll()
+        let grouped = Dictionary(grouping: callHistoryRecords, by: { $0.date })
```
vhuseinova-msft, line 28: it looks like grouping isn't used as the records are sorted on line 34

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/Manager/DebugInfoManager.swift
```diff
@@ -11,30 +11,30 @@ protocol DebugInfoManagerProtocol {
...
+
+        let mapped = grouped.map({ (callDate: Date, callRecord: [CallHistoryRecordData]) in
```
vhuseinova-msft, line 30: minor: types may be skipped

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/Manager/DebugInfoManager.swift
vhuseinova-msft, line 28: It's possible to use arrays in DB (with CoreData usage) or just connections 1:Many instead of creating separate instance with the same callStarted dates and process them afterwards

jpeng-ms:
this release branch seems to be very old and it doesn't even have chat: 
https://skype.visualstudio.com/SCC/_build/results?buildId=53477717&view=logs&j=275f1d19-1bd8-5591-b06b-07d489ea915a&t=d6d1571a-a17f-534d-f3cd-109ff09fc86e

Did you cherry picked a particular commit on purpose to get ride of chat?

PR #610, opened 2023-01-27
[Feature][Calling][CSS] Call history

pavelprystinka:
## Purpose

* Call history

- [ ] Yes
- [ ] No

## Pull Request Type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

This PR has considered:
- [ ] Color Theming
- [ ] View Data Injection
- [ ] Demo App UI/UX update

## How to Test

* Open...
* Click on...

## What to Check


grigoryk:
@pavelprystinka can you please check how much of my feeback on https://github.com/Azure/communication-ui-library-android/pull/710 applies here?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallCompositeOptions/DebugInfo.swift
```diff
@@ -7,11 +7,11 @@ import Foundation
 
 /// A Call Composite debug information.
 public struct DebugInfo {
-    /// The current or last known call id for the current CallComposite object. 
-    /// `Nil` is returned if a call hasn't started for CallComposite.
-    public let currentOrLastCallId: String?
+    /// The history of calls up to 30 days.
+    public let callHistoryRecords: [CallHistoryRecord]
```
emlynmac, line 11: This is an API breaking change.
It also requires CallHistoryRecord to be made public; why do we need to have this change?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Data/CallHistoryRepository.swift
```diff
@@ -0,0 +1,48 @@
...
+
+import Foundation
+
+protocol CallHistoryRepositoryProtocol {
+    func insert(callStartedOn: Date, callId: String)
+    func getAll() -> [Date: [String]]
+}
+
+class CallHistoryRepository: CallHistoryRepositoryProtocol {
+    private let storageKey: String = "com.azure.android.communication.ui.calling.CallHistory"
```
emlynmac, line 14: Not sure why we have an android reference here?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Data/CallHistoryRepository.swift
```diff
@@ -0,0 +1,48 @@
...
+            do {
+                let encoder = JSONEncoder()
+                let data = try encoder.encode(historyRecords)
+                UserDefaults.standard.set(data, forKey: storageKey)
+            } catch { }
+        }
+    }
+
+    func getAll() -> [Date: [String]] {
+        if let data = UserDefaults.standard.data(forKey: storageKey) {
```
emlynmac, line 38: Need to make sure the userDefaults object can be passed into the class, for the purpose of controlling testing.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Data/CallHistoryRepository.swift
```diff
@@ -0,0 +1,48 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import Foundation
+
+protocol CallHistoryRepositoryProtocol {
```
emlynmac, line 8: We probably don't need a protocol here

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Data/CallHistoryRepository.swift
```diff
@@ -0,0 +1,48 @@
...
+            if var existingCall = historyRecords[callStartedOn] {
+                existingCall.append(callId)
+            } else {
+                historyRecords[callStartedOn] = [callId]
+            }
+            do {
+                let encoder = JSONEncoder()
+                let data = try encoder.encode(historyRecords)
+                UserDefaults.standard.set(data, forKey: storageKey)
+            } catch { }
```
emlynmac, line 33: Probably want to do something in the catch block

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/Manager/DebugInfoManager.swift
```diff
@@ -11,30 +11,23 @@ protocol DebugInfoManagerProtocol {
...
 
-    init(store: Store<AppState, Action>) {
-        self.store = store
-        self.store.$state
-            .receive(on: DispatchQueue.main)
-            .sink { [weak self] state in
-                self?.receive(state)
-            }.store(in: &cancellables)
+    init(callHistoryRepository: CallHistoryRepositoryProtocol) {
+        self.callHistoryRepository = callHistoryRepository
```
emlynmac, line 17: The previous version of the function stored the local events as they came in.
The overall architecture I believe was to keep the manager looking after the non-redux side.
We probably should keep the manager object responsible for serializing and grabbing objects from the backing store, for simplicity.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Service/Calling/CallHistoryService.swift
```diff
@@ -0,0 +1,41 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import Combine
+import Foundation
+
+protocol CallHistoryServiceProtocol {
```
emlynmac, line 9: No need for an empty protocol

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/CallingDemoView.swift
```diff
@@ -130,6 +143,26 @@ struct CallingDemoView: View {
         .accessibility(identifier: AccessibilityId.startExperienceAccessibilityID.rawValue)
     }
 
+    var showCallHistoryButton: some View {
+        Button("Show call history") {
+            let callComposite = CallComposite()
```
emlynmac, line 148: There are a lot of shared (between UIKit and SwiftUI implementation) objects here. 
Every time this view is re-rendered, each of these will be created.
Let's move the view logic out into a view model, so that it can be shared between both and keep the complexity manageable.

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/CallingDemoViewController.swift
```diff
@@ -354,6 +356,28 @@ class CallingDemoViewController: UIViewController {
         present(settingsViewHostingController, animated: true, completion: nil)
     }
 
+    @objc func onShowHistoryBtnPressed() {
+        let callComposite = CallComposite()
```
emlynmac, line 360: This should be in a shared view model component, used by both UI toolkits.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Data/CallHistoryRepository.swift
```diff
@@ -0,0 +1,48 @@
...
+        }
+    }
+
+    func getAll() -> [Date: [String]] {
+        if let data = UserDefaults.standard.data(forKey: storageKey) {
+            do {
+                let decoder = JSONDecoder()
+                return try decoder.decode([Date: [String]].self, from: data)
```
emlynmac, line 41: Why not use CallHistoryRecord and make it Codable?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Service/Calling/CallHistoryService.swift
```diff
@@ -0,0 +1,41 @@
...
+}
+
+class CallHistoryService: CallHistoryServiceProtocol {
```
emlynmac, line 12: Can this functionality exist in the manager class?

emlynmac (changes requested):
We need to preserve the public interface.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallCompositeOptions/DebugInfo.swift
pavelprystinka: This is a change comparing to previous beta. It is approved by ARB.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Data/CallHistoryRepository.swift
pavelprystinka: Will it be useful in unit tests? If not, will remove.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Data/CallHistoryRepository.swift
pavelprystinka: Typo, fixed.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Data/CallHistoryRepository.swift
pavelprystinka: CallHistoryRecord is part of the public interface. Not sure if ARB will like it.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Data/CallHistoryRepository.swift
pavelprystinka: Good idea, moved to init.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/Manager/DebugInfoManager.swift
pavelprystinka: As this functionality is available for Contoso by only constructing CallComposite and calling method to get diagnosticsInfo, store does not exist yet, so I separated it in two classes. Plus there is extra limitation on Android with context (another reason to have separate service and manager), so decided to follow same architecture here.

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/CallingDemoView.swift
pavelprystinka: That was intention to have a new CallComposite (not initialized with options) instance to return a call history

emlynmac:
It would be preferable to move the persistence of the call IDs into the lower-level Calling SDK, so that all consumers of that library will benefit.
Have we considered this approach?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Data/CallHistoryRepository.swift
emlynmac: Protocols are generally used when there are multiple implementations for a set of functionality.
Unit testing should be possible by passing in dependencies to the object directly.

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/CallingDemoView.swift
```diff
@@ -15,6 +15,9 @@ struct CallingDemoView: View {
     @State var isSettingsDisplayed: Bool = false
     @State var isStartExperienceLoading: Bool = false
     @State var errorMessage: String = ""
+    @State var isShowingCallHistory: Bool = false
+    @State var callHistoryTitle: String = ""
```
vhuseinova-msft, line 19: It would be nice to create a ViewModel for CallingDemoView to move `callHistoryTitle` and `callHistoryMessage` there as properties (as it's not very useful to have them as `State`)

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/CallingDemoView.swift
vhuseinova-msft, line 148: Also, it's not be a good idea to create a new `CallComposite` object for showing the history

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/CallingDemoView.swift
```diff
@@ -130,6 +143,26 @@ struct CallingDemoView: View {
...
 
+    var showCallHistoryButton: some View {
+        Button("Show call history") {
+            let callComposite = CallComposite()
+            let debugInfo = callComposite.debugInfo
+            let callHistory = debugInfo.callHistoryRecords
+            callHistoryTitle = "Total calls: \(callHistory.count)"
+            callHistoryMessage = "Last Call: none"
+            if let lastHistoryRecord = callHistory.last {
+                var formatter = DateFormatter()
```
vhuseinova-msft, line 154: It would be better to have a separate property for `DateFormatter` instead of creating a new object for a button tap. But in this case, the DateFormatter should be a property of ViewModel.

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/CallingDemoView.swift
```diff
@@ -130,6 +143,26 @@ struct CallingDemoView: View {
...
+                formatter.dateFormat = "yyyy-MM-dd HH:mm:ss"
+                callHistoryMessage = "Last Call: \(formatter.string(from: lastHistoryRecord.callStartedOn))"
+                lastHistoryRecord.callIds.forEach { callId in
```
vhuseinova-msft, line 157: minor suggestion: [joined(separator:) ](https://developer.apple.com/documentation/swift/array/joined(separator:)-7uber)func can be used instead

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/CallingDemoView.swift
```diff
@@ -280,9 +313,10 @@ extension CallingDemoView {
     }
 
     private func onError(_ error: CallCompositeError, callComposite: CallComposite) {
+        let callId = callComposite.debugInfo.callHistoryRecords.last?.callIds.last ?? "Unknown"
         print("::::CallingDemoView::getEventsHandler::onError \(error)")
         print("::::CallingDemoView error.code \(error.code)")
-        print("::::CallingDemoView debug info \(callComposite.debugInfo.currentOrLastCallId ?? "Unknown")")
+        print("::::CallingDemoView debug info \(callId)")
```
vhuseinova-msft, line 319: Should we print the last `callHistoryRecords` here instead?

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/CallingDemoViewController.swift
```diff
@@ -494,6 +518,18 @@ class CallingDemoViewController: UIViewController {
                                                              right: LayoutConstants.buttonHorizontalInset)
         settingsButton.accessibilityIdentifier = AccessibilityId.settingsButtonAccessibilityID.rawValue
 
+        showCallHistoryButton = UIButton()
+        showCallHistoryButton.setTitle("Show call history", for: .normal)
+        showCallHistoryButton.backgroundColor = .systemBlue
+//        showCallHistoryButton.setTitleColor(UIColor.white, for: .normal)
```
vhuseinova-msft, line 524: Is it still needed?

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/CallingDemoViewController.swift
```diff
@@ -494,6 +518,18 @@ class CallingDemoViewController: UIViewController {
...
+//        showCallHistoryButton.setTitleColor(UIColor.systemGray6, for: .disabled)
+        showCallHistoryButton.contentEdgeInsets = UIEdgeInsets.init(top: LayoutConstants.buttonVerticalInset,
+                                                                    left: LayoutConstants.buttonHorizontalInset,
+                                                                    bottom: LayoutConstants.buttonVerticalInset,
+                                                                    right: LayoutConstants.buttonHorizontalInset)
+        showCallHistoryButton.layer.cornerRadius = 8
```
vhuseinova-msft, line 530: It looks like a lot of code is reused for buttons creation, it may be a good idea to create a separate func with reused code

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Data/CallHistoryRepository.swift
```diff
@@ -0,0 +1,53 @@
...
+    func insert(callStartedOn: Date, callId: String) {
+        var historyRecords = getAll()
+
+        if let thresholdDate = Calendar.current.date(byAdding: DateComponents(day: -31), to: Date()) {
+            historyRecords = historyRecords.filter { callHistoryRecord in
+                callHistoryRecord.key >= thresholdDate
+            }
+
+            if var existingCall = historyRecords[callStartedOn] {
+                existingCall.append(callId)
```
vhuseinova-msft, line 30: Should we use an encrypted value for `callId`?
https://developer.apple.com/documentation/applearchive/encrypting_and_decrypting_a_string

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Data/CallHistoryRepository.swift
vhuseinova-msft: We probably can use internal extension for `CallHistoryRecord` implementing `Codable` in this case

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/CallingDemoViewModel.swift
```diff
@@ -0,0 +1,25 @@
...
+//
+
+import Foundation
+#if DEBUG
+@testable import AzureCommunicationUICalling
+#else
+import AzureCommunicationUICalling
+#endif
+
+class CallingDemoViewModel: ObservableObject {
```
vhuseinova-msft, line 13: `CallComposite` should be property of `CallingDemoViewModel` instead of View or ViewController

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/EntryViewController.swift
```diff
@@ -16,6 +16,7 @@ class EntryViewController: UIViewController {
     private var callingSDKWrapperMock: UITestCallingSDKWrapper?
 #endif
     private var cancellables = Set<AnyCancellable>()
+    private var callingViewModel = CallingDemoViewModel()
```
vhuseinova-msft, line 19: it looks like `callingViewModel` isn't changed anywhere so it can be a constant instead of variable

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/CallingDemoViewModel.swift
pavelprystinka: That will not be enough. There is already lots of duplicate code between View and ViewController. Initialization with/without mock SDK, events subscription, launch.
Will be done separately from this feature.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Tests/Service/CallHistoryRepositoryTests.swift
```diff
@@ -0,0 +1,85 @@
...
+        sut.insert(callStartedOn: callStartDate1, callId: callId2)
+
+        records = sut.getAll()
+        var firstCall = records.first(where: { $0.callStartedOn == callStartDate1 })!
+
+        XCTAssertEqual(records.count, 1)
+        XCTAssertEqual(firstCall.callStartedOn, callStartDate1)
+        XCTAssertEqual(firstCall.callIds.count, 2)
+        XCTAssertTrue(firstCall.callIds.contains(callId1))
+        XCTAssertTrue(firstCall.callIds.contains(callId2))
```
vhuseinova-msft, line 42: Minor: Lines 40-42 can be changed to
`XCTAssertEqual(firstCall.callIds, [callId1, callId2])`

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Data/CallHistoryRepository.swift
```diff
@@ -0,0 +1,63 @@
...
+}
+
+class CallHistoryRepository: CallHistoryRepositoryProtocol {
+    private let storageKey: String = "com.azure.ios.communication.ui.calling.CallHistory"
+    private let logger: Logger
+    private let userDefaults: UserDefaultsStorageProtocol
+
+    init(logger: Logger, userDefaults: UserDefaultsStorageProtocol) {
+        self.logger = logger
+        self.userDefaults = userDefaults
```
vhuseinova-msft, line 20: Should we add data clean up here (for the data that are older than 31 day)?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Tests/Service/CallHistoryServiceTests.swift
```diff
@@ -0,0 +1,57 @@
...
+    private var storeFactory: StoreFactoryMocking!
+    override func setUp() {
+        super.setUp()
+        callHistoryRepository = CallHistoryRepositoryMocking()
+        storeFactory = StoreFactoryMocking()
+    }
+
+    override func tearDown() {
+        super.tearDown()
+        callHistoryRepository = nil
```
vhuseinova-msft, line 21: `storeFactory` isn't set to nil. It shouldn't be a critical issue but it would be nice to be consistent with `setUp`

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Tests/Service/CallHistoryServiceTests.swift
```diff
@@ -0,0 +1,57 @@
...
+
+    func test_callingHistoryService_whenCallIDChanges_shouldCallinsertOnRepository() {
+        let sut = makeSUT()
+        let callId = "call id"
+        let callStartDate = Date()
+        XCTAssertFalse(callHistoryRepository.insertWasCalled())
+
+        sut.receive(AppState(callingState: CallingState(callStartDate: callStartDate)))
+        // CallHistoryService is saving call history record async, so we need to allow time to call repository
+        Thread.sleep(forTimeInterval: 5)
```
vhuseinova-msft, line 32: Could you please use expectations and closures instead of `sleep` func?
https://developer.apple.com/documentation/xctest/asynchronous_tests_and_expectations

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Data/CallHistoryRepository.swift
pavelprystinka: I'm not sure it good to do such work while initialization phase. We do clean on insert, which is called on every call id change. I've updated logic so when getAll is called old records are also filtered out.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Data/CallHistoryRepository.swift
pavelprystinka: Privacy team did not consider this as any private data.

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/CallingDemoView.swift
pavelprystinka: It is intentionally to tests that new instance of CallComposite (not launched) can retrieve call history.

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/CallingDemoView.swift
```diff
@@ -15,7 +15,11 @@ struct CallingDemoView: View {
     @State var isSettingsDisplayed: Bool = false
     @State var isStartExperienceLoading: Bool = false
     @State var errorMessage: String = ""
+    @State var isShowingCallHistory: Bool = false
+    @State var callHistoryTitle: String = ""
+    @State var callHistoryMessage: String = ""
```
emlynmac, line 20: Why is this @State?

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/CallingDemoViewController.swift
```diff
@@ -354,6 +360,24 @@ class CallingDemoViewController: UIViewController {
...
     }
 
+    @objc func onShowHistoryBtnPressed() {
+        let callHistory = callingViewModel.callHistory
+        var callHistoryTitle = "Total calls: \(callHistory.count)"
+        var callHistoryMessage = "Last Call: none"
+        if let lastHistoryRecord = callHistory.last {
+            let formattedDate = callingViewModel.dateFormatter.string(from: lastHistoryRecord.callStartedOn)
+            callHistoryMessage = "Last Call: \(formattedDate)\n"
+            callHistoryMessage += "Call Ids:\n"
```
emlynmac, line 370: This is duplicated in the SwiftUI side; move to the view model to implement logic once?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Data/UserDefaultsStorage.swift
```diff
@@ -0,0 +1,13 @@
...
+//
+
+import Foundation
+
+protocol UserDefaultsStorageProtocol {
+    func data(forKey defaultName: String) -> Data?
+    func set(_ value: Any?, forKey key: String)
+}
+
+extension UserDefaults: UserDefaultsStorageProtocol {}
```
emlynmac, line 13: Empty implementation definition.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Data/UserDefaultsStorage.swift
```diff
@@ -0,0 +1,13 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import Foundation
+
+protocol UserDefaultsStorageProtocol {
+    func data(forKey defaultName: String) -> Data?
+    func set(_ value: Any?, forKey key: String)
```
emlynmac, line 10: Any is too broad; better to provide specific set methods for supported types.

pavelprystinka: This is how UserDefaults are implemented:
    open func set(_ value: Any?, forKey defaultName: String)

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Data/UserDefaultsStorage.swift
pavelprystinka: It's to make UserDefaults also implementing a protocol

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Data/CallHistoryRepository.swift
```diff
@@ -0,0 +1,66 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import Foundation
+
+class CallHistoryRepository {
+    private let callHistoryDispatchQueue = DispatchQueue(label: "CallHistoryDispatchQueue")
```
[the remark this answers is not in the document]
pavelprystinka: This is the way I found on one of the Apple's articles.

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/CallingDemoViewController.swift
```diff
@@ -527,6 +561,22 @@ class CallingDemoViewController: UIViewController {
...
+
+        let showHistoryButtonHStack = UIStackView(arrangedSubviews: [showHistoryButtonHSpacer1,
+                                                                  showCallHistoryButton,
+                                                                  showHistoryButtonHSpacer2])
+        showHistoryButtonHStack.axis = .horizontal
+        showHistoryButtonHStack.alignment = .fill
+        showHistoryButtonHStack.distribution = .fill
+        showHistoryButtonHStack.translatesAutoresizingMaskIntoConstraints = false
+
         let startButtonHSpacer1 = UIView()
```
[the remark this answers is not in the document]
pavelprystinka: Implemented it on consistency with other buttons

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/CallingDemoViewController.swift
pavelprystinka: This is a UI logic, not a viewModel's

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/CallingDemoView.swift
pavelprystinka: It is required for alert on the View

eng/pipelines/nightly.yml
```diff
@@ -41,7 +41,7 @@ jobs:
     displayName: "Run XCUITests on AppCenter with specific devices"
     condition: ne(dependencies.ConfigureDevices.outputs['DetermineDeviceSet.AppCenterDevices'], '')
     pool:
-      name: vsts-mac-1301-xcode-141
```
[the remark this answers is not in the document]
pavelprystinka: If I need to make this branch run on that pipeline I need this change.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Data/UserDefaultsStorage.swift
emlynmac: You're defining the protocol here, so let's get the protocol better designed with type safety.

eng/pipelines/nightly.yml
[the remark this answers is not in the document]
emlynmac: It's going to create a conflict when it comes time to merge.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Data/UserDefaultsStorage.swift
emlynmac: Which is a really bad idea; you're defining a protocol that happens to correspond to the public API of user defaults and then creating a empty inferface which happens to adhere to these exact methods.

Why do you need the protocol at all?

You just need to have the ability to operate on a specific UserDefaults object and have the ability to inject that instance as needed.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Data/CallHistoryRepository.swift
```diff
@@ -0,0 +1,66 @@
...
+    private let storageKey: String = "com.azure.ios.communication.ui.calling.CallHistory"
+    private let logger: Logger
+    private let userDefaults: UserDefaultsStorageProtocol
+
+    init(logger: Logger, userDefaults: UserDefaultsStorageProtocol) {
+        self.logger = logger
+        self.userDefaults = userDefaults
+    }
+
+    func insert(callStartedOn: Date, callId: String) {
```
emlynmac, line 19: There is no way to tell from the calling function whether this succeeded or failed.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Data/CallHistoryRepository.swift
```diff
@@ -0,0 +1,66 @@
...
+        callHistoryDispatchQueue.async {
```
emlynmac, line 20: Why do you want to process this on another queue?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Data/CallHistoryRepository.swift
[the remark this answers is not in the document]
emlynmac: I would like to see the article, as I'm not sure why there's a queue here at all.
It seems it is used to run the insert, which may fail - and there's no way to ascertain if it failed.

Why do you need the queue - what problem are you solving with it?

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/CallingDemoView.swift
```diff
@@ -130,6 +144,22 @@ struct CallingDemoView: View {
         .accessibility(identifier: AccessibilityId.startExperienceAccessibilityID.rawValue)
     }
 
+    var showCallHistoryButton: some View {
+        Button("Show call history") {
+            let callHistory = callingViewModel.callHistory
```
vhuseinova-msft, line 149: The calculation logic should be moved to the VM

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/CallingDemoViewController.swift
```diff
@@ -23,6 +23,7 @@ class CallingDemoViewController: UIViewController {
         static let buttonHorizontalInset: CGFloat = 20.0
         static let buttonVerticalInset: CGFloat = 10.0
     }
+    @ObservedObject var callingViewModel: CallingDemoViewModel
```
vhuseinova-msft, line 26: Do we need to use `callingViewModel` as `@ObservedObject` here?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Service/Calling/CallHistoryService.swift
```diff
@@ -0,0 +1,41 @@
...
+    func receive(_ state: AppState) {
+        guard let updatedCallId = state.callingState.callId,
+                !updatedCallId.isEmpty,
+              let callStartDate = state.callingState.callStartDate,
+              self.updatedCallId != updatedCallId else {
+            return
+        }
+        self.updatedCallId = updatedCallId
+
+        callHistoryDispatchQueue.async {
```
vhuseinova-msft: I'm wondering if we should use `sync` function here instead of `async`

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Tests/Service/CallHistoryRepositoryTests.swift
```diff
@@ -0,0 +1,79 @@
...
+        XCTAssertEqual(records.count, 0)
+
+        sut.insert(callStartedOn: callStartDate1, callId: callId1)
+        // CallHistoryService is saving call history record async, so we need to allow time to call repository
+        wait(for: [setExpectation], timeout: 1)
+    }
+
+    func test_callingHistoryRepository_whenRecordsInserted_shouldReturnRecords2() {
+        let setExpectation = XCTestExpectation(description: "Defaul user storage set call is expected")
+        setExpectation.expectedFulfillmentCount = 4
```
vhuseinova-msft, line 36: `assertForOverFulfill` should be set to `true`

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Data/CallHistoryRepository.swift
vhuseinova-msft: also, if it's done on a separate queue, it may need to be done in sync way

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/CallingDemoViewController.swift
emlynmac: It's formatting a string for display - the logic here is something that could and should be shared; the view model would be the logical place to share the formatting logic, for the purpose of consistency and DRY design.

AzureCommunicationUI/AzureCommunicationUIDemoApp/Sources/Views/CallingDemoView.swift
emlynmac: It should be a property on the view model, as it is used in both UIKit and SwiftUI implementations and should be consistent.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallCompositeOptions/CallHistoryRecord.swift
```diff
@@ -0,0 +1,21 @@
...
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import Foundation
+
+/// Call history.
+public struct CallHistoryRecord {
+
+    /// Local date call started on.
```
emlynmac, line 11: Date is not tied to a particular timezone.  Local should be removed here.
ref: https://developer.apple.com/documentation/foundation/date

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Data/CallHistoryRepository.swift
```diff
@@ -0,0 +1,67 @@
...
+                continuation.resume(returning: nil)
+            } catch let error {
+                self.logger.error("Failed to save call history, reason: \(error.localizedDescription)")
+                continuation.resume(returning: error)
+            }
+        }
+    }
+
+    func getAll() -> [CallHistoryRecord] {
+        return getAllAsDictionary()
```
emlynmac, line 40: This could be more succinctly writen:
```
    func getAll() -> [CallHistoryRecord] {
        return getAllAsDictionary()
            .map(CallHistoryRecord.init(callStartedOn:callIds:))
    }
   ```

File: AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Data/CallHistoryRepository.swift [hunk 40]
```diff
@@ -0,0 +1,67 @@
...
+            .map({ (callStartedOn, callIds) in
+                return CallHistoryRecord(callStartedOn: callStartedOn, callIds: callIds)
+            })
+    }
+
+    private func cleanupOldRecords(_ historyRecords: [Date: [String]]) -> [Date: [String]] {
```
emlynmac, line 46: I would expect a `cleanup` function to remove the keys from the underlying storage.
This function is just a filter, which means that over time, the backing store is going to grow and grow.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallCompositeOptions/CallHistoryRecord.swift
pavelprystinka: It is according to guidelines: https://azure.github.io/azure-sdk/ios_design.html#ios-request-options-types

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Redux/State/CallingState.swift
```diff
@@ -30,17 +30,20 @@ struct CallingState: Equatable {
...
     let isTranscriptionActive: Bool
+    let callStartDate: Date?
 
     init(status: CallingStatus = .none,
          operationStatus: OperationStatus = .none,
          callId: String? = nil,
          isRecordingActive: Bool = false,
-         isTranscriptionActive: Bool = false) {
+         isTranscriptionActive: Bool = false,
+         callStartDate: Date? = nil) {
```
emlynmac, line 40: Maybe put the callID log error here?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Service/Calling/CallHistoryService.swift
```diff
@@ -0,0 +1,43 @@
...
+                !updatedCallId.isEmpty,
+              let callStartDate = state.callingState.callStartDate,
+              self.updatedCallId != updatedCallId else {
+            return
+        }
+        self.updatedCallId = updatedCallId
+        recordCallHistory(callStartedOn: callStartDate, callId: updatedCallId)
+    }
+
+    func recordCallHistory(callStartedOn: Date, callId: String) {
```
emlynmac, line 38: Can this be private?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Tests/Service/CallHistoryServiceTests.swift
```diff
@@ -0,0 +1,53 @@
...
+
+    func test_callingHistoryService_whenCallIDChanges_shouldCallinsertOnRepository() {
+        let sut = makeSUT()
+        let callId = "call id"
+        let callStartDate = Date()
+        XCTAssertFalse(callHistoryRepository.insertWasCalled())
+
+        sut.receive(AppState(callingState: CallingState(callStartDate: callStartDate)))
+
+        XCTAssertFalse(sut.recordCallHistoryWasCalled())
```
vhuseinova-msft, line 33: This unit test can be split into 2:
1) to check that the record call history wasn't called. THis can be checked with[ inverted expectations](https://developer.apple.com/documentation/xctest/xctestexpectation/2806573-inverted?language=objc)
2) to check that record call history was called with usual expectation + check that date and call id is the same as was sent in the test

the comment can be addressed in a separate PR

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Tests/CallHistoryRepositoryMocking.swift
```diff
@@ -0,0 +1,29 @@
...
+
+import Foundation
+@testable import AzureCommunicationUICalling
+
+class CallHistoryRepositoryMocking: CallHistoryRepository {
+    var insertCallCount: Int = 0
+
+    init() {
+        super.init(logger: LoggerMocking(),
+                   userDefaults: UserDefaults())
```
emlynmac, line 14: This should not use UserDefaults() directly - create a non-persistent one from the test harness, as the CallHistoryRepositoryTests do

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Service/Calling/CallHistoryService.swift
pavelprystinka: Need to access from test mock. I believe there is not protected access level, nor alternative?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Data/CallHistoryRepository.swift
pavelprystinka: a filtered out records are used in insert() method. This way we get rid of old records.

emlynmac: Ok, so the functionality is not very clear; the insert gets all, which actually gets a filtered list and then saves the updated list back.
Let's rename the cleanup function to `filterExpiredRecords`

While we're there, let's break out the 31 day hard-coded value to a variable.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Tests/Service/CallHistoryServiceTests.swift
pavelprystinka: Noted in user story

vhuseinova-msft (approved):

emlynmac (approved):

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Redux/State/CallingState.swift
pavelprystinka: Task is created to address this.

PR #680, opened 2023-05-30
[Calling] Enable support for CallKit.

raosanat:
## Purpose

* Enable/Disable support for CallKit in UI SDK. CallKit support in AzureCommunicationCalling SDK will be used. 

- [ ] Yes
- [x] No

## Pull Request Type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

This PR has considered:
- [ ] Color Theming
- [ ] View Data Injection
- [x] Demo App UI/UX update

## How to Test

* Make an outgoing call or join call and CallKit window should appear in the background.

### DIFF
--- a/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallCompositeOptions/RemoteOptions.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallCompositeOptions/RemoteOptions.swift
@@ -27,6 +27,9 @@ public struct RemoteOptions {
     /// The limit for string length is 256.
     public let displayName: String?
 
+    /// CallKit in ACS SDK to be enabled or not
+    public let enableCallKitInSDK: Bool
+
     /// Create an instance of a RemoteOptions with options.
     /// - Parameters:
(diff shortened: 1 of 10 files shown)


AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallCompositeOptions/RemoteOptions.swift
```diff
@@ -27,16 +27,21 @@ public struct RemoteOptions {
     /// The limit for string length is 256.
     public let displayName: String?
 
+    /// CallKit in ACS SDK to be enabled or not
+    public let enableCallKitInSDK: Bool
```
pavelprystinka, line 31: RemoteOptions is holding a set on configuration fields for a one call (a single call).
We have CallCompositeOptions for CallComposite wide options (shared between multiple .launch() calls). I think enableCallKitInSDK would be more suitable there.
Also I think this property name could be better: enableCallKit

pavelprystinka (changes requested):
Let's have a branch feature/callkit_support from develop and point this PR to there instead on main.

PR #685, opened 2023-06-15
[Calling][Feature] Mic Unavailable scenario

Mohtasim:
## Purpose

* Mic used by other app implementation

- [ ] Yes
- [ ] No

## Pull Request Type

- [ ] Bugfix
- [ ] Feature

## Pull request checklist

This PR has considered:
- [ ] Color Theming
- [ ] View Data Injection
- [ ] Demo App UI/UX update

## How to Test

* Open...
* Click on...

## What to Check

### DIFF
--- a/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallComposite.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallComposite.swift
@@ -35,7 +35,7 @@ public class CallComposite {
     private var errorManager: ErrorManagerProtocol?
     private var lifeCycleManager: LifeCycleManagerProtocol?
     private var permissionManager: PermissionsManagerProtocol?
-    private var audioSessionManager: AudioSessionManagerProtocol?
+    private var audioSessionManager: AudioSessionManager?
     private var remoteParticipantsManager: RemoteParticipantsManagerProtocol?
     private var avatarViewManager: AvatarViewManagerProtocol?
     private var customCallingSdkWrapper: CallingSDKWrapperProtocol?
@@ -176,7 +176,7 @@ public class CallComposite {
                 logger: logger,
                 store: store,
                 networkManager: NetworkManager(),
-                audioSessionManager: AudioSessionManager(store: store, logger: logger),
+                audioSessionManager: audioSessionManager ?? AudioSessionManager(store: store, logger: logger),
                 localizationProvider: localizationProvider,
                 accessibilityProvider: accessibilityProvider,
                 debugInfoManager: debugInfoManager,
--- a/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallComposite.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallComposite.swift
@@ -35,7 +35,7 @@ public class CallComposite {
     private var errorManager: ErrorManagerProtocol?
     private var lifeCycleManager: LifeCycleManagerProtocol?
     private var permissionManager: PermissionsManagerProtocol?
-    private var audioSessionManager: AudioSessionManager?
+    private var audioSessionManager: AudioSessionManagerProtocol?
     private var remoteParticipantsManager: RemoteParticipantsManagerProtocol?
     private var avatarViewManager: AvatarViewManagerProtocol?
     private var customCallingSdkWrapper: CallingSDKWrapperProtocol?
@@ -167,7 +167,8 @@ public class CallComposite {
         let debugInfoManager = createDebugInfoManager()
         self.debugInfoManager = debugInfoManager
         self.callHistoryService = CallHistoryService(store: store, callHistoryRepository: self.callHistoryRepository)
-
+        let audioSessionManager = AudioSessionManager(store: store, logger: logger)
+        self.audioSessionManager = audioSessionManager
         return CompositeViewFactory(
             logger: logger,
             avatarManager: avatarViewManager,
@@ -176,7 +177,7 @@ public class CallComposite {
                 logger: logger,
                 store: store,
                 networkManager: NetworkManager(),
-                audioSessionManager: audioSessionManager ?? AudioSessionManager(store: store, logger: logger),
+                audioSessionManager: audioSessionManager,
                 localizationProvider: localizationProvider,
                 accessibilityProvider: accessibilityProvider,
                 debugInfoManager: debugInfoManager,
--- a/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallComposite.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallComposite.swift
@@ -176,6 +176,7 @@ public class CallComposite {
                 logger: logger,
                 store: store,
                 networkManager: NetworkManager(),
+                audioSessionManager: AudioSessionManager(store: store, logger: logger),
                 localizationProvider: localizationProvider,
                 accessibilityProvider: accessibilityProvider,
                 debugInfoManager: debugInfoManager,
--- a/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/Factories/CompositeViewModelFactory.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/Factories/CompositeViewModelFactory.swift
@@ -206,6 +206,7 @@ class CompositeViewModelFactory: CompositeViewModelFactoryProtocol {
         LoadingOverlayViewModel(localizationProvider: localizationProvider,
                                 accessibilityProvider: accessibilityProvider,
                                 networkManager: networkManager,
+                                audioSessionManager: audioSessionManager,
                                 store: store)
     }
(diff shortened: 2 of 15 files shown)


AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/Manager/AudioSessionManager.swift
```diff
@@ -100,7 +101,7 @@ class AudioSessionManager: AudioSessionManagerProtocol {
         store.dispatch(action: .localUserAction(.audioDeviceChangeSucceeded(device: currentDevice)))
     }
 
-    private func activateAudioSessionCategory() {
```
ShaunaSong, line 103: Why we use public here?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/Manager/AudioSessionManager.swift
```diff
@@ -23,7 +22,7 @@ class AudioSessionManager: AudioSessionManagerProtocol {
          logger: Logger) {
         self.store = store
         self.logger = logger
-        let currentAudioDevice = getCurrentAudioDevice()
+        let currentAudioDevice = self.getCurrentAudioDevice()
```
ShaunaSong, line 25: why we add self. here ?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/Manager/AudioSessionManager.swift
Mohtasim: Previously the function was only used by this class. But we need to know in outer modules whether the AudioSessionCatagory was activated successfully or not. So I had to make it public for that.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/Manager/AudioSessionManager.swift
Mohtasim: Following the pattern for `setupAudioSession()` .. I see no harm removing it. But same method, file needs to be consistent.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/Manager/AudioSessionManager.swift
ShaunaSong: Could we double check this public method will be exposed to Contoso as public api? 

Also could we check the method as `func activateAudioSessionCategory() -> Bool`,  does it work as expected?

ahammer (dismissed):

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/Manager/AudioSessionManager.swift
ahammer: Since we don't expose AudioSessionManager on the API, I don't think it's fields are at risk of being exposed.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallComposite.swift
```diff
@@ -176,6 +176,7 @@ public class CallComposite {
                 logger: logger,
                 store: store,
                 networkManager: NetworkManager(),
+                audioSessionManager: AudioSessionManager(store: store, logger: logger),
```
pavelprystinka, line 179: There is constructViewFactoryAndDependencies() and cleanUpManagers() to create and clean up manages.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/Manager/AudioSessionManager.swift
pavelprystinka: Can it be made internal (without a modifier)?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/Setup/SetupViewModel.swift
```diff
@@ -167,4 +174,13 @@ class SetupViewModel: ObservableObject {
...
         errorInfoViewModel.show()
     }
+
+    private func handleMicUnavailableEvent() {
+        store.dispatch(action: .errorAction(.statusErrorAndCallReset(internalError: .micNotAvailable,
+                                                                     error: nil)))
+        // only show banner again when user taps on button explicitly
+        // banner would not reappear when other events^1 send identical error state again
+        // 1: camera on/off, audio on/off, switch to background/foreground, etc.
+        errorInfoViewModel.show()
```
pavelprystinka, line 184: Do we need to call errorInfoViewModel.show() here? 
Isn't it covered by error handling mechanism in response to store.dispatch(action: .errorAction...)

pavelprystinka (changes requested):

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/Setup/SetupViewModel.swift
Mohtasim: It needs to be called out for show.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallComposite.swift
Mohtasim: Thank you. Implementing rest of the flow according to protocol.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallComposite.swift
```diff
@@ -35,7 +35,7 @@ public class CallComposite {
     private var errorManager: ErrorManagerProtocol?
     private var lifeCycleManager: LifeCycleManagerProtocol?
     private var permissionManager: PermissionsManagerProtocol?
-    private var audioSessionManager: AudioSessionManagerProtocol?
+    private var audioSessionManager: AudioSessionManager?
```
pavelprystinka, line 38: It is better to keep abstraction of AudioSessionManagerProtocol here.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallComposite.swift
```diff
@@ -176,6 +176,7 @@ public class CallComposite {
                 logger: logger,
                 store: store,
                 networkManager: NetworkManager(),
+                audioSessionManager: audioSessionManager ?? AudioSessionManager(store: store, logger: logger),
```
pavelprystinka, line 179: Is there ever a case when audioSessionManager is nil here?
A few lines above audioSessionManager is already created
I guess here you are trying to avoid passing nil to the CompositeViewModelFactory

You could change lines above to:
`let audioSessionManager = AudioSessionManager(store: store, logger: logger)
  self.audioSessionManager = audioSessionManager`

and then use local non-nil constant audioSessionManager to pass to CompositeViewModelFactory

pavelprystinka (changes requested):

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/Factories/CompositeViewModelFactory.swift
```diff
@@ -82,13 +83,15 @@ class CompositeViewModelFactory: CompositeViewModelFactoryProtocol {
     init(logger: Logger,
          store: Store<AppState, Action>,
          networkManager: NetworkManager,
+         audioSessionManager: AudioSessionManager,
```
pavelprystinka, line 86: Should take AudioSessionManagerProtocol

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/Calling/CallingViewComponent/Overlay/LoadingOverlayViewModel.swift
```diff
@@ -15,16 +15,19 @@ class LoadingOverlayViewModel: OverlayViewModelProtocol {
     private var audioPermission: AppPermission.Status = .unknown
     var cancellables = Set<AnyCancellable>()
     var networkManager: NetworkManager
+    var audioSessionManager: AudioSessionManager
 
     init(localizationProvider: LocalizationProviderProtocol,
          accessibilityProvider: AccessibilityProviderProtocol,
          networkManager: NetworkManager,
+         audioSessionManager: AudioSessionManager,
```
pavelprystinka, line 23: Should take in AudioSessionManagerProtocol.
Also you will need to extend protocol with function you are using here.

pavelprystinka (changes requested):

pavelprystinka (approved):

ahammer (approved):

PR #745, opened 2023-10-11
[Calling][Feature][Release] Implementation of UFD diagnostics notification banners

lucianopa-msft:
## Purpose

* This implements handle of User Facing Diagnostics feature in the UISDK as notification banners in call screen following the specified design: https://www.figma.com/file/BOeihN2NJE7ROdyqbihCxi/Mobile-UI-Library?type=design&node-id=7588-30394&mode=design&t=Jp08jcHG0gYzeXF3-0
* 

- [ ] Yes
- [x] No

## Pull Request Type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

This PR has considered:
- [x] Color Theming
- [x] iOS supported versions (all major versions greater than or equal current target deployment version)
- [x] VoiceOver and Keyboard Accessibility
- [x] Internationalization and Right to Left layouts
- [x] Size classes and window sizes (iPhone vs iPad, notched devices, multitasking, different window sizes, etc)
- [x] View Data Injection
- [x] Demo App UI/UX update

## How to Test

There a couple of diagnostics a some are hard to test because they indicate a underlying device issue, but a few ones to test may be:
Speaking While Muted
* Join a call
* Mute yourself 
* Start speaking 
* After some time we should see the banner

Network Quality
* Before join a call switch on [network link conditioner iOS feature](https://nshipster.com/network-link-conditioner/) to a bad network.
* Join a call
* Is likely to see a bad network banner 
* Also turn on/off internet should give reconnecting state.

## What to Check


AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallCompositeOptions/LocalOptions.swift
```diff
@@ -17,12 +17,13 @@ public struct LocalOptions {
...
     let roleHint: ParticipantRole?
-    ///  The CameraOn is used when we skip the setup screen
+    /// The CameraOn is used when we skip the setup screen
     let cameraOn: Bool?
-    ///  The MicrophoneOnis used when we skip the setup screen
+    /// The MicrophoneOn is used when we skip the setup screen
     let microphoneOn: Bool?
-    ///  The SkipSetupScreen is used when we skip the setup screen
+    /// The SkipSetupScreen is used when we skip the setup screen
     let skipSetupScreen: Bool?
```
iaulakh, line 25: are we keeping this in local options? I think we have discussion to move this to place where we handle theme and localization

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Redux/Action/DiagnosticsAction.swift
```diff
@@ -0,0 +1,10 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+enum DiagnosticsAction: Equatable {
+    case networkQuality(diagnostic: NetworkQualityDiagnosticModel)
+    case network(diagnostic: NetworkDiagnosticModel)
+    case media(diagnostic: MediaDiagnosticModel)
```
iaulakh, line 9: do we want to add action to clear state when contoso press close button on banner? this will help avoid showing banner if Contoso close on scenarios like rotating screen, going to background and coming back

iaulakh (commented):
would it be possible to add UI tests through mocking? please look at PR for lobby admit

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallCompositeOptions/LocalOptions.swift
lucianopa-msft: The only changes here are whitespaces ...

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Redux/Action/DiagnosticsAction.swift
lucianopa-msft: We already have dismiss implemented, it is just not through an action. But behavior is the same, I can add an action if required tho.

PR #757, opened 2023-10-30
Set up CI with Azure Pipelines - Localization

ShaunaSong:


azure-pipelines.yml
```diff
@@ -0,0 +1,137 @@
+# This pipeline was generated from the classic pipeline "SCC-SPOOL-communication-ui-library-iOS-localization" on 2023-10-30 with https://aka.ms/1ESPTMigration (v1.0.0): https://dev.azure.com/skype/SCC/_build?definitionId=12523
+#
+# The following items require attention:
```
ahammer, line 3: These items that require attention (as per this doc) are they addressed here in this yaml? I.e. are the secrets in the set in the UI? Do they need to be specified in the file as null?

ShaunaSong: Removed the unused paramters.

Added the secrets '$(AUTHKEY)' into current pipeline variable.

pavelprystinka (approved):

iaulakh (approved):

PR #829, opened 2024-03-14
[Calling] fix API feedback 

iaulakh:
## Purpose

* hide leave call to disable

### DIFF
--- a/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallCompositeOptions/CallCompositeLeaveCallConfirmation.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallCompositeOptions/CallCompositeLeaveCallConfirmation.swift
@@ -1,0 +1,26 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import Foundation
+
+/// Enum defining options for leaving a call confirmation.
+public enum CallCompositeLeaveCallConfirmation {
+    /// Leave call confirmation is enabled.
+    case enable
+    /// Leave call confirmation is disabled.
+    case disable
+
+    /// Provides a `String` representation of the mode.
+    /// - For `.enable`, returns "enable".
+    /// - For `.disable`, returns "disable".
+    var rawValue: String {
+        switch self {
+        case .enable:
+            return "enable"
+        case .disable:
+            return "disable"
+        }
+    }
+}
--- a/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallCompositeOptions/CallCompositeLeaveCallConfirmation.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallCompositeOptions/CallCompositeLeaveCallConfirmation.swift
@@ -7,9 +7,9 @@ import Foundation
 
 /// Enum defining options for leaving a call confirmation.
 public enum CallCompositeLeaveCallConfirmation {
-    /// Leave call confirmation is enabled.
+    /// Enables the leave call confirmation.
     case enable
-    /// Leave call confirmation is disabled.
+    /// Disables the leave call confirmation.
     case disable
 
     /// Provides a `String` representation of the mode.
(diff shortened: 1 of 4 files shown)


AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallCompositeOptions/CallScreenControlBarOptions.swift
```diff
@@ -7,12 +7,12 @@ import Foundation
 
 /// User-configurable options for creating CallScreenControlBar.
 public struct CallScreenControlBarOptions {
```
ahammer, line 9: is this a new struct specifically for control bar?

Is there a specific growth pattern here? I feel this could be in LocalOptions or something existing already?

ahammer: Also if possible, avoid bool, use enum, i.e. in case we want more paths here later, i.e.

LeaveConfirmation.Always
LeaveConfirmation.Never
LeaveConfirmation.NotWhenAlone

etc. Booleans lock us into true/false and if we have future asks the collide, we don't have a seam to expand on.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallCompositeOptions/CallScreenControlBarOptions.swift
```diff
@@ -7,12 +7,12 @@ import Foundation
...
-    /// Hide the leave call confirm dialog.
-    public let hideLeaveCallConfirmDialog: Bool
+    /// disable the leave call confirmation.
```
pavelprystinka, line 10: disable -> Disable

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallCompositeOptions/CallScreenControlBarOptions.swift
pavelprystinka: > Is there a specific growth pattern here?

I think yes, like Button options, and Custom buttons

iaulakh: Yes Adam, this is new struct for control bar. After team discussion we decided to add struct, considering future features about customization.

iaulakh: Adam, I think here boolean is good, as we will either hide or show dialog and this boolean is just for show and hide. what other settings we may want in future that can impact?

iaulakh: In future, if we have any customer ask for role based configuration like below, we can deprecate bool and can expose options as well as roles. At this time, it seems unnecessary. 

LeaveConfirmationBehavior.Always
LeaveConfirmationBehavior.Never
LeaveConfirmationBehavior.NotWhenAlone
LeaveConfirmationBehavior.OnlyForMeetingHost

Mohtasim (approved):

ShaunaSong (dismissed):

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallCompositeOptions/CallCompositeLeaveCallConfirmation.swift
```diff
@@ -0,0 +1,26 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import Foundation
+
+/// Enum defining options for leaving a call confirmation.
+public enum CallCompositeLeaveCallConfirmation {
```
pavelprystinka, line 9: For iOS we are not have to prefix public types with CallComposite.
I think LeaveCallConfirmation will be enough.
But we wee need to sufix it with Options:
LeaveCallConfirmationOptions

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallCompositeOptions/CallScreenControlBarOptions.swift
```diff
@@ -5,14 +5,14 @@
 
 import Foundation
 
-/// User-configurable options for creating CallScreenControlBar.
+/// Configuration options for customizing the CallScreenControlBar.
```
pavelprystinka, line 8: CallScreenControlBar -> call screen control bar

pavelprystinka (approved):

ahammer (approved):

PR #923, opened 2024-08-16
Task  3825794[SDE] UI control and UX workflow works in VoiceOver (iOS)

inancyGit:
Task 3825794 iOS :
-> Added VoiceOver in Timer
-> Added a format for timer

## Purpose

* ...

- [ ] Yes
- [ ] No

## Pull Request Type

- [ ] Bugfix
- [ ] Feature

## Pull request checklist

This PR has considered:
- [ ] Color Theming
- [ ] View Data Injection
- [ ] Demo App UI/UX update

## How to Test

* Open...
* Click on...

## What to Check

### DIFF
--- a/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallCompositeOptions/CallDurationTimer.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallCompositeOptions/CallDurationTimer.swift
@@ -37,4 +37,10 @@ public class CallDurationTimer {
         }
         callTimerAPI.onReset()
     }
+    public func timer() -> String {
+        guard let callTimerAPI = callTimerAPI else {
+            return ""
+        }
+        return callTimerAPI.timerString
+    }
 }
--- a/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallCompositeOptions/CallDurationTimer.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallCompositeOptions/CallDurationTimer.swift
@@ -37,10 +37,4 @@ public class CallDurationTimer {
         }
         callTimerAPI.onReset()
     }
-    public func timer() -> String {
-        guard let callTimerAPI = callTimerAPI else {
-            return ""
-        }
-        return callTimerAPI.timerString
-    }
 }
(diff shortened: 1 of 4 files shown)


AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/CallCompositeOptions/CallDurationTimer.swift
```diff
@@ -37,4 +37,10 @@ public class CallDurationTimer {
         }
         callTimerAPI.onReset()
     }
+    public func timer() -> String {
```
Mohtasim, line 40: This will become public API. We only want to provide contoso Elapsed Time in Number, formatting it will be their own responsibility. Can we think of a way where we do not use the CallTimerAPI interface? 

Since we already have the elapsedTime (TimeInterval), can we think reusing that and keeping other business logic in our viewModel as it is?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/Manager/CallDurationManager.swift
```diff
@@ -7,12 +7,14 @@ import Combine
...
+    var timerString: String {get set}
     func onStart()
     func onStop()
     func onReset()
 }
 
 class CallDurationManager: CallTimerAPI, ObservableObject {
+    @Published var timerString: String = ""
     private var timer: Timer?
     var timeElapsed: TimeInterval? = 0
```
Mohtasim, line 19: Could you think of having timeElapsed as @Published and have the update at the ViewModel maybe?

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/Manager/CallDurationManager.swift
```diff
@@ -7,12 +7,14 @@ import Combine
...
 
 protocol CallTimerAPI {
+    var timerString: String {get set}
     func onStart()
     func onStop()
     func onReset()
 }
 
 class CallDurationManager: CallTimerAPI, ObservableObject {
+    @Published var timerString: String = ""
```
Mohtasim, line 17: Lets try to reuse elapsedTime and format that to a string in the ViewModel which can be accessed by the View for accessibility.

Mohtasim (changes requested):

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/Manager/CallDurationManager.swift
```diff
@@ -10,11 +10,12 @@ protocol CallTimerAPI {
     func onStart()
     func onStop()
     func onReset()
+    var timeElapsed: TimeInterval {get set}
```
Mohtasim, line 13: Can we get rid of this from the protocol?

Mohtasim (changes requested):

Mohtasim (approved):

PR #953, opened 2024-09-06
[Calling][Bugfix] Add accessibility announcement for remote participant onhold and resume

xumo-95:
## Purpose

Fix the issue where voice over doesn't have announcement when remote participant is on hold

- [ ] Yes
- [x] No

## Pull Request Type

- [x] Bugfix
- [ ] Feature

## Pull request checklist

This PR has considered:
- [ ] Color Theming
- [ ] View Data Injection
- [ ] Demo App UI/UX update

## How to Test

Turn on Voiceover
Open Application using double tap.
Navigate to 'Start Experience' button using swipe gesture and double tap to activate it.
Call setup screen will appear
Now activate 'Join Call' using swipe gesture and double tap to activate it
Meeting will be joined.
Observed the voiceover announcement when other person in meeting is on hold.

## What to Check

### DIFF
--- a/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/Calling/Grid/Cell/ParticipantGridCellViewModel.swift
+++ b/AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/Calling/Grid/Cell/ParticipantGridCellViewModel.swift
@@ -186,13 +186,9 @@ class ParticipantGridCellViewModel: ObservableObject, Identifiable {
     }
 
     private func postParticipantStatusAccessibilityAnnouncements(isHold: Bool, participantModel: ParticipantInfoModel) {
-        if isHold {
-            accessibilityProvider.postQueuedAnnouncement(
-                localizationProvider.getLocalizedString(.onHoldAccessibilityLabel, participantModel.displayName))
-        } else {
-            accessibilityProvider.postQueuedAnnouncement(
-                localizationProvider.getLocalizedString(.participantResumeAccessibilityLabel,
-                                                        participantModel.displayName))
-        }
+        let holdResumeAccessibilityLabel = isHold ?
+            localizationProvider.getLocalizedString(.onHoldAccessibilityLabel, participantModel.displayName) :
+            localizationProvider.getLocalizedString(.participantResumeAccessibilityLabel, participantModel.displayName)
+        accessibilityProvider.postQueuedAnnouncement(holdResumeAccessibilityLabel)
     }
 }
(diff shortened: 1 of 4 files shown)


AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/Calling/Grid/Cell/ParticipantGridCellViewModel.swift
```diff
@@ -183,4 +184,15 @@ class ParticipantGridCellViewModel: ObservableObject, Identifiable {
         return callType == .oneToNOutgoing &&
                (participantStatus == nil || participantStatus == .connecting || participantStatus == .ringing)
     }
+
+    private func postParticipantStatusAccessibilityAnnouncements(isHold: Bool, participantModel: ParticipantInfoModel) {
+        if isHold {
```
ahammer, line 189: nit: The only thing forking here is a string, so the if/else
can be reduced to
isHold? .onHoldAccessibilityLabel : .participantResumeAccessibilityLabel

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/Calling/Grid/Cell/ParticipantGridCellViewModel.swift
```diff
@@ -114,6 +114,7 @@ class ParticipantGridCellViewModel: ObservableObject, Identifiable {
 
         if self.isHold != isOnHold {
             self.isHold = isOnHold
+            postParticipantStatusAccessibilityAnnouncements(isHold: self.isHold, participantModel: participantModel)
```
ahammer, line 117: On android this has been upgraded to have a centralized "AccessibilityAnnouncementManager" that has an easy to configure registry of events and announcements.

Like if you think about it, this file is ParticipantGridCellViewModel. That's a bit to close to the view layer for something that is essentially a "data change" notification. I.e. if we make a new layout and don't use this VM, we end up having to duplicate these announcements.

But this is a improvement for a11y announcements for the future on iOS.

AzureCommunicationUI/sdk/AzureCommunicationUICalling/Sources/Presentation/SwiftUI/Calling/Grid/Cell/ParticipantGridCellViewModel.swift
xumo-95: Thanks for the comment! I've updated the code

ShaunaSong (approved):

